PR #4, opened 2020-05-26
Add go report and fix part of reported errors

Fei-Guo:
To improve code quality, we should check go report. This change fixes some of the reported warnings. A full report can be found in https://goreportcard.com/report/github.com/alibaba/openyurt


rambohe-ch:
lgtm

PR #7, opened 2020-05-27
implement `revert` subcommand

charleszheng44:
1. implement revert subcommand, which allows user to revert a yurtcluster back to a kubernetes cluster
2. extract common functions and put them in pkg/yurtctl/util/kubernetes/util.go
3. fix golint issues
4. add yurtctl tutorial draft docs/tutorial/yurtctl.md
5. add docs/todo.md


Fei-Guo:
We should double check how to make the new node controller sa take effect but I think this can be handled in another PR.

PR #11, opened 2020-05-27
bug fix

lonelymemo:
1.change origin nodelifecycle to new defined nodelifecycle
2.fix  edge-controller-manager name err


rambohe-ch:
@lonelymemo would you divide this pr into two prs. one for yurt-controller-manager, and the other one for yurtctl。

and i think name fix for yurtctl is ok, but now no yurt-controller-manager exec file in the docker image, so yurtctl will not work

rambohe-ch:
@lonelymemo 
please sign CLA at first

PR #18, opened 2020-05-28
fix(setup_edgenode): lose last " when replace --kubeconfig in 10-kubeadm.conf

ruster-cn:
do not think the last " when setup_edgenode replace --kubeconfig。when run setup_edgenode,the user get 10-kubeadm.conf that do not have Incorrect file format。


charleszheng44:
@ruster-cn thanks for submitting the pr. The original `10-kubeadm.conf` looks like following

the revised code will add a double quote at the end of the line(something like following) that doesn't exist in the original file.
  
```bash
[Service]
ExecStart=
ExecStart=... --kubeconfig=/var/lib/openyurt/kubelet.conf"
```

Is this double-quote required?

Fei-Guo:
It seems that the problem does not apply. Please reopen if you hit an error and please paste the error log.

PR #23, opened 2020-05-28
Fix yurthub tests

Fei-Guo:
This change fixes health_checker_test and storage_test. 

The ports 8080/8081/8082 can easily be occupied by other processes. This change picks ports that are less likely to be occupied.

"/etc/kubernetes/cache" directory often needs permission to create. This change creates test directory in "/tmp/cache/".


rambohe-ch:
lgtm

PR #49, opened 2020-06-11
set the server-addr before deploying the yurthub

### LINKED ISSUE #47: yurt-hub can't start when I reboot node
# yurt-hub can't start when I reboot node. 

**The reason is that the yurt-hub configuration file takes the apiserver environment variables, but the kubelet does not inject the apiserver environment variables because the yurt-hub is not running properly, so both yurt-hub and kubelet are not running normally at the end.**

charleszheng44:
Instead of getting the apiserver's address from the env during the runtime, let servant pod set it before deploying the yurthub.

fix #47

### DIFF
--- a/config/setup/yurthub.yaml
+++ b/config/setup/yurthub.yaml
@@ -33,7 +33,7 @@ spec:
     command:
     - yurthub 
     - --v=2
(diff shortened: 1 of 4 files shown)


felix0080:
This kind of repair method lacks automaticity and should provide automatic tools

Issue #53, opened 2020-06-16
"revert" not remove the annotation "node.beta.alibabacloud.com/autonomy=true" from edge nodes


vincent-pli:
As described above, to reproduce the issue:
In multiple nodes k8s env.
1. `yurtctl convert` set one node as cloud-node
2. `yurtctl markautonomous`, then check the `edge` node, an annotation: "node.beta.alibabacloud.com/autonomy=true" is added.
3. `yurtctl revert`, then check the original `edge` node, the  annotation: 
"node.beta.alibabacloud.com/autonomy=true"  still there.

PR #54, opened 2020-06-16
Remove annotation: "node.beta.alibabacloud.com/autonomy=true" when re…

vincent-pli:
Fix issue: #53 

`revert` should remove annotation: "node.beta.alibabacloud.com/autonomy=true" on the edge nodes added by `yurtctl markautonomous`.


charleszheng44:
fix #53

PR #61, opened 2020-07-05
fix a sudden power off may result in local cache loss

rambohe-ch:
- what this PR does ?
yurthub's data cache processing is:
  - step1. create temporary data files without O_SYNC
  - step2. delete old data files
  - step3. rename temporary data file into cache data file
after up steps completed, the cache data may not persisted on local disk for not sync write, so when power of machine is suddenly off, the cache data is lost.

in this PR, we modify the up step1 from not sync write to sync write


zyjhtangtang:
LGTM

Issue #62, opened 2020-07-05
support kubernetes v1.16.x version


rambohe-ch:
in order to support k8s v1.16.x version in openyurt, we need to improve the following associated components
  - yurthub:  bookmark process
  - yurt-controller-manager: following kube-controller-manager of k8s v1.16.x version
  - yurtctl: convert v1.16.x k8s to openyurt

rambohe-ch:
@charleszheng44 i will add a pr to slove the adaption for yurthub and yurt-controller-manager, and would you take care of the adaption of yurtctl?

charleszheng44:
> @charleszheng44 i will add a pr to slove the adaption for yurthub and yurt-controller-manager, and would you take care of the adaption of yurtctl?

Yes. just added 1.16 + as a valid version. see #64

PR #64, opened 2020-07-05
Yurtctl 1.16

charleszheng44:
add 1.16 and 1.16+  to the valid version list. see #62


rambohe-ch:
@charleszheng44 the image version of yurthub and yurt-controller-manager are updated to v0.1.1, does yurtctl need any adaption for the new version?

charleszheng44:
> @charleszheng44 the image version of yurthub and yurt-controller-manager are updated to v0.1.1, does yurtctl need any adaption for the new version?

all components are using the image tag "latest". Have you linked the "latest" to "v0.1.1"? if yes, then everything should work fine

PR #69, opened 2020-07-10
initial commit of the yurttunnel

charleszheng44:
1. add the main body of the yurttunel-server and the yurttunnel-agent
2. add implementations of tool functions


pkg/yurttunnel/agent/cmd.go
```diff
@@ -0,0 +1,156 @@
...
+	o := &YurttunnelAgentOptions{
+		kubeConfig: "/etc/kubernetes/kubelet.conf",
+		caFile:     "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+	}
+	return o
+}
+
+// validate validates the YurttunnelServerOptions
+func (o *YurttunnelAgentOptions) validate() error {
+	if o.kubeConfig == "" {
```
rambohe-ch, line 85: the condition is o.kubeConfig != "" ?

pkg/yurttunnel/agent/cmd.go
```diff
@@ -0,0 +1,156 @@
...
+			return fmt.Errorf("%s doesn't exist", o.kubeConfig)
+		}
+		if o.caFile == "" {
+			return errors.New("either '--kube-config' or " +
+				"'--ca-file' need to be set")
+		}
+		if _, err := os.Stat(o.caFile); os.IsNotExist(err) {
+			return fmt.Errorf("%s doesn't exist", o.caFile)
+		}
+		if o.serverAddr == "" {
```
rambohe-ch, line 96: serverAddr is the address of yurt tunnel server, and kubeConfig is rest config for accessing kube-apiserver, both these two parameters should be set ?

pkg/yurttunnel/agent/cmd.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		serverAddr = o.serverAddr
+	}
+	klog.Infof("tunnel server address: %s", serverAddr)
+
+	// 2. create a certificate manager
+	config, err := rest.InClusterConfig()
```
rambohe-ch, line 128: InClusterConfig can not be used by pods that run on edge node for pods can not access kube-apiserver by KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT setting

pkg/yurttunnel/agent/cmd.go
```diff
@@ -0,0 +1,156 @@
...
+func (o *YurttunnelAgentOptions) validate() error {
+	if o.kubeConfig == "" {
+		if _, err := os.Stat(o.kubeConfig); os.IsNotExist(err) {
+			return fmt.Errorf("%s doesn't exist", o.kubeConfig)
+		}
+		if o.caFile == "" {
+			return errors.New("either '--kube-config' or " +
+				"'--ca-file' need to be set")
+		}
+		if _, err := os.Stat(o.caFile); os.IsNotExist(err) {
```
rambohe-ch, line 93: when ca file is not exist, would you get is from kube-apiserver?

pkg/yurttunnel/agent/cmd.go
charleszheng44: will fix

pkg/yurttunnel/agent/cmd.go
charleszheng44: if serverAddr is not set, we will get it through the function GetServerAddr. if kubeconfig not set, we will create a incluster config? So if neither of them are set, the agent can still work?

pkg/yurttunnel/agent/cmd.go
charleszheng44: will fix

pkg/yurttunnel/agent/cmd.go
```diff
@@ -0,0 +1,156 @@
...
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringVar(&o.nodeName, "node-name", o.nodeName,
+		"The name of the edge node.")
+	flags.StringVar(&o.serverAddr, "server-addr", o.serverAddr,
+		"The address of yurttunnel-server")
+	flags.StringVar(&o.kubeConfig, "kube-config", o.kubeConfig,
```
rambohe-ch, line 57: maybe we need add a kubeServerAddr parameter for accessing kube-apiserver when kubeConfig is not set or kubeConfig not exist

pkg/yurttunnel/agent/cmd.go
charleszheng44: You mean from kubeconfig?

pkg/yurttunnel/server/cmd.go
```diff
@@ -0,0 +1,182 @@
...
+		"server will accept master requests at: %s",
+		o.serverAgentAddr, o.serverMasterAddr)
+}
+
+// run starts the yurttunel-server
+func (o *YurttunnelServerOptions) run(stopCh <-chan struct{}) error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	// 1. create the master's clientset
+	clientSet, csErr := kubernetes.CreateClientSet(o.kubeConfig)
```
rambohe-ch, line 120: kubeConfig may not exist, could we add another parameter as a backup?

charleszheng44: `kubeconfig` will be a mandatory option.

pkg/yurttunnel/agent/cmd.go
charleszheng44: if we want to generate the clientset (or rest.Config) based on the apiserver address, we will also need the certificate, which is generated by certmanager. However, certmanager can only be created if a clientset is available. There exist a mutual dependency here. May be we should make `kubeconfig` as a mandatory option?

pkg/yurttunnel/agent/cmd.go
charleszheng44: if `serverAddr` is not set, we can get it by calling `GetServerAddr` function

pkg/yurttunnel/pki/pki.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, fmt.Errorf("loading CA from %s fail: %s", caFile, err)
+	}
+	caPEM, err := ioutil.ReadFile(caFile)
+	if err != nil {
+		return nil, err
+	}
+
+	rootCertPool := x509.NewCertPool()
+	rootCertPool.AppendCertsFromPEM(caPEM)
+	return rootCertPool, nil
```
rambohe-ch, line 124: would you use GenCertPoolUseCA func to refactor the code snippet?

charleszheng44: fixed

rambohe-ch:
lgtm

Issue #67, opened 2020-07-08
edge node support arm/arm64 arch


rambohe-ch:
in order to support arm/arm64 arch for edge node, we need to refactor the following components.
1. yurthub: add arm/arm64 image building
2. yurtctl: adapt for deploying arm/arm64 image

@hwq830 would you take care of it?

hwq830:
Got it.

Fei-Guo:
#71 fixes the issue.

PR #71, opened 2020-07-15
add support to build multi-arch binaries and imgaes (arm/arm64)

hwq830:
add support to build multi-arch binaries and imgaes (arm/arm64)

image


rambohe-ch:
lgtm

and close #67

Issue #56, opened 2020-06-24
edge node cannot be properly set if `--bootstrap-kubeconfig` is used by kubelet


charleszheng44:
Several users have reported that if kubelet of edge nodes using `--bootstrap-kubeconfig`, the `/var/lib/openyurt/kubelet.conf` will keep being reset. This is because `/var/lib/openyurt/kubelet.conf` is created based on `/etc/kubernetes/kubelet.conf`, while `/etc/kubernetes/kubelet.conf` will be reset if kubelet is relying on the `--bootstrap-kubeconfig`.

Fei-Guo:
This should be fixed by #76. Feel free to reopen the issue if the problem still exists.

PR #76, opened 2020-07-23
fix the yurtctl bug

charleszheng44:
yurtctl used to keep the --bootstrap-kubeconfig option, if the bootstrap
kubeconfig exist, which may results in the 10.kubeadm.conf being reset.
To fix this, we remove the --bootstrap-kubeconfig option from the
10.kubeadm.conf when converting kubernetes to yurt cluster.


charleszheng44:
fix #56

PR #88, opened 2020-07-31
Improvement: Give the capacity of slice and map

dddddai:
Give the capacity of slice and map to avoid expansion


rambohe-ch:
lgtm

Issue #81, opened 2020-07-30
change label "alibabacloud.com/xxxx" into a more common name? 


feilengcui008:
I notice there are several labels prefixed with the name "alibabacloud.com", e.g. "alibabacloud.com/is-edge-worker", maybe something like "openyurt.com/xxx" is better.

rambohe-ch:
@feilengcui008 thank you for filing the issue.
@charleszheng44 maybe we can fix these specific labels by using compile settings. and modify alibabacloud.com prefix to openyurt.io

PR #90, opened 2020-08-02
Allow user to specify label prefix

charleszheng44:
1. allow user to set the label prefix (default: openyurt.io)
2. implement a new make rule that allows users to generate setup yaml file
with different PROJECT_PREFIX, LABEL_PREFIX, image REPO and TAG

    e.g.,

    ```bash
    PROJECT_PREFIX=edge LABEL_PREFIX=alibabacloud.com make gen-yaml WHAT=cmd/yurthub
    ```


charleszheng44:
fix #81

PR #93, opened 2020-08-03
Test:add yurthub node autonomy test and docs

Crazy-Kitty:
1) Yurthub node autonomy test, including creating pod, checking pod status after restarting node, and removing pod and so on;
2) Add an example of testing node autonomy in docs

### DIFF
--- a/test/e2e/yurthub/yurthub.go
+++ b/test/e2e/yurthub/yurthub.go
@@ -16,21 +16,138 @@ limitations under the License.
 
 package yurthub
 
-/*
-TODO
-*/
-
 import (
+	"encoding/json"
+	nd "github.com/alibaba/openyurt/test/e2e/common/node"
+	"github.com/alibaba/openyurt/test/e2e/common/ns"
+	p "github.com/alibaba/openyurt/test/e2e/common/pod"
+	ycfg "github.com/alibaba/openyurt/test/e2e/yurtconfig"
 	"github.com/onsi/ginkgo"
+	"github.com/onsi/gomega"
+	apiv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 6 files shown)


test/e2e/yurthub/yurthub.go
```diff
@@ -16,21 +16,138 @@ limitations under the License.
...
 	"k8s.io/kubernetes/test/e2e/framework"
+	"strconv"
+	"time"
 )
 
 const (
 	YURTHUB_NAMESPACE_NAME = "yurthub-test"
+	STOP_NODE_WAIT_MINITE  = 5
+	START_NODE_WAIT_MINITE = 1
+	YURT_NODE_LABLE        = "alibabacloud.com/is-edge-worker"
```
Fei-Guo, line 41: YURT_NODE_LABLE -> YURT_NODE_LABEL

test/e2e/yurthub/yurthub.go
```diff
@@ -16,21 +16,138 @@ limitations under the License.
...
+						Annotations: map[string]string{YURT_NODE_ANNOTATION: "true"},
+					},
+				}
+				patchData, err := json.Marshal(patchNode)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred(), "fail marshal patch node")
+
+				node, err := c.CoreV1().Nodes().Patch(pod.Spec.NodeName, types.StrategicMergePatchType, patchData)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred(), "fail patch node autonomy")
+
+				ginkgo.By("stop node")
```
Fei-Guo, line 116: I think this is the place where user needs to manually restart the node if the test runs against minikube. Please add sufficient instruction in the log so that the user knows what to do next.

Crazy-Kitty:
> Nice change. Thanks for addressing my comments.

Thank you very much for your comments.
I will check my code with some basic tools carefully to find some problems when I submit code next time.

PR #98, opened 2020-08-06
Refactor: implement the HandlerWrapper and decouple server/agent interface from implementation 

charleszheng44:
1. implement the handlerwrapper, which takes in one http.handler and wraps it within another
2. decouple the anp tunnel server/agent using TunnelServer and TunnelAgent interface

### DIFF
--- a/pkg/yurttunnel/agent/agent.go
+++ b/pkg/yurttunnel/agent/agent.go
@@ -21,19 +21,32 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"time"
 
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/klog"
-	anpagent "sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 
 	"github.com/alibaba/openyurt/pkg/yurttunnel/constants"
 	"github.com/alibaba/openyurt/pkg/yurttunnel/pki/certmanager"
 )
+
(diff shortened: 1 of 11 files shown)


pkg/yurttunnel/server/handlerwrapper.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+}
+
+// WrapWrapHandler wraps the coreHandler with chains
+func WrapHandler(coreHandler http.Handler, middlewares ...Middleware) http.Handler {
+	handler := coreHandler
+	if len(middlewares) == 0 {
+		return handler
+	}
+	for i := len(middlewares) - 1; i > 0; i-- {
```
rambohe-ch, line 48: if condition miss i=0, so middlewares[0] can not reach

pkg/yurttunnel/agent/agent.go
```diff
@@ -35,6 +35,34 @@ import (
 	"github.com/alibaba/openyurt/pkg/yurttunnel/pki/certmanager"
 )
 
+// TunnelAgent sets up tunnel to TunnelServer, receive requests
+// from tunnel, and forwards requests to kubelet
+type TunnelAgent interface {
```
rambohe-ch, line 40: now we make up interface for tunnel agent and tunnel server, it is better to separate the interface implementation(for example: anp tunnel) into independent dir or file.

pkg/yurttunnel/server/handlerwrapper.go
charleszheng44: fixed

pkg/yurttunnel/agent/agent.go
charleszheng44: create independent files for them

rambohe-ch:
lgtm

PR #105, opened 2020-08-17
remove kubelet config bootstrap args when manually

carolove:
setup at reset the kubelet step

when setup openyurt edge node, at reset kubelet step,
the bootstrap args in kubelet config is no need.


rambohe-ch:
@carolove thank you for adding pr, please sign CLA at first

charleszheng44:
lgtm

PR #113, opened 2020-08-28
Refactor: refactor the Makefile and build scripts, allow users to specify binar…

charleszheng44:
refactor the Makefile and build scripts, allow users to specify binay and architecture combinations

Example:
    # compile yurthub, yurt-controller-manager and yurtctl-servant with architectures arm64 and arm in the mainland China
    make release WHAT="yurthub yurt-controller-manager yurtctl-servant" ARCH="arm64 arm" REGION=cn


Fei-Guo:
LGTM

PR #120, opened 2020-09-29
Feature: add hubself cert manage mode that use bearer token to bootstrap yurthub agent

rambohe-ch:
yurthub is currently communicating with master using the kubelet component's certificate(/etc/kubernetes/kubelet.conf), and when the node certificate expires, yurthub will not work because kubelet cannot update the node certificate via yurthub.

To solve this case, add node certificate generation and cycle updating ability to yurthub, so yurthub without kubelet node certificate can communicate with master. and when node certificate expires, yurthub can update it by itself.

### DIFF
--- a/pkg/yurthub/certificate/hubself/cert_mgr.go
+++ b/pkg/yurthub/certificate/hubself/cert_mgr.go
@@ -1,0 +1,612 @@
+/*
+Copyright 2020 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hubself
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/alibaba/openyurt/cmd/yurthub/app/config"
+	"github.com/alibaba/openyurt/pkg/projectinfo"
+	hubcert "github.com/alibaba/openyurt/pkg/yurthub/certificate"
+	"github.com/alibaba/openyurt/pkg/yurthub/certificate/interfaces"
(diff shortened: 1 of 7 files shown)


pkg/yurthub/certificate/hubself/cert_mgr.go
```diff
@@ -0,0 +1,596 @@
+package hubself
```
Fei-Guo, line 1: Add file header.

Fei-Guo:
> To solve this case, add node certificate generation and cycle updating ability to yurthub, so yurthub without kubelet node certificate can communicate with master. and when node certificate expires, yurthub can update it by itself.

Can you elaborate:
1) Where to get the `JoinToken` when starting the cert manager?
2) What is the workflow of handling node cert expiration with the new code? I assume the `(ycm *yurtHubCertManager) Update()` method will be called but I don't see the caller.

charleszheng44:
Overall, the PR is good. I just have a question regarding the "jointoken." When the yurthub starts to run, the kubelet should have connected to the kube-apiserver already. Can we reuse the kubelet's bootstrap config instead of passing the "jointoken" through the command line?

charleszheng44 (commented):
Overall, the PR is good. I just have a question regarding the "jointoken." When the yurthub starts to run, the kubelet should have connected to the kube-apiserver already. Can we reuse the kubelet's bootstrap config instead of passing the "jointoken" through the command line?

charleszheng44 (approved):
/lgtm

pkg/yurthub/certificate/hubself/cert_mgr.go
rambohe-ch: fixed

rambohe-ch:
> > To solve this case, add node certificate generation and cycle updating ability to yurthub, so yurthub without kubelet node certificate can communicate with master. and when node certificate expires, yurthub can update it by itself.
> 
> Can you elaborate:
> 
> 1. Where to get the `JoinToken` when starting the cert manager?
when transfer  k8s worker node into  openyurt worker node, currently yurtctl deploy a yurthub on the node. so bootstrap join token will be created by yurtctl before yurthub deployment. and we will add this implementation in another pull request.

> 2. What is the workflow of handling node cert expiration with the new code? I assume the `(ycm *yurtHubCertManager) Update()` method will be called but I don't see the caller.
yeah, the caller for cert update will be added in another pull request

rambohe-ch:
> Overall, the PR is good. I just have a question regarding the "jointoken." When the yurthub starts to run, the kubelet should have connected to the kube-apiserver already. Can we reuse the kubelet's bootstrap config instead of passing the "jointoken" through the command line?

if no join token is set for yurthub when hubself cert mode is selected, orginal kubelet.conf also will be used.

PR #129, opened 2020-10-29
bugfix: remove conn reference info in dialer when underlying tcp connection is closed by http transport

rambohe-ch:
- background:
  1. tcp connection creation path:  yurthub proxy request --> http.Transport.RoundTrip() --> connRotation.dialer.Dial() --> net.Dial(), and net.Conn reference is stored in the connRotation.dialer
  2. tcp connection is closed in the followint two paths.
        - path1:  yurthub healthchecker --> connRotation.dialer.Close(address string)
        - path2:  when idle timeout happen, http.Transport --> net.Conn.Close()

in the close path1, net.Conn reference is removed in connRotation.dialer, but in close path2,  the connection reference info will always be remained in the connrotation dialer, so may cause yurthub memory leak.

- solution:
  add closable connection and overwrite net.Conn Close func, so close path2 becomes like this:  
        --> close path2: when idle timeout happen, http.Transport --> closableConn.Close() --> net.Conn.Close(), in the closableConn.Close(), we will remove the reference info in dialer.

### DIFF
--- a/pkg/yurthub/util/connrotation.go
+++ b/pkg/yurthub/util/connrotation.go
@@ -25,6 +25,28 @@ import (
 	"k8s.io/klog"
 )
 
+// closableConn is used to remove reference in dialer
+// when conn is closed by http transport
+type closableConn struct {
+	net.Conn
+	dialer *Dialer
+	addr   string
+}
+
+// Close is called by http transport, so remove the conn reference in dialer
+// and close conn.
+func (c *closableConn) Close() error {
+	c.dialer.mu.Lock()
+	remain := len(c.dialer.addrConns[c.addr])
+	if remain >= 1 {
+		delete(c.dialer.addrConns[c.addr], c)
+		remain--
+	}
+	c.dialer.mu.Unlock()
+	klog.Infof("close connection from %s to %s for %s dialer, remain %d connections", c.Conn.LocalAddr().String(), c.addr, c.dialer.name, remain)
+	return c.Conn.Close()
+}
+
 // DialFunc is a shorthand for signature of net.DialContext.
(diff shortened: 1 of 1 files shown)


pkg/yurthub/util/connrotation.go
```diff
@@ -79,11 +101,12 @@ func (d *Dialer) Close(address string) {
 	d.mu.Lock()
```
Fei-Guo, line 101: Will the problem be resolved by expanding the lock scope in `Close` method to include the for loop, so that you don't need to copy the map. 

I don't see an immediate reason why the conn.Close() has to be called asynchronously.

pkg/yurthub/util/connrotation.go
```diff
@@ -79,11 +101,12 @@ func (d *Dialer) Close(address string) {
...
 	conns := d.addrConns[address]
 	delete(d.addrConns, address)
+	d.addrConns[address] = nil
```
Fei-Guo, line 104: Why do you want to add an entry to the map again?

Fei-Guo (commented):
A few comments.

pkg/yurthub/util/connrotation.go
rambohe-ch: It may take a while to complete conn.Close(), and it may be conn.Dial() to create a new one immediately when the connection is closed. So in order not to block connection creation, choose to copy the map.

pkg/yurthub/util/connrotation.go
rambohe-ch: fixed

Fei-Guo (approved):

PR #132, opened 2020-10-30
refactor: when cache data of hub agent is removed not on purpose, return 404 instead of empty objects for kubelet request

rambohe-ch:
- background:
   1. at present, when cache data of yurthub(under /etc/kubernetes/cache directory) is removed, local proxy will return empty object slice for kubelet request.
   2. when network between edge and cloud is disconnected, yurthub will use empty cache to response kubelet request. if response of list pods request is empty objects slice(len(ListPods.Items[])=0)，all pods that are running on the current node will be killed. so it will effect the edge business.

- solution:
   if specified data is not found in cache data(for example: list pods)，yurthub not return empty objects slice for request, yurthub will return 404 status code for request, so kubelet only re-request data but not kill pods.

### DIFF
--- a/pkg/yurthub/proxy/local/local.go
+++ b/pkg/yurthub/proxy/local/local.go
@@ -26,11 +26,13 @@ import (
 	"time"
 
 	manager "github.com/alibaba/openyurt/pkg/yurthub/cachemanager"
+	"github.com/alibaba/openyurt/pkg/yurthub/storage"
 	"github.com/alibaba/openyurt/pkg/yurthub/util"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


pkg/yurthub/proxy/local/local.go
```diff
@@ -152,14 +159,13 @@ func (lp *LocalProxy) localWatch(w http.ResponseWriter, req *http.Request) error
 	if !ok {
 		err := fmt.Errorf("unable to start watch - can't get http.Flusher: %#v", w)
 		util.Err(errors.NewInternalError(err), w, req)
-		return err
+		return nil
```
Fei-Guo, line 162: I understand the intention of `return nil` but the logic looks counterintuitive. 

How about call `util.Err()` in every failure case in localWatch/localPost/localReqCache and remove the code
`util.Err(errors.NewBadRequest(err.Error()), w, req)` in line 86?  The refactor effort should be small.

rambohe-ch: fixed

Fei-Guo: Thanks for addressing the comment.

Fei-Guo (approved):

PR #143, opened 2020-11-15
refactor: extract the process of getting tunnel server info into serveraddr dir

rambohe-ch:
refactor the process of getting tunnel server address for the following reasons:
1. put the getting tunnel server info func(`GetYurttunelServerDNSandIP`) in certmanager dir is confusing
2. in func `GetTunnelServerAddr`, get x-tunnel-server-svc twice and it is not necessary.

so we did the following refactor
1. create a new dir named `serveraddr` and put it as sub dir of server.
2. add `getTunnelServerResources` and `extractTunnelServerDNSandIPs` func to separate resource getting and resource extraction, for avoiding get service twice.

### DIFF
--- a/pkg/yurttunnel/server/serveraddr/addr.go
+++ b/pkg/yurttunnel/server/serveraddr/addr.go
@@ -1,0 +1,281 @@
+/*
+Copyright 2020 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package serveraddr
+
+import (
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/alibaba/openyurt/pkg/projectinfo"
+	"github.com/alibaba/openyurt/pkg/yurttunnel/constants"
+
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/yurttunnel/server/serveraddr/addr.go
```diff
@@ -0,0 +1,283 @@
...
+	for _, tmpIP := range ips {
+		// we use the first non-loopback IP address.
+		if tmpIP.String() != "127.0.0.1" {
+			ip = tmpIP
+			break
+		}
+	}
+
+	if ip == nil {
+		if len(dnsNames) == 0 {
```
charleszheng44, line 63: may be we can remove unnecessary else clause?
```go
if len(dnsNames) == 0 {
    return "", errors.New("there is no available ip")
}
host = dnsNames[0]
```

pkg/yurttunnel/server/serveraddr/addr.go
```diff
@@ -0,0 +1,283 @@
...
+		} else {
+			host = dnsNames[0]
+		}
+	} else {
+		host = ip.String()
+	}
+
+	for _, port := range svc.Spec.Ports {
+		if port.Name == constants.YurttunnelServerAgentPortName {
+			if svc.Spec.Type == corev1.ServiceTypeNodePort {
```
charleszheng44, line 74:
```go
tcpPort = port.Port
if svc.Spec.Type == corev1.ServiceTypeNodePort {
    tcpPort = port.NodePort
}
```

charleszheng44 (changes requested):
LGTM excepting some minor issues.

pkg/yurttunnel/server/serveraddr/addr.go
rambohe-ch: fixed

pkg/yurttunnel/server/serveraddr/addr.go
rambohe-ch: fixed

PR #152, opened 2020-11-17
Bugfix: fix some yurtctl errors

wenjun93:


pkg/yurtctl/util/kubernetes/util.go
```diff
@@ -56,7 +56,7 @@ var (
 	// PropagationPolicy defines the propagation policy used when deleting a resource
 	PropagationPolicy = metav1.DeletePropagationForeground
 	// WaitServantJobTimeout specifies the timeout value of waiting for the ServantJob to be succeeded
-	WaitServantJobTimeout = time.Minute * 2
+	WaitServantJobTimeout = time.Minute * 5
```
Fei-Guo, line 59: Why do you increase the timeout?

wenjun93: In my test env, 2min is too short for the job finish , however, after timeout, the convert job succeeded

wenjun93: sure, thanks

pkg/yurtctl/cmd/convert/convert.go
```diff
@@ -198,12 +201,20 @@ func (co *ConvertOptions) RunConvert() (err error) {
 	klog.V(4).Info("the server version is valid")
 
 	// 2. label nodes as cloud node or edge node
-	nodeLst, err := co.clientSet.CoreV1().Nodes().List(metav1.ListOptions{})
+	var nodeList *v1.NodeList
+	var requirement *labels.Requirement
+	// skip master node
```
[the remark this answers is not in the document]
wenjun93: seems the nodelist is not used for tunnel deploy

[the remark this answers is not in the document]
charleszheng44: The nodelist here, will be used to add the cloud/edge node label to each node. If we skip the master node, then the master node will not be marked as the cloud node while the `yurt-tunnel-server` can only be deployed on the cloud node.

charleszheng44:
LGTM

PR #155, opened 2020-11-19
refactor: adapt k8s.io/klog/v2 for yurt tunnel

rambohe-ch:
because underlying imported lib(anp) in yurt tunnel used klog/v2 as logger, so adapt klog/v2 in yurt tunnel for optimizing log output

### DIFF
--- a/pkg/yurttunnel/agent/cmd.go
+++ b/pkg/yurttunnel/agent/cmd.go
@@ -18,14 +18,13 @@ package agent
 
 import (
 	"errors"
-	"flag"
 	"fmt"
 	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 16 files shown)


pkg/yurttunnel/agent/cmd.go
```diff
@@ -79,8 +78,8 @@ func NewYurttunnelAgentCommand(stopCh <-chan struct{}) *cobra.Command {
 		"The identifiers of the agent, which will be used by the server when choosing agent.")
 
 	// add klog flags as the global flagsets
-	klog.InitFlags(nil)
-	flags.AddGoFlagSet(flag.CommandLine)
+	//klog.InitFlags(nil)
```
Fei-Guo, line 81: How about just removing those lines instead of commenting them out.

rambohe-ch: fixed

PR #159, opened 2020-11-23
the configmap of request redirect is not correct

vincent-pli:
The `configmap` which used for request redirect named `yurt-tunnel-server-cfg`, the code will generated a wrong name:
`yurttunnel-server-cfg`, that's make the functionality not work.


rambohe-ch:
LGTM

PR #167, opened 2020-12-06
bugfix: fix panic when csr is not found in lister

rambohe-ch:
fix an panic when `yca.csrInformer.Lister().Get(csrName)` returns nil and error, and then comes into `approveYurttunnelCSR` process because not return when error happened, then a panic will be triggered.

First of all, this scenario is not easy to reproduce. 
It requires that an update of the CSR has just occurred before the CSR is deleted. This starts the CSR sync processing, and Lister can returns an error.


charleszheng44 (approved):
LGTM

PR #170, opened 2020-12-10
Fix: trim all trailing '-' characters of project prefix

stormgbs:

Trim one or more '-' in ProjectPrefix.


rambohe-ch:
LGTM

PR #174, opened 2020-12-18
Test: add integration test for yurttunnel

charleszheng44:

Add integration test for yurttunnel

None

test/integration/yurttunnel_test.go

go test -v test/integration/yurttunel_test.go

None


test/integration/yurttunnel_test.go
```diff
@@ -0,0 +1,208 @@
...
+			Certificates: []tls.Certificate{
+				genCert(t, GeneircCertFile, GenericKeyFile),
+			},
+			ClientCAs: genCAPool(t, RootCAFile),
+		},
+		Addr:    fmt.Sprintf(":%d", DummyServerPort),
+		Handler: m,
+	}
+
+	klog.Info("[TEST] dummy-server is listening on :9515")
```
rambohe-ch, line 93: how about use var DummyServerPort  to substitute hard code 9515

test/integration/yurttunnel_test.go
```diff
@@ -0,0 +1,208 @@
...
+			genCert(t, GeneircCertFile, GenericKeyFile),
+		},
+		RootCAs:    genCAPool(t, RootCAFile),
+		ServerName: "127.0.0.1",
+	}
+	tunnelAgent := ta.NewTunnelAgent(
+		&tlsCfg,                             /* tlsCfg */
+		fmt.Sprintf(":%d", ServerAgentPort), /* tunnelServerAddr */
+		"dummy-agent",                       /* nodeName */
+		"127.0.0.1",                         /* agentIdentifiers */
```
rambohe-ch, line 179: agentIdentifiers should be set with "ipv4=127.0.0.1", not only ip address.
and i think that agentIdentifiers definition is confusing, it's not so easy for user to setup. maybe we need to add some verification or optimization for  agentIdentifiers

test/integration/yurttunnel_test.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		t.Fatalf("fail to generate http request: %v", err)
+	}
+
+	rep, err := c.Do(r)
+	if err != nil {
+		t.Fatalf("fail to send request to the server: %v", err)
+	}
+	defer rep.Body.Close()
+
```
rambohe-ch, line 125: how about add a snippet of http status code verification

test/integration/yurttunnel_test.go
charleszheng44: Oops, I forgot it. Yes, I agree. The `agentIdentifier` can be confusing, Maybe we can add a detailed explanation with an example in the help message? 

The ANP agent will validate the identifier. If the identifier is invalid, the agent will be registered as a random backend on the server-side. If the server can't find a backend by identifier, it will choose a random one, that's why even though the identifier is invalid here, but the server can still redirect the request.

test/integration/yurttunnel_test.go
charleszheng44: Yes. will switch to the status code

test/integration/yurttunnel_test.go
charleszheng44: Yes. Will replace the magic number with the var.

Issue #171, opened 2020-12-10
[feature request] the  definition of NodePool and UnitedDeployment 


kadisi:
**What would you like to be added**:

In the edge scenario, compute nodes have strong regional attributes. There may only be one WOker node in the same physical location, and there may also be multiple worker nodes in the same physical location. Therefore, from the work Node resource perspective of edge nodes, they need to be divided into different node pools (NodePool) to represent the same set of features. After dividing nodes pool, there will be a demand for application of grouping management, users need to be deployed application according to the unit, combined with the concept of node pool, to deploy applications on different nodes in the pool, pool dimensions at the nodes to expansion of application, upgrade, such as operation, at the same time, network access is also carried out in accordance with the node pool dimensions of network communication.

**Why is this needed**:

So we define `NodePool` and `UnitedDeployment` for both scenarios using Kubernetes CRD as the abstract

`NodePool`:

Nodes can be added and removed from the node pool.
Nodes within the node pool can be managed uniformly, such as label, annotation, and taints.

`UnitedDeployment`:
UnitedDeployment can use one of k8s Deployment, Daemonset, StatefulSet as a template for Deployment in a different node pool. At the same time, you can set the number of POD replicas deployed by different node pools

ref #124 

cc @huangyuqi @Fei-Guo @rambohe-ch @charleszheng44 

task list:

- [x] [add nodepool and uniteddeployment crd proposal ](https://github.com/alibaba/openyurt/pull/173)
- [x] [add yurt-app-manager api](https://github.com/alibaba/openyurt/pull/175)
- [x] [add yurt-app-manager nodepool controller add webhook](https://github.com/alibaba/openyurt/pull/178)
- [x] [add yurt-app-manager unitedeployment controller add webhook](https://github.com/alibaba/openyurt/pull/177)
- [x] [add makefile cmd for yurt-app-manager](https://github.com/alibaba/openyurt/pull/182)

Fei-Guo:
Sounds great. Looking forward to them. By the way, is this a feature request? This should be proposal since you already implement them! :)

kadisi:
> Sounds great. Looking forward to them. By the way, is this a feature request? This should be proposal since you already implement them! :)

ok @Fei-Guo  ref #173

yixingjia:
Looking forward for the NodePool feature!  

By the way please also consider the follow scenario if possible in the future:

**Requirement**:

[...]

PR #175, opened 2020-12-22
add nodepool and uniteddeployment api

kadisi:

ref: #171 

add an nodepool and uniteddeployment api.  

yurtapp-manager is a kubernetes operator for nodepool and uniteddeployment.

@Fei-Guo @charleszheng44


Fei-Guo (commented):
It seems that you choose `unit` as group name which seems to be confusing. How about just use `apps` as group name?
Also, how about rename yurtunit-manager to yurt-app-manager (we may have yurt-device-manager in the future).

kadisi:
> It seems that you choose `unit` as group name which seems to be confusing. How about just use `apps` as group name?
> Also, how about rename yurtunit-manager to yurt-app-manager (we may have yurt-device-manager in the future).

@Fei-Guo good idea

kadisi:
> It seems that you choose `unit` as group name which seems to be confusing. How about just use `apps` as group name?
> Also, how about rename yurtunit-manager to yurt-app-manager (we may have yurt-device-manager in the future).

good idea

kadisi:
@Fei-Guo PTAL

Fei-Guo (approved):
Thanks for addressing the comments.

PR #179, opened 2020-12-31
fix: deploy servantJob and yurt-controller-manager using host network mode.

zyjhtangtang:
deploy servantJob and yurt-controller-manager using host network mode.

The two components were deployed using container network mode by default.  Now, modify them to host network mode.
(1)add hostNetwork:true for ServantJobTemplate.
(2)add hostNetwork:true for YurtControllerManagerDeployment.

NONE

When k8s cluster don't install container network plugin，the two components will fail to deploy.

make test
make build
_output/bin/yurtctl convert --provider [minikube|ack]

### DIFF
--- a/pkg/yurtctl/constants/constants.go
+++ b/pkg/yurtctl/constants/constants.go
@@ -129,6 +129,7 @@ spec:
         app: yurt-controller-manager
     spec:
       serviceAccountName: yurt-controller-manager
+      hostNetwork: true
       affinity:
         nodeAffinity:
           # we prefer allocating ecm on cloud node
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@zyjhtangtang why we need to use host network mode for servantJob and yurt-controller-manager ?

zyjhtangtang:
> @zyjhtangtang why we need to use host network mode for servantJob and yurt-controller-manager ?

When k8s cluster don't install container network plugin，the two components will fail to deploy.

rambohe-ch:
> > @zyjhtangtang why we need to use host network mode for servantJob and yurt-controller-manager ?
> 
> When k8s cluster don't install container network plugin，the two components will fail to deploy.

ok, would you add this info into pull request?

PR #183, opened 2021-01-04
Feature: create cluster-info configmap by yurtctl for adapting cert-mode=hubself of yurthub

rambohe-ch:
create a cluster-info configmap in the cluster before converting a k8s cluster to openyurt cluster, because yurthub will use this configmap to bootstrap the edge node.  and when k8s cluster was set up by kubeadm tool, the cluster-info configmap is created by default.


zyjhtangtang:
LGTM

PR #192, opened 2021-01-08
fix: yurtctl fail to convert minikube cluster.

zyjhtangtang:
(1) support domain name for server addr.
(2)add  Usages and  Groups when creating bootstrap token.

No issue, the reason for the modification is as follows:
(1)minikube use domain name control-plane.minikube.internal to access apiserver.
(2)bootstrap token need set  authentication,signing for  Usages and system:bootstrappers:kubeadm:default-node-token for groups.

make test
make build
_output/bin/yurtctl convert --provider [minikube|ack|kubeadm]

### DIFF
--- a/config/yurtctl-servant/setup_edgenode
+++ b/config/yurtctl-servant/setup_edgenode
@@ -87,36 +87,14 @@ error() {
 
 check_addr()
 {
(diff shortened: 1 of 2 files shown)


rambohe-ch:
LGTM

PR #196, opened 2021-01-12
bugfix: nodepool taint sync error

kadisi:
Signed-off-by: kadisi <iamkadisi@163.com>

If the host has the same  as taint the node pool before joining the node pool, joining fails


rambohe-ch:
LGTM

PR #197, opened 2021-01-12
Add serveraddr/addr.go unit test

BearLu:

### DIFF
--- a/pkg/yurttunnel/server/serveraddr/addr_test.go
+++ b/pkg/yurttunnel/server/serveraddr/addr_test.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2020 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package serveraddr
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestGetDefaultDomainsForSvcInputParamEmptyChar(t *testing.T) {
+	domains := getDefaultDomainsForSvc("", "")
+	if len(domains) != 0 {
+		t.Error("domains len is not equal zero")
+	}
+}
+
+func TestGetDefaultDomainsForSvc(t *testing.T) {
+	ns := "hello"
+	name := "world"
+	domains := getDefaultDomainsForSvc(ns, name)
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@BearLu Thank you for uploading pr.  how about rebase these three commits into one commit?

rambohe-ch:
@BearLu please sign CLA

pkg/yurttunnel/server/serveraddr/addr_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	ns :="hello"
+	name := "world"
+	domains = getDefaultDomainsForSvc(ns, name)
+	if len(domains) == 0 {
+		t.Log("domains len is zero")
+	} else {
+		if len(domains) == 4{
+			// expected, actual
+			assert.Equal(t, name, domains[0])
```
rambohe-ch, line 23: maybe it does not good to import new lib(`github.com/stretchr/testify/assert`), how about only use `testing` to replace it?

pkg/yurttunnel/server/serveraddr/addr_test.go
```diff
@@ -0,0 +1,29 @@
+package serveraddr
+
+import (
+	"testing"
+	"github.com/stretchr/testify/assert"
+	"fmt"
+)
+
+func TestGetDefaultDomainsForSvc(t *testing.T) {
+	domains := getDefaultDomainsForSvc("", "")
```
rambohe-ch, line 10: you can add a new test case or a sub test case for ns and name empty setting.

pkg/yurttunnel/server/serveraddr/addr_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"testing"
+)
+
+func TestGetDefaultDomainsForSvcInputParamEmptyChar(t *testing.T) {
+	domains := getDefaultDomainsForSvc("", "")
+	if len(domains) == 0 {
+		t.Log("domains len is zero")
+	}
```
rambohe-ch, line 27: if length of domains is not 0, the case is also passed. so it is better to use `if len(domains) != 0` verification and then `t.Errorf()` to output error message.

BearLu: OK,I will modify it later.

pkg/yurttunnel/server/serveraddr/addr_test.go
BearLu: I use testing replace it

pkg/yurttunnel/server/serveraddr/addr_test.go
BearLu: I modified it

pkg/yurttunnel/server/serveraddr/addr_test.go
BearLu: I modified it

rambohe-ch:
LGTM

PR #198, opened 2021-01-14
fix bug when ServeHTTP panic

aholic:
when client close the connection in advance, reserveproxy.ServeHTTP will panic, and everything after ServeHTTP will be skipped except "defer". so the element in reqChan(function WithMaxInFlightLimit) leaks.

reproduce:
1. send a watch curl to yurthub, and stop using ctrl+c （eg. `curl '127.0.0.1:10261/apis/node.k8s.io/v1beta1/runtimeclasses?resourceVersion=1235193475&timeout=5m&timeoutSeconds=300&watch=true' -o /dev/null -v`）
2. do step-1 for 250 times(depends on the config `MaxRequestInFlight`)
3. send any request to yurt, you'll get 429

this pr moves the operation of the chan into defer, to make sure it will run after ServeHTTP

see the reproduce steps mentioned above


rambohe-ch:
LGTM

PR #199, opened 2021-01-14
delete system namespaces for yurt-app-manager yaml

kadisi:
Signed-off-by: kadisi <iamkadisi@163.com>

### DIFF
--- a/config/yurt-app-manager/manager/manager.yaml
+++ b/config/yurt-app-manager/manager/manager.yaml
@@ -1,9 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  labels:
-    control-plane: yurt-app-manager
-  name: system
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


config/yurt-app-manager/manager/manager.yaml
```diff
@@ -1,9 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  labels:
-    control-plane: yurt-app-manager
-  name: system
```
rambohe-ch, line 6: system namespace is deleted, but manager deployment is still using system namespace, so deployment may cause  error.

kadisi: all the yaml config of yurt-app-manager is created by `kustomize`, Finally, the `kube-system` namespace will be used uniformly。

rambohe-ch:
LGTM

PR #214, opened 2021-02-03
Fix rebuild yurtctl exception

vincent-pli:
When rebuild the `yurtctl` binary:
`make WHAT=cmd/yurtctl` hits exception, it said:
rm: cannot remove '/root/go/src/github.com/alibaba/openyurt/_output/bin': Is a directory

So fix it.


vincent-pli:
Caused by my old version of `openyurt`, close it

PR #218, opened 2021-02-08
update yurt-controller-manager.yaml

DrmagicE:
Update the `yurt-controller-manager.yaml` and  keep it in sync with https://github.com/alibaba/openyurt/blob/master/pkg/yurtctl/constants/constants.go


rambohe-ch:
LGTM

PR #232, opened 2021-03-05
Feature: upgrade apiserver-network-proxy to v0.0.15

charleszheng44:
1. upgrade the apiserver-network-proxy to the latest release version
2. use `destHost` proxy strategy by default

NONE

NONE

follow the yurt-tunnel tutorial

NONE

### DIFF
--- a/config/setup/yurt-tunnel-agent.yaml
+++ b/config/setup/yurt-tunnel-agent.yaml
@@ -59,6 +59,7 @@ spec:
         - yurt-tunnel-agent
         args:
         - --node-name=$(NODE_NAME)
+        - --node-ip=$(NODE_IP)
         image: openyurt/yurt-tunnel-agent:latest
         imagePullPolicy: IfNotPresent
         name: yurt-tunnel-agent
@@ -74,10 +75,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-        - name: POD_IP
+        - name: NODE_IP
           valueFrom:
             fieldRef:
-              fieldPath: status.podIP
+              fieldPath: status.hostIP
       hostNetwork: true
       restartPolicy: Always
       volumes:
--- a/pkg/yurttunnel/agent/cmd.go
+++ b/pkg/yurttunnel/agent/cmd.go
@@ -19,7 +19,6 @@ package agent
 import (
 	"errors"
 	"fmt"
-	"net"
 	"os"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


Fei-Guo (approved):

config/setup/yurt-tunnel-agent.yaml
```diff
@@ -78,6 +79,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: status.podIP
+        - name: NODE_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
```
rambohe-ch, line 85: The value of POD_IP is same as node ip because hostNetwork field is true. so we should keep only one env here.

charleszheng44: Ah, I didn't notice that. Will remove the NODE_IP as we will always use the host network for the tunnel component.

pkg/yurttunnel/agent/cmd.go
```diff
@@ -116,7 +123,7 @@ func (o *YurttunnelAgentOptions) complete() error {
 
 		podIP := os.Getenv("POD_IP")
 		if len(podIP) != 0 && net.ParseIP(podIP).To4() != nil {
-			o.agentIdentifiers = fmt.Sprintf("%s,ipv4=%s", o.agentIdentifiers, podIP)
+			o.agentIdentifiers = fmt.Sprintf("%s&ipv4=%s&ipv4=%s", o.agentIdentifiers, podIP, o.nodeIP)
```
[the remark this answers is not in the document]
charleszheng44: OK, let's just use the nodeIP as the identifier for now.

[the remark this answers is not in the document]
rambohe-ch: ok, please remove the podIP

config/setup/yurt-tunnel-agent.yaml
rambohe-ch: ok, keep the NODE_IP env and remove the POD_IP env, but keep hostNetwork is true.

rambohe-ch:
@charleszheng44 please fix the unit test error.

pkg/yurttunnel/agent/cmd.go
```diff
@@ -100,6 +101,10 @@ func (o *YurttunnelAgentOptions) validate() error {
 		return errors.New("--node-name is not set")
 	}
 
+	if o.nodeIP == "" {
+		return errors.New("--node-ip is not set")
```
rambohe-ch, line 105: how about get nodeIP for NODE_IP env when `o.nodeIP` is empty, so don't need to update the command parameter in yaml file.

pkg/yurttunnel/agent/cmd.go
```diff
@@ -112,12 +117,7 @@ func (o *YurttunnelAgentOptions) complete() error {
 	var err error
 
 	if len(o.agentIdentifiers) == 0 {
-		o.agentIdentifiers = fmt.Sprintf("host=%s", o.nodeName)
```
rambohe-ch, line 115: we need to keep the host={nodeName} in `o.agentIdentifiers`

rambohe-ch:
LGTM

rambohe-ch (approved):
LGTM

PR #234, opened 2021-03-11
Add support for the conversion between kind and OpenYurt cluster

### LINKED ISSUE #230: Add support for the conversion between kind cluster and OpenYurt cluster.
OpenYurt supports the conversion between the OpenYurt cluster and the Kubernetes cluster created by minikube, kubeadm, and ack. It may be necessary to add support for the conversion between kind cluster and OpenYurt cluster.

Peeknut:

1.Only one node in the cluster is installed with openyurt, and then any node in the cluster can be converted.

2.Add support for the conversion between kind and OpenYurt cluster.

fixes #230.

make test
make build
cp _output/bin/yurtctl /bin/
yurtctl convert
yurtctl revert

### DIFF
--- a/hack/lib/release-images.sh
+++ b/hack/lib/release-images.sh
@@ -1,11 +1,11 @@
 # Copyright 2020 The OpenYurt Authors.
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -26,7 +26,7 @@ YURT_BUILD_IMAGE="golang:1.13-alpine"
 readonly -a YURT_BIN_TARGETS=(
     yurthub
     yurt-controller-manager
-    yurtctl-servant
+    yurtctl
     yurt-tunnel-server
     yurt-tunnel-agent
     yurt-app-manager
@@ -41,7 +41,7 @@ readonly -a SUPPORTED_ARCH=(
 readonly SUPPORTED_OS=linux
 
 readonly -a bin_targets=(${WHAT:-${YURT_BIN_TARGETS[@]}})
-readonly -a bin_targets_without_servant=("${bin_targets[@]/yurtctl-servant}")
+readonly -a bin_targets_process_servant=("${bin_targets[@]/yurtctl-servant/yurtctl}")
 readonly -a target_arch=(${ARCH:-${SUPPORTED_ARCH[@]}})
 readonly region=${REGION:-us}
 
@@ -74,7 +74,7 @@ function build_multi_arch_binaries() {
         cd /opt/src; umask 0022; \
         rm -rf ${YURT_LOCAL_BIN_DIR}/* ;"
     for arch in ${target_arch[@]}; do
-        sub_commands+="GOARCH=$arch bash ./hack/make-rules/build.sh $(echo ${bin_targets_without_servant[@]}); "
+        sub_commands+="GOARCH=$arch bash ./hack/make-rules/build.sh $(echo ${bin_targets_process_servant[@]}); "
     done
     sub_commands+="chown -R $(id -u):$(id -g) /opt/src/_output"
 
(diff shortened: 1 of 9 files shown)


pkg/yurtctl/constants/constants.go
```diff
@@ -186,7 +186,7 @@ spec:
         - /bin/sh
         - -c
         args:
-        - "nsenter -t 1 -m -u -n -i -- /bin/yurtctl convert edgenode --yurthub-image {{.yurthub_image}} --join-token {{.joinToken}}"
+        - "cp /usr/local/bin/yurtctl /tmp && nsenter -t 1 -m -u -n -i -- /var/tmp/yurtctl convert edgenode --yurthub-image {{.yurthub_image}} --join-token {{.joinToken}} && rm /tmp/yurtctl"
```
charleszheng44, line 189: can we execute the `/usr/local/bin/yurtctl` directly?

pkg/yurtctl/constants/constants.go
```diff
@@ -231,7 +231,7 @@ spec:
         - /bin/sh
         - -c
         args:
-        - "nsenter -t 1 -m -u -n -i -- /bin/yurtctl revert edgenode"
+        - "cp /usr/local/bin/yurtctl /tmp && nsenter -t 1 -m -u -n -i -- /var/tmp/yurtctl revert edgenode && rm /tmp/yurtctl"
```
charleszheng44, line 234: Same as the previous comment.

pkg/yurtctl/util/edgenode/util.go
```diff
@@ -113,10 +113,8 @@ func GetNodeName() (string, error) {
 	}
 
 	//2. find --hostname-override in 10-kubeadm.conf
-	nodeName, err := GetSingleContentFromFile(KubeletSvcPath, KubeletHostname)
-	if err != nil {
-		return "", err
-	} else if nodeName != "" {
+	nodeName, _ := GetSingleContentFromFile(KubeletSvcPath, KubeletHostname)
+	if nodeName != "" {
```
charleszheng44, line 117: Can we log the error instead of ignoring it?

pkg/yurtctl/util/edgenode/util.go
```diff
@@ -128,10 +126,8 @@ func GetNodeName() (string, error) {
 	}
 	for _, ef := range environmentFiles {
 		ef = strings.Split(ef, "-")[1]
-		nodeName, err = GetSingleContentFromFile(ef, KubeletHostname)
-		if err != nil {
-			return "", err
-		} else if nodeName != "" {
+		nodeName, _ = GetSingleContentFromFile(ef, KubeletHostname)
+		if nodeName != "" {
```
charleszheng44, line 130: Same as the previous comment.

pkg/yurtctl/util/kubernetes/util.go
```diff
@@ -60,7 +60,7 @@ const (
 
 var (
 	// PropagationPolicy defines the propagation policy used when deleting a resource
-	PropagationPolicy = metav1.DeletePropagationForeground
+	PropagationPolicy = metav1.DeletePropagationBackground
```
charleszheng44, line 63: Why replace `foreground` with `background`? Shouldn't we want to make sure all dependents are garbage collected? Will `backgound` cause orphan dependents?

hack/lib/release-images.sh
```diff
@@ -83,21 +83,46 @@ function build_multi_arch_binaries() {
...
+                      exit 1
+                 esac
+                 cat << EOF > $docker_file_path
+FROM ${base_image}
+ADD ${binary_name} /usr/local/bin/yurtctl
+EOF
+               else
+                 yurt_component_image="${REPO}/${binary_name}:${TAG}-${arch}"  # openyurt/${binary_name}:latest-${arch}
+                 base_image="k8s.gcr.io/debian-iptables-${arch}:v11.0.2"
+                 cat <<EOF > "${docker_file_path}"
```
charleszheng44, line 120: nit: This change looks like a generic refactoring, which can be submitted in an independent PR.

charleszheng44 (changes requested):

pkg/yurtctl/constants/constants.go
Peeknut: The operations need to modify the files on the Convert/RevertServant Pod host, so we need to mount yurtctl to the host and run it in the host environment. If run yurtctl directly in the container, the content of the file in the pod is modified.

pkg/yurtctl/util/kubernetes/util.go
Peeknut: `foreground` needs to wait for all dependents to be collected before deleting the owner.

In kind test, it may take too long to collect the dependents. When the next convert is executed, the resources created last time still exist, so the conversion cannot be performed this time.

`background` will delete the owner first, and then the garbage collector will delete its dependencies in the background.

hack/lib/release-images.sh
Peeknut: This change is related to the changes in the pkg/yurtctl/constants/constants.go, maybe it's better in a pr?

rambohe-ch (approved):

pkg/yurtctl/constants/constants.go
charleszheng44: That makes sense. Thanks for explaining

charleszheng44 (approved):

PR #236, opened 2021-03-16
Proposal: managing edge devices by integraing Edgex Foundry into OpenYurt

charleszheng44:
This PR adds the proposal of integrating EdgeX into OpenYurt

NONE

NONE

NONE

NONE


yixingjia:
> The implementation of "Custom Resource Definition(CRD) + EdgeX" proposal is excellent. I like it very much. :)
> However, before deep-diving into the implementation detail, I'd like to ask some questions.
> 
>     1. What is the ultimate goal of the Proposal? It is to integrate EdgeX with OpenYurt or to add Device Management support into OpenYurt?
> 
>     2. If the main goal is to support Device Management, what is the reason as well as the supporting data to pick up the K8S CRD solution? Is there any other solution to supporting Device Management? what are their Pros and Cons?
> 
>     3. If we can conclude that CRD is a reasonable solution, do we really have to involve EdgeX to support DeviceMangement? Is there any other solution?
> 
>     4. EdgeX Device Profile/Device/DeviceProfile concept to manage devices is perfect. But I am not sure whether we do need to depend on EdgeX to implement the feature?

Hi Walnux, Thanks a lot for you comments/suggestions!

Let me try to answer some of the questions with my understanding.

This proposal is try to define a reference implementation on how to management edge device in Kubernetes style.  

In K8S scenario we can treat Devices as a kind of resources, so it's quite straight forward to use CRD/CR to represent devices and manage their life cycle. 

The overall design of K8S and Open yurt is quite loosely coupled. Technically it can work with any other edge device management solutions. EdgeX will be the first reference implementation because it is more mature compare with other open source IoT edge device management solutions with pretty active community and contributors.  The communication between K8S and EdgeX here is through standard RestFull API which is loosely coupled.

If you check the Device manager design logic in OpenYurt you will find out that it is a CRD <-> Command translator.  
In the future when you prefer to use other solutions you can easily create another command translator. 

Again this proposal is still in the discussion phase, everything can be discussed !

yixingjia:
> Here, actually, the devices can be accessed/operated through OpenYurt Device Manger and EdgeX Command Service. to the EdgeX based application the Device Manger might like a backdoor to operate the devices, so how to sync the users who access/operate device from OpenYurt Device Manger and EdgeX command Service is a problem.
> In most cases, the connected devices might be very sensitive resources, and shouldn't be uploaded to the cloud(Api Server) or remoted accessed/operated. So using this solution, we have to carefully think about how to isolate and protect these devices resources.

Nice point!  the permession of the devices will depend on the native K8s autha/authz methodology. Only people with enough privileges will be able to see the device, beside that the master node not necessarily placed on the public cloud, it can be in on-prem datacenter or a fog server which depend on your scenario.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+    // Associated Device Service - One per device
+    Service string `json:"service"`
+    // Associated Device Profile - Describes the device
+    Profile string `json:"profile"`
+    // DeviceProperties defines the desired states of device properties
+    DeviceProperties map[string]DeviceProperty `json:"deviceProperties,omitempty"`
+}
+
+// DeviceStatus defines the observed state of Device
+type DeviceStatus struct {
```
yixingjia, line 242: Do we need to consider more device status like the admin status and operation status?

some case like you just install a device but not ready for usage. 

https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/adminstate.go
https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/operatingstate.go

Those status definition is pretty common in IOT solutions.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+
+#### 2. CRUD objects on EdgeX Foundry
+
+To conciliate the device states between OpenYurt and EdgeX Foundry, a new component, DeviceManager, will be installed on the OpenYurt. The DeviceManager includes three controllers, i.e., DeviceProfile Controller, DeviceService Controller, and the Device Controller, which act as mediators between the OpenYurt and the EdgeX Foundry and are responsible for reconciling the states of the device-related custom resources with the states of the corresponding objects on the EdgeX Foundry.
+
+Following is the process of connecting a new device to the OpenYurt through EdgeX Foundry.
+
+![connect-device](../img/connect-device.png)
+
+The vendor applies the DeviceProfile CR.
```
yixingjia, line 341: missing 1. ?

Maybe we can add a blank line between each item, otherwise seems the MD document render will mix them together.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+2. The DeviceManager creates the related ValueDescriptor on the EdgeX MetaData Service.
+3. The DeviceManager create the DeviceProfile on the EdgeX CoreData Service.
+4. The end-user applies the Device CR.
+5. The DeviceManager creates the Device object on the EdgeX CoreData Service.
```
yixingjia, line 345: it should be metadata service, the core data is used to store the sensor data.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+
+To conciliate the device states between OpenYurt and EdgeX Foundry, a new component, DeviceManager, will be installed on the OpenYurt. The DeviceManager includes three controllers, i.e., DeviceProfile Controller, DeviceService Controller, and the Device Controller, which act as mediators between the OpenYurt and the EdgeX Foundry and are responsible for reconciling the states of the device-related custom resources with the states of the corresponding objects on the EdgeX Foundry.
+
+Following is the process of connecting a new device to the OpenYurt through EdgeX Foundry.
+
+![connect-device](../img/connect-device.png)
+
+The vendor applies the DeviceProfile CR.
+2. The DeviceManager creates the related ValueDescriptor on the EdgeX MetaData Service.
+3. The DeviceManager create the DeviceProfile on the EdgeX CoreData Service.
```
yixingjia, line 343: it should be the metadata service.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+4. The end-user applies the Device CR.
+5. The DeviceManager creates the Device object on the EdgeX CoreData Service.
+6. The Device supported commands are created on the EdgeX Command Service.
```
yixingjia, line 346: The command service is stateless, it doesn't store any data, it will read data from the metadata service and then do the command translation if I remember correctly .

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+7. The end-user applies the DeviceService CR.
```
yixingjia, line 347: the device service should be created before user can create a device. since device will reference the device service.

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+  name: testsensor
+spec:
+  ...
+  profile: color-sensor
+  deviceProperties:
+    lightcolor:
+      name: lightcolor
+      setURL: http://edgex-core-command:48082/api/v1/device/<device-id>/command/<put-command-id>
+      desiredValue: green
+status:
```
yixingjia, line 475: A little confused about the "status" section. do you mean the status is the device's current status data?

docs/proposals/20210310-edge-device-management.md
```diff
@@ -0,0 +1,498 @@
...
+      name: lightcolor
+      getURL: http://edgex-core-command:48082/api/v1/device/<device-id>/command/<set-command-id>
+      desiredValue: green
+```
+
+### System Architecture
+
+Edge devices are usually located in the network regions that are unreachable from the cloud, while `DeviceService` needs to talk to edge devices directly. Therefore, we need to deploy the `DeviceService` on the edge node that connects to edge devices. On the other hand, when modifying the device properties, the device controller will need to send a request to the `DeviceService`. Hence, we will deploy a replica of the device controller in each network region. Specifically, the device controller will connect to the YurtHub, and the YurtHub will only pull the device CRDs related to connected edge devices. The system architecture is shown below,
+
+![system-architecture](../img/system-arch.png)
```
yixingjia, line 487: We should consider follow scenario:
EdgeX per network region:  
If we only deploy the Edgex on the cloud then we have an assumption that the network connection between the edge and cloud is quite stable. Since the device service need to push the sensor data to the core data service for persistent and later processing. If we only allow deploy the Edgex on the cloud we should expect that there are data losing during the network breakdown period. 

Beside that deploy the Edgex  near/on the Edge will shorten the latency.  Support multiple EdgeX instances will make it easier for scale out and information isolation.

yixingjia (commented):
a few comments

docs/proposals/20210310-edge-device-management.md
charleszheng44: Would you explain a little bit what is the `operation status` used for?

docs/proposals/20210310-edge-device-management.md
charleszheng44: will add

docs/proposals/20210310-edge-device-management.md
charleszheng44: Yep, I made a mistake. Above mentioned objects should be stored in the metadata service, except the valuedescriptor

docs/proposals/20210310-edge-device-management.md
charleszheng44: will change to metadata service.

docs/proposals/20210310-edge-device-management.md
charleszheng44: Aha, I see. Thanks for clarifying.

docs/proposals/20210310-edge-device-management.md
charleszheng44: 👌

docs/proposals/20210310-edge-device-management.md
charleszheng44: Yes, the current observed status of the device.

docs/proposals/20210310-edge-device-management.md
charleszheng44: Yes. As we discussed offline, we will add a new CRD, EdgeXService (the name can be changed later), which will define the way the EdgeX service is deployed. Also when initiating the EdgeX service, we will synchronize all existing objects (DeviceProfile, DeviceService and Device) on the EdgeX to the OpenYurt.

Fei-Guo (approved):

PR #237, opened 2021-03-17
refactor: divide the yurthubServer into hubServer and proxyServert

rambohe-ch:
Initially, yurtHubServer that listen on 127.0.0.1:10261 handle all requests from clients on edge node, so request can not be proxied to kube-apiserver but instead  handled by yurthub itself when path of request is conflict with local yurthub requests.

in order to solve the above conflict, we divide the yurtHubServer into hubServer and porxyServer, and they listen on different port.
- proxyServer listens on 127.0.0.1:10261,  keep the same port as before
- hubServer listen on 127.0.0.1:10267,  can handle requests like pprof, token update, healthz, metrics etc. 

by the way, we add a parameter `--profiling` to enable/disable debug pprof yurthub, default is enable pprof.

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"net"
 	"net/url"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


Fei-Guo (approved):
LGTM

yixingjia:
Any cons on use a special  prefix-path for yurt-hub related request compare with use a separator port?

eg:  /v1/yurthub/xxx

rambohe-ch:
> @rambohe-ch Any cons on use a special prefix-path for yurt-hub related request compare with use a separator port?
> 
> eg: /v1/yurthub/xxx

@yixingjia At first, YurtHub has used path to distinguish requests, but had ran into problems.
- There are some common path for kube-apiserver and YurtHub, like /debug/pprof, /metrics, so path is not enough to distinguish requests.
- on the other hand, The use of different ports can be more clearly defined function for proxy and hub agent.

PR #265, opened 2021-04-19
refactor cache manager for yurthub

rambohe-ch:
#### Backgound:
1.  list requests from edge node have three types:
 - type1: only get specified object by using `fieldselector=metadata.name=xxx`, the request path like: `https://{ip:port}/api/v1/namespaces/default/pods?fieldselector=metadata.name=xxx`
 - type2: get a kind of objects by using  `fieldselector=spec.nodeName=xxx` or `labelselector=foo=bar`, the request path like: `https://{ip:port}/api/v1/pods?fieldselector=spec.nodeName=xxx`
 - type3: get all pods, the request path like: `https://{ip:port}/api/v1/pods`

2. cache manager handles list request in yurt-hub:
- in yurthub cache manager, we cache object one by one for list response, and the code link: https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_manager.go#L336
- for type2 list request, query cache handle is the same as type3 list request, and the code link: https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_manager.go#L167

#### problem:
- response for type3 list will not update all caches on local disk, so some deleted objects will remain in the cache.
- if a client on edge node has both type2 and type3 list request, a error will occur for type2 response.

#### solution:
1. replace local cache response instead of one by one cache for type2 and type3 list request
2. fix cache conflict for type2 and type3 list request, a client can only support one type list request  for one resource in type2 and type3。and at present, kubelet/kube-proxy/flannel/coredns have only one type list request for one resource in type2 and type3.

make test
make all
and make sure yurthub cache manager can work


pkg/yurthub/storage/disk/storage.go
```diff
@@ -343,6 +343,84 @@ func (ds *diskStorage) Update(key string, contents []byte) error {
...
+			}
+			dirExisted = true
+		}
+	}
+
+	// 2. create new file with contents
+	for key, data := range contents {
+		err := ds.create(key, data)
+		if err != nil {
+			klog.Errorf("failed to create %s in replace, %v", key, err)
```
Fei-Guo, line 384: If error happens, do we have retry mechanism? If not, do we have mechanism to do consistency check?

pkg/yurthub/cachemanager/cache_manager.go
```diff
@@ -587,5 +585,24 @@ func (cm *cacheManager) CanCacheFor(req *http.Request) bool {
 		return false
 	}
 
+	cm.Lock()
+	defer cm.Unlock()
+	if info.Verb == "list" && info.Name == "" {
+		key, _ := util.KeyFunc(comp, info.Resource, info.Namespace, info.Name)
+		if _, ok := cm.listReqCollector[key]; !ok {
+			for k := range cm.listReqCollector {
```
Fei-Guo, line 593: Can you explain the purpose of having listReqCollector? 

Also, can you give an example that "len(k) > len (key) " and an example for "len(k) < len(key)"?

pkg/yurthub/storage/disk/storage.go
rambohe-ch: I have not seriously thought about how to guarantee that object can be cached successfully. so i'd like to keep this subject as a TODO item. and we can solve it in the soon future.

pkg/yurthub/cachemanager/cache_manager.go
rambohe-ch: 1. because func queryListObject() will get all pods for both requests instead of getting pods by request selector. so cache manager can not support same path list requests that has different selector. for example:
  -  request1: http://{ip:port}/api/v1/default/pods?labelSelector=foo=bar
  - request2: http://{ip:port}/api/v1/default/pods?labelSelector=foo2=bar2

2. because func queryListObject() will get all pods for both requests instead of getting pods by request selector. so cache manager can not support getting same resource list requests that has different path. for example:
  - request1: http://{ip/port}/api/v1/pods?fieldSelector=spec.nodeName=foo
  - request2: http://{ip/port}/api/v1/default/pods?fieldSelector=spec.nodeName=foo

in order to solve the above two problems, we need listReqCollector to keep list request info,  and based on listReqCollector to verify list requests can cache or not.

> Also, can you give an example that "len(k) > len (key) " and an example for "len(k) < len(key)"?
 the second above example.

Fei-Guo (approved):

Issue #248, opened 2021-04-01
pkg/yurthub/util/util_test.go:37:4: call to (*T).Fatalf from a non-test goroutine


advancedwebdeveloper:
Hi.
Caught a bug:

> $ make
> go fmt ./pkg/... ./cmd/...
> go fmt ./pkg/... ./cmd/...
> go: downloading k8s.io/apimachinery v0.16.10-beta.0
> go: downloading k8s.io/client-go v0.16.9
> go: downloading github.com/kubernetes/kubernetes v1.16.9
> go: downloading k8s.io/klog v1.0.0
> go: downloading github.com/daviddengcn/go-colortext v0.0.0-20160507010035-511bcaf42ccd
> go: downloading k8s.io/api v0.16.9
> go: downloading github.com/spf13/cobra v1.0.0
> go: downloading k8s.io/apiserver v0.16.9
> go: downloading google.golang.org/grpc v1.26.0
> go: downloading sigs.k8s.io/apiserver-network-proxy v0.0.15
> go: downloading github.com/prometheus/client_golang v0.9.2
> go: downloading github.com/gorilla/mux v1.7.4
> go: downloading k8s.io/component-base v0.16.9
> go: downloading k8s.io/utils v0.0.0-20200603063816-c1c6865ac451
> go: downloading k8s.io/cluster-bootstrap v0.16.9
> go: downloading github.com/google/uuid v1.1.1
> go: downloading k8s.io/klog/v2 v2.0.0
> go: downloading github.com/go-logr/logr v0.1.0
> go: downloading github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
> go: downloading github.com/prometheus/common v0.0.0-20181126121408-4724e9255275
> go: downloading github.com/prometheus/procfs v0.0.11
> go: downloading sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.15
> go: downloading google.golang.org/protobuf v1.23.0
> go: downloading google.golang.org/genproto v0.0.0-20190819201941-24fa4b261c55
> go: downloading golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7
> go: downloading github.com/gogo/protobuf v1.2.2-0.20190723190241-65acae22fc9d
> go: downloading golang.org/x/sys v0.0.0-20200519105757-fe76b779f299
> go: downloading github.com/google/gofuzz v1.1.0
[...]

PR #269, opened 2021-04-20
fix t.Fatalf from a non-test goroutine

contrun:
Fixes https://github.com/openyurtio/openyurt/issues/248.


SataQiu (commented):
/lgtm
@rambohe-ch

rambohe-ch (approved):
LGTM

PR #270, opened 2021-04-20
feature: add dns controller to sync cluster node dns records

SataQiu:
In the edge computing scenario, the IP addresses of the edge nodes are likely to be the same. So we can not rely on the node IP to forward the request but should use the node hostname(unique in one cluster). 
This PR provides the ability for the tunnel-server to handle requests in the form of `scheme://[hostname]:[port]/[req_path]`.

### How to use it
The dns controller will generate a ConfigMap named `yurt-tunnel-nodes` in the `kube-system` namespace.
You need to manually config CoreDNS to use it according to the steps like the following:
First, mount the ConfigMap as volume of CoreDNS Pod, for example
```yaml
        volumeMounts:
        - mountPath: /etc/edge
          name: hosts
          readOnly: true
...

      volumes:
      - configMap:
          defaultMode: 420
          name: yurt-tunnel-nodes
        name: hosts
```
Second, update CoreDNS ConfigMap (`kube-system/coredns`), add the `hosts` section
```yaml
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health :10260
        hosts /etc/edge/tunnel-nodes {
            reload 300ms
            fallthrough
        }
        kubernetes cluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream
           fallthrough in-addr.arpa ip6.arpa
           endpoint http://169.254.2.1:10261
        }
        prometheus :9153
        proxy . /etc/resolv.conf
        cache 30
        reload
    }
kind: ConfigMap
metadata:
  name: coredns
  namespace: kube-system
```

### DIFF
--- a/config/setup/yurt-tunnel-server.yaml
+++ b/config/setup/yurt-tunnel-server.yaml
@@ -127,6 +127,7 @@ spec:
         - yurt-tunnel-server
         args:
         - --bind-address=$(NODE_IP)
+        - --insecure-bind-address=$(NODE_IP)
         - --proxy-strategy=destHost
         - --v=2
         env:
--- a/pkg/yurttunnel/dns/dns.go
+++ b/pkg/yurttunnel/dns/dns.go
@@ -1,0 +1,462 @@
+/*
+Copyright 2021 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package dns
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	corelister "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	"github.com/openyurtio/openyurt/pkg/yurttunnel/constants"
+	"github.com/openyurtio/openyurt/pkg/yurttunnel/util"
+)
+
+const (
+	maxRetries    = 15
+	minSyncPeriod = 30
+
+	yurttunnelDNSRecordConfigMapNs = "kube-system"
+	yurttunnelDNSRecordNodeDataKey = "tunnel-nodes"
+
+	dnatPortPrefix = "dnat-"
+)
+
+var (
+	yurttunnelDNSRecordConfigMapName = fmt.Sprintf("%s-tunnel-nodes",
+		strings.TrimRightFunc(projectinfo.GetProjectPrefix(), func(c rune) bool { return c == '-' }))
+)
+
+// DNSRecordController interface defines the method for synchronizing
+// the node dns records with k8s DNS component(such as CoreDNS)
+type DNSRecordController interface {
+	Run(stopCh <-chan struct{})
+}
+
+// coreDNSRecordController implements the DNSRecordController
+type coreDNSRecordController struct {
+	lock                 sync.Mutex
+	kubeClient           clientset.Interface
+	sharedInformerFactor informers.SharedInformerFactory
+	nodeLister           corelister.NodeLister
+	nodeListerSynced     cache.InformerSynced
+	svcInformerSynced    cache.InformerSynced
+	cmInformerSynced     cache.InformerSynced
+	queue                workqueue.RateLimitingInterface
+	tunnelServerIP       string
+	insecurePort         int
+	syncPeriod           int
+}
+
+// NewCoreDNSRecordController create a CoreDNSRecordController that synchronizes node dns records with CoreDNS configuration
+func NewCoreDNSRecordController(client clientset.Interface,
+	informerFactory informers.SharedInformerFactory,
+	listenInsecureAddr string,
+	syncPeriod int) (DNSRecordController, error) {
+
+	_, insecurePortStr, err := net.SplitHostPort(listenInsecureAddr)
+	if err != nil {
+		return nil, err
+	}
+
+	insecurePort, err := strconv.Atoi(insecurePortStr)
+	if err != nil {
+		return nil, err
+	}
+
+	dnsctl := &coreDNSRecordController{
+		kubeClient:           client,
+		syncPeriod:           syncPeriod,
+		insecurePort:         insecurePort,
+		sharedInformerFactor: informerFactory,
+		queue:                workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "tunnel-dns"),
+	}
+
+	nodeInformer := informerFactory.Core().V1().Nodes()
+	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    dnsctl.addNode,
+		UpdateFunc: dnsctl.updateNode,
+		DeleteFunc: dnsctl.deleteNode,
+	})
+	dnsctl.nodeLister = nodeInformer.Lister()
+	dnsctl.nodeListerSynced = nodeInformer.Informer().HasSynced
+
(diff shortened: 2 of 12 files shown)


pkg/yurttunnel/dns/node.go
```diff
@@ -0,0 +1,210 @@
...
+// getNodeHostIP returns the provided node's "primary" IP
+func getNodeHostIP(node *v1.Node) (string, error) {
+	// Re-sort the addresses with InternalIPs first and then ExternalIPs
+	allIPs := make([]net.IP, 0, len(node.Status.Addresses))
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == v1.NodeInternalIP {
+			ip := net.ParseIP(addr.Address)
+			if ip != nil {
+				allIPs = append(allIPs, ip)
+			}
```
rambohe-ch, line 194: how about add a break to exit for loop?

pkg/yurttunnel/dns/node.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+	}
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == v1.NodeExternalIP {
+			ip := net.ParseIP(addr.Address)
+			if ip != nil {
+				allIPs = append(allIPs, ip)
+			}
```
rambohe-ch, line 202: how about add a break to exit for loop?

pkg/yurttunnel/dns/node.go
```diff
@@ -0,0 +1,210 @@
...
+
+// onNodeAdd processes node add or update event
+func (dnsctl *coreDNSRecordController) onNodeAdd(node *corev1.Node) error {
+	klog.V(2).Infof("adding node dns record for %v", node.Name)
+	ip, err := getNodeHostIP(node)
+	if err != nil {
+		return err
+	}
+	if isEdgeNode(node) {
+		ip, err = dnsctl.getTunnelServerIP()
```
rambohe-ch, line 97: how about get clusterIP of kube-system/x-tunnel-server-svc service only once instead of getting clusterIP per nodeEvent?

pkg/yurttunnel/dns/node.go
```diff
@@ -0,0 +1,210 @@
...
+		_, err := dnsctl.kubeClient.CoreV1().ConfigMaps(constants.YurttunnelServerServiceNs).Update(cm)
+		return err
+	}
+	var buff bytes.Buffer
+	buff.WriteString(record)
+	buff.WriteString(data)
+	cm.Data[yurttunnelDNSRecordNodeDataKey] = buff.String()
+	if _, err := dnsctl.kubeClient.CoreV1().ConfigMaps(constants.YurttunnelServerServiceNs).Update(cm); err != nil {
+		return err
+	}
```
rambohe-ch, line 138: how about refactor the above routine as follows?
```
	if regex.FindString(data) != "" {
		updatedData := regex.ReplaceAllString(data, record)
		if updatedData == data {
			return nil
		}
		cm.Data[yurttunnelDNSRecordNodeDataKey] = updatedData
	} else {
	      var buff bytes.Buffer
	      buff.WriteString(record)
	      buff.WriteString(data)
	      cm.Data[yurttunnelDNSRecordNodeDataKey] = buff.String()
       }
	_, err := dnsctl.kubeClient.CoreV1().ConfigMaps(constants.YurttunnelServerServiceNs).Update(cm); err != nil {
       return err
```

pkg/yurttunnel/dns/dns.go
```diff
@@ -0,0 +1,301 @@
...
+	if err != nil {
+		klog.Errorf("failed to sync dns record as whole, %v", err)
+		return
+	}
+
+	var buff bytes.Buffer
+	for _, node := range nodes {
+		if isEdgeNode(node) {
+			buff.WriteString(fmt.Sprintf("%s\t%s\n", ip, node.Name))
+		}
```
rambohe-ch, line 273: cloud nodes are missed?

pkg/yurttunnel/dns/service.go
```diff
@@ -0,0 +1,41 @@
...
+package dns
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/yurttunnel/constants"
+)
+
+func (dnsctl *coreDNSRecordController) addService(obj interface{}) {
```
rambohe-ch, line 26: why need to list/watch all of service for syncing dns records?

pkg/yurttunnel/dns/node.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func (dc *coreDNSRecordController) onNodeDelete(node *corev1.Node) error {
```
rambohe-ch, line 142: how about extract common routine in func onNodeDelete and func onNodeAdd?

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/onsi/ginkgo v1.13.0
 	github.com/onsi/gomega v1.10.1
 	github.com/opencontainers/go-digest v1.0.0 // indirect
+	github.com/pkg/errors v0.8.1
```
rambohe-ch, line 19: why import pkg/errors?

rambohe-ch:
@SataQiu Thank you for making the pull request.
would you add more detail info like background, reason, solution etc to explain why we need this pull request?

pkg/yurttunnel/dns/dns.go
```diff
@@ -0,0 +1,463 @@
...
+	for i := range nodes {
+		node := nodes[i]
+		if !isEdgeNode(node) {
+			ip, err = getNodeHostIP(node)
+			if err != nil {
+				klog.Errorf("failed to parse node address for %v, %v", node.Name, err)
+				continue
+			}
+		}
+		records = append(records, formatDNSRecord(ip, node.Name))
```
rambohe-ch, line 356: records for edge nodes are frogotten?

pkg/yurttunnel/dns/dns.go
```diff
@@ -0,0 +1,463 @@
...
+
+func (dnsctl *coreDNSRecordController) updateTunnelServerSvcDnatPorts(ports []string) error {
+	svc, err := dnsctl.kubeClient.CoreV1().Services(constants.YurttunnelServerServiceNs).
+		Get(constants.YurttunnelServerServiceName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to sync tunnel server service, %v", err)
+	}
+
+	// parse tunnel server tcp port
+	tunnelServerAgentPort, _ := strconv.Atoi(constants.YurttunnelServerAgentPort)
```
rambohe-ch, line 411: tunnelServerAgentPort for insecure request is not 10262,  is 10264(YurttunnelServerMasterInsecurePort)

pkg/yurttunnel/dns/util.go
```diff
@@ -0,0 +1,115 @@
...
+	result = make([]string, len(records))
+	copy(result, records)
+
+	found := false
+	for i, v := range result {
+		if strings.HasSuffix(v, hostname) {
+			found = true
+			if v != record {
+				result[i] = record
+				changed = true
```
rambohe-ch, line 104: how about add a break to exit for loop?

pkg/yurttunnel/util/util.go
```diff
@@ -33,3 +50,43 @@ func RunMetaServer(addr string) {
...
+		Get(YurttunnelServerDnatConfigMapName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(4).Infof("configmap %s/%s is not found",
+				YurttunnelServerDnatConfigMapNs,
+				YurttunnelServerDnatConfigMapName)
+		} else {
+			klog.Errorf("fail to get configmap %s/%s: %v",
+				YurttunnelServerDnatConfigMapNs,
+				YurttunnelServerDnatConfigMapName, err)
```
rambohe-ch, line 69: I think that we need to add error return value for func `GetConfiguredDnatPorts`, and when we failed to get yurt-server-tunnel-cfg configmap, we can return error instead of empty slice.

pkg/yurttunnel/dns/dns.go
```diff
@@ -0,0 +1,463 @@
...
+	go wait.Until(dnsctl.worker, time.Second, stopCh)
+
+	// sync dns hosts as a whole
+	go wait.Until(dnsctl.syncDNSRecordAsWhole, time.Duration(dnsctl.syncPeriod)*time.Second, stopCh)
+
+	// sync tunnel server svc
+	go wait.Until(func() {
+		if err := dnsctl.syncTunnelServerService(); err != nil {
+			klog.Errorf("failed to sync tunnel server service, %v", err)
+		}
```
rambohe-ch, line 226: why we need to add syncTunnelServerService here?

SataQiu: Based on past experience, we should not rely solely on the `watch` mechanism, as there is a risk of lost events. We need to make sure that the information is synchronized through a periodic `list` function.

pkg/yurttunnel/dns/dns.go
SataQiu: We will record whether it is an edge node or not, but the destination ip is different.

config/setup/yurt-tunnel-server.yaml
```diff
@@ -127,6 +127,7 @@ spec:
         - yurt-tunnel-server
         args:
         - --bind-address=$(NODE_IP)
+        - --insecure-bind-address=$(NODE_IP)
```
SataQiu, line 130: Ensure that the tunnel-server listening address can be accessed by other cloud nodes in the cluster.

rambohe-ch: yeah, tunnel-server need to listen on node ip for insecure requests.

pkg/yurttunnel/dns/dns.go
```diff
@@ -0,0 +1,454 @@
...
+
+	updatedSvcPorts := make([]corev1.ServicePort, 0, len(svc.Spec.Ports))
+	for i := range svc.Spec.Ports {
+		port := svc.Spec.Ports[i]
+		if strings.HasPrefix(port.Name, dnatPortPrefix) && !dnatPortMap[int(port.Port)] {
+			continue
+		}
+		updatedSvcPorts = append(updatedSvcPorts, port)
+	}
+	svc.Spec.Ports = updatedSvcPorts
```
rambohe-ch, line 447: how about add a flag to check service port need to be updated or not?

rambohe-ch (approved):
LGTM

PR #283, opened 2021-04-27
bugfix: fix the bug that dns controller updates dns records incorrectly

SataQiu:

Follow up #270
bugfix: fix the bug that dns controller updates dns records incorrectly

/cc @rambohe-ch


rambohe-ch (approved):
LGTM

PR #284, opened 2021-04-27
enhancement: add tunnel server internal service in order to prevent x-tunnel-server-svc attached SLB to listen unsecure port

rambohe-ch:
- background: 
   1. In order to deploy promtheus(also metrics-server) and yurt-tunnel-server on different cloud nodes,  prometheus need to use  {hostname:port} (take port=9100 as a example) to access edge node through yurt-tunnel-server and yurt-tunnel-agent and resolve hostname to service ip of yurt-tunnel-server. on the same time, we need to add 9100:10264 mapping info in yurt-tunnel-server service for redirect port.
   2. the original yurt-tunnel-server service(`x-tunnel-server-svc`) is used by yurt-tunnel-agent to connect yurt-tunnel-server. and `x-tunnel-server-svc` service may be configured as NodePort type or LoadBalancer type for public network access. 
   3. If `x-tunnel-server-svc` service type is LoadBalancer, the proxy ports  in `x-tunnel-service-svc` service will be listened by corresponding SLB（in public cloud. but listen internal port(9100) on public network will cause security risks.

- solution:
 we can add a new service(type=ClusterIP) named `x-tunnel-server-internal-svc` for yurt-tunnel-server, so `x-tunnel-server-svc` service is used for public network access,  and newly added service is used for internal component(like prometheus/metrics server) access. and internal port mapping(like above 9100:10264) will only be added in `x-tunnel-server-internal-svc` svc. so security risk can be solved.

make all
and make sure the yurt-tunnel can work with dns

### DIFF
--- a/config/setup/yurt-tunnel-server.yaml
+++ b/config/setup/yurt-tunnel-server.yaml
@@ -81,6 +81,24 @@ spec:
     k8s-app: yurt-tunnel-server
 ---
 apiVersion: v1
+kind: Service
+metadata:
+  name: x-tunnel-server-internal-svc
+  namespace: kube-system
+  labels:
+    name: yurt-tunnel-server
+spec:
+  ports:
+    - port: 10250
+      targetPort: 10263
+      name: https
+    - port: 10255
+      targetPort: 10264
+      name: http
+  selector:
+    k8s-app: yurt-tunnel-server
+---
+apiVersion: v1
 kind: ConfigMap
 metadata:
   name: yurt-tunnel-server-cfg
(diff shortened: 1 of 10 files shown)


SataQiu (commented):
LGTM
defer @Fei-Guo for more reviews

Fei-Guo:
The major use case of keeping the x-tunnel-server-svc is to allow the tunnel-agent (sit in edge node) to find the tunnel-server (sit in cloud) through the public IP (i.e., the node public IP) + nodeport.

Fei-Guo (approved):

PR #287, opened 2021-04-29
bugfix: keep image tar file name consistent with image name

SataQiu:
bugfix: keep image tar file name consistent with image name.

For example, the image `openyurt/yurtctl-servant:v0.3.0-amd64` should be saved as `yurtctl-servant-linux-amd64.tar` instead of `yurtctl-linux-amd64.tar`.

make release


Fei-Guo (commented):
Good catch.

Fei-Guo (approved):

PR #288, opened 2021-05-02
【feature】adapt openyurt for k8s version v1.18

rambohe-ch:
adapt openyurt for k8s version v1.18, and the main work as following:
1. the version of k8s.io libs(like client-go) were upgraded to v1.18.8
2. add `context` parameter in api for calling client-go v1.18
3. refactor serializer of cache manager to adapt ClientNegotiator in k8s.io/apimachinery v1.18

make test
make build
and make sure all components can work correctly with k8s v1.18 cluster


Fei-Guo:
LGTM, leave to @kadisi for another pass.

PR #289, opened 2021-05-06
[feature]: add dummy network interface for yurthub

rambohe-ch:
- background:
   Yurthub is a node daemon that serves as a proxy for the outbound traffic from the Kubernetes node daemons(like kubelet, kube-proxy and so on), and yurthub is listening on `127.0.0.1:10261` for proxying outbound traffic. so components with hostNetwork=false can not use yurthub to proxy outbound request because of traffic from `container network namespace` to `127.0.0.1` can not route on edge node.

- solution:
   Add a dummy network interface on edge node by yurthub, and yurthub listens on `new dummy interface`, so components in its own network namespace can access yurthub through dummy interface. 
   the default ip address of new dummy interface is: `169.254.2.1`, the interface name is: `yurthub-dummy0`

- usage for components on edge node:
  to access kube-apiserver through yurthub for components, at first make sure that components can support to set yurthub address instead of kube-apiserver address. and the address of yurthub as following:
   1.  yurthub address for components on host network:  http://127.0.0.1:10261
   2. yurthub address for components on dependent network namespace:  http://169.254.2.1:10261

make all
and make sure yurthub can work correctly.

### DIFF
--- a/cmd/yurthub/app/start.go
+++ b/cmd/yurthub/app/start.go
@@ -29,6 +29,7 @@ import (
 	"github.com/openyurtio/openyurt/pkg/yurthub/gc"
 	"github.com/openyurtio/openyurt/pkg/yurthub/healthchecker"
 	"github.com/openyurtio/openyurt/pkg/yurthub/kubernetes/serializer"
+	"github.com/openyurtio/openyurt/pkg/yurthub/network"
 	"github.com/openyurtio/openyurt/pkg/yurthub/proxy"
 	"github.com/openyurtio/openyurt/pkg/yurthub/server"
 	"github.com/openyurtio/openyurt/pkg/yurthub/storage/factory"
@@ -157,10 +158,27 @@ func Run(cfg *config.YurtHubConfiguration, stopCh <-chan struct{}) error {
 	}
 	trace++
 
+	if cfg.EnableDummyIf {
+		klog.Infof("%d. create dummy network interface %s and init iptables manager", trace, cfg.HubAgentDummyIfName)
+		networkMgr, err := network.NewNetworkManager(cfg)
+		if err != nil {
+			klog.Errorf("could not create network manager, %v", err)
+			return err
+		}
+		networkMgr.Run(stopCh)
+		trace++
+		klog.Infof("%d. new %s server and begin to serve, dummy proxy server: %s", trace, projectinfo.GetHubName(), cfg.YurtHubProxyServerDummyAddr)
+	}
+
 	klog.Infof("%d. new %s server and begin to serve, proxy server: %s, hub server: %s", trace, projectinfo.GetHubName(), cfg.YurtHubProxyServerAddr, cfg.YurtHubServerAddr)
-	s := server.NewYurtHubServer(cfg, certManager, yurtProxyHandler)
+	s, err := server.NewYurtHubServer(cfg, certManager, yurtProxyHandler)
+	if err != nil {
+		klog.Errorf("could not create hub server, %v", err)
+		return err
+	}
 	s.Run()
 
+	klog.Infof("hub agent exited")
 	<-stopCh
 	return nil
 }
--- a/pkg/yurthub/network/dummyif_linux.go
+++ b/pkg/yurthub/network/dummyif_linux.go
@@ -1,0 +1,123 @@
+/*
+Copyright 2021 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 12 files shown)


zyjhtangtang:
lgtm

charleszheng44:
Just have a question, on the edge node, except kubelet, which component will need to access yurthub?

rambohe-ch:
> Just have a question, on the edge node, except kubelet, which component will need to access yurthub?

any components(like kubelet, kube-proxy, flannel and so on) that want to use yurthub cache when cloud-edge network disconnected.

pkg/yurthub/network/dummyif_unsupported.go
```diff
@@ -0,0 +1,51 @@
...
+type DummyInterfaceController interface {
+	EnsureDummyInterface(ifName string, ifIP net.IP) error
+	DeleteDummyInterface(ifName string) error
+	ListDummyInterface(ifName string) ([]net.IP, error)
+}
+
+type unsupportedInterfaceController struct {
+}
+
+func NewDummyInterfaceController() DummyInterfaceController {
```
Fei-Guo, line 34: Should NewDummyInterfaceController be NewUnsupportedDummyInterfaceController?

Where is the unsupportedDummyController used?

pkg/yurthub/network/dummyif_linux.go
```diff
@@ -0,0 +1,117 @@
...
+
+		for _, addr := range addrs {
+			if addr.IP != nil && addr.IP.Equal(ifIP) {
+				return nil
+			}
+		}
+
+		klog.Infof("ip address for %s interface changed to %s", ifName, ifIP.String())
+		return dic.AddrReplace(link, &netlink.Addr{IPNet: netlink.NewIPNet(ifIP)})
+	}
```
Fei-Guo, line 62: if err != nil, why not returning err but calling addDummyInterface?  Maybe a comment is needed.

cmd/yurthub/app/start.go
```diff
@@ -157,10 +158,28 @@ func Run(cfg *config.YurtHubConfiguration, stopCh <-chan struct{}) error {
...
+	klog.Infof("%d. create dummy network interface %s and init iptables manager", trace, cfg.HubAgentDummyIfName)
+	networkMgr, err := network.NewNetworkManager(cfg)
+	if err != nil {
+		klog.Errorf("could not create network manager, %v", err)
+		return err
+	}
+	networkMgr.Run(stopCh)
+	trace++
+
+	klog.Infof("%d. new %s server and begin to serve, proxy server: %s;%s, hub server: %s", trace,
```
Fei-Guo, line 170: You should add "dummy proxy server" before "%s" in the log line.

cmd/yurthub/app/start.go
```diff
@@ -157,10 +158,28 @@ func Run(cfg *config.YurtHubConfiguration, stopCh <-chan struct{}) error {
 	}
 	trace++
 
-	klog.Infof("%d. new %s server and begin to serve, proxy server: %s, hub server: %s", trace, projectinfo.GetHubName(), cfg.YurtHubProxyServerAddr, cfg.YurtHubServerAddr)
-	s := server.NewYurtHubServer(cfg, certManager, yurtProxyHandler)
+	klog.Infof("%d. create dummy network interface %s and init iptables manager", trace, cfg.HubAgentDummyIfName)
```
Fei-Guo, line 161: I think the entire change in this file should be guarded by "if cfg.EnableDummyIf ".    This change should introduce zero overhead if people does not need/disable this feature.

pkg/yurthub/network/dummyif_unsupported.go
rambohe-ch: when os is not linux, unsupportedDummyController will be used by NetworkManager, and when os is linux, dummyif_linux.go will be used.

pkg/yurthub/network/dummyif_linux.go
rambohe-ch: yeah, add error handle here, if err is `Link not found` ,  we will call addDummyInterface, otherwise return err.

cmd/yurthub/app/start.go
rambohe-ch: fixed

Fei-Guo (approved):
LGTM

felix0080:
@rambohe-ch hello rambohe, I'd like to consult you about..  What does iptable do in your commit file

PR #302, opened 2021-05-18
add edge-pod-network doc

wenjun93:
add doc for edge pod network.


rambohe-ch:
@wenjun93 Thank you for adding the doc of edge pod network. and how about add the following the contents:
1.  the background of the pod network autonomy 
2. the problems for making sure pod network autonomy

and then explain the solutions of problems. 
by the way, It is easier to understand if an arch pic of edge pod network can be added.

wenjun93:
> @wenjun93 Thank you for adding the doc of edge pod network. and how about add the following the contents:
> 
> 1. the background of the pod network autonomy
> 2. the problems for making sure pod network autonomy
> 
> and then explain the solutions of problems.
> by the way, It is easier to understand if an arch pic of edge pod network can be added.

updated

Fei-Guo:
1) Please fix the lint errors
2) Is this possible to show the code diff if the patch is simple?  

Especially the local IPAM plugin change is requested by many users many times. It will be nice to provide more details so that user can DIY.

wenjun93:
@Fei-Guo The doc proposed 

> 1. Please fix the lint errors
> 2. Is this possible to show the code diff if the patch is simple?
> 
> Especially the local IPAM plugin change is requested by many users many times. It will be nice to provide more details so that user can DIY.

the patch is not simply a few lines of code change, we might discuss it.

Fei-Guo (approved):

PR #349, opened 2021-06-11
change cache path from fixed to configurable parameter

### LINKED ISSUE #343: [feature request]metadata cache path can be configured
The k8s's resource metadata are saved to the fixed path `/etc/kubernetes/cache/`.  I think that it's better to change it according to bootstrap parameter.

**Why is this needed**:
Maybe some user want to save metadata to specified path. 

**others**
/kind feature

luckymrwang:
/kind feature

Fixes #343

```release-note
NONE
```


Fei-Guo:
please fix the go-lint error.

luckymrwang:
> please fix the go-lint error.

@Fei-Guo I have fixed the go-lint error.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -30,8 +30,7 @@ import (
 )
 
 const (
-	cacheBaseDir = "/etc/kubernetes/cache/"
```
rambohe-ch, line 33: I think we need keep the `cacheBaseDir` as default path for disk storage. if input parameter `dir` is empty, so we can use the default path.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -42,9 +41,6 @@ type diskStorage struct {
 
 // NewDiskStorage creates a storage.Store for caching data into local disk
 func NewDiskStorage(dir string) (storage.Store, error) {
-	if dir == "" {
```
rambohe-ch, line 45: It's need to check the validity of input parameter. so please do not delete it.

Fei-Guo: Actually, nil check is not enough for validity if we allow user input directory. We need to check more such as the directory is at least available.

pkg/yurthub/storage/disk/storage.go
luckymrwang: `DiskCachePath` has been defined the default path `/etc/kubernetes/cache` to backward compatibility in the `cmd/yurthub/app/options/options.go`.

luckymrwang: > I think we need keep the `cacheBaseDir` as default path for disk storage. if input parameter `dir` is empty, so we can use the default path.

`DiskCachePath` has been defined default path `/etc/kubernetes/cache/` to `backward compatibility` in the `cmd/yurthub/app/options/options.go`.

pkg/yurthub/storage/disk/storage.go
luckymrwang: > Actually, nil check is not enough for validity if we allow user input directory. We need to check more such as the directory is at least available.

@Fei-Guo The following code can cover it what the `dir` is empty or not valid path.

Fei-Guo:
I think it is ok for now that if people explicitly specify the "" path then we fail it.
/lgtm

pkg/yurthub/storage/disk/storage.go
luckymrwang: > It's need to check the validity of input parameter. so please do not delete it.

I have fixed it

luckymrwang:
> I think it is ok for now that if people explicitly specify the "" path then we fail it.
> /lgtm

I changed the previous code which juded `dir` @rambohe-ch referred to. Please review it again.

PR #350, opened 2021-06-11
feature: set tunnel server count dynamically when --server-count flag is not explicitly specified by the user

SataQiu:
#### What type of PR is this?
/kind feature

feature: set tunnel server count dynamically when --server-count flag is not explicitly specified by the user

Fixes #

```release-note
feature: set tunnel server count dynamically when --server-count flag is not explicitly specified by the user
```

### DIFF
--- a/cmd/yurt-tunnel-server/app/options/options.go
+++ b/cmd/yurt-tunnel-server/app/options/options.go
@@ -17,8 +17,10 @@ limitations under the License.
 package options
 
 import (
+	"context"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/yurt-tunnel-server/app/options/options.go
```diff
@@ -152,6 +157,55 @@ func (o *ServerOptions) Config() (*config.Config, error) {
...
+		}
+	}
+
 	klog.Infof("yurttunnel server config: %#+v", cfg)
 	return cfg, nil
 }
+
+// guessServerCount returns the replicas of the tunnel server pod
+// owner ReplicaSet if that is available
+func guessServerCount(client kubernetes.Interface) (int, error) {
```
Fei-Guo, line 177: guessServerCount is not a good name.  How about tryObtainServerCountFromReplicaSet?

rambohe-ch:
/assign @wawlian 
Do you have any comments about `--server-count` of yurt-tunnel

cmd/yurt-tunnel-server/app/options/options.go
SataQiu: Updated!

SataQiu:
/close
I prefer to implement the control logic through an `edge-operator` :)

PR #354, opened 2021-06-15
add tolerations for yurt-tunnel-agent in order to avoid confict with nodepool taints

cuirunxing-hub:
add tolerations for yurt-tunnel-agent in order to avoid confict with nodepool taints

#### What type of PR is this?
 /kind bug


rambohe-ch:
@cuirunxing-hub please update the following files in the meantime.
https://github.com/openyurtio/openyurt/blob/master/config/setup/yurt-tunnel-agent.yaml
https://github.com/openyurtio/openyurt/blob/master/config/yaml-template/yurt-tunnel-agent.yaml

PR #369, opened 2021-06-24
Add StorageWrapper test case

MIBc:


pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,135 @@
...
+				Labels: map[string]string{
+					"tag": "test",
+				},
+			},
+		})
+		err = sWrapper.Update("kubelet/pods/default/mypod1", updatePod)
+		if err != nil {
+			t.Errorf("failed to update obj, %v", err)
+		}
+	})
```
qclc, line 94: How about verifying the label of the updated Pod here to ensure that the update is successful?

pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			t.Errorf("failed to replace objs, %v", err)
+		}
+	})
+
+	t.Run("Test delete storage", func(t *testing.T) {
+		err = sWrapper.Delete("kubelet/pods/default/mypod1")
+		if err != nil {
+			t.Errorf("failed to delete obj, %v", err)
+		}
```
qclc, line 133: I think it would be better to add a step to determine if the `mypod1` file exists on disk after the delete operation.

MIBc: done

pkg/yurthub/cachemanager/storage_wrapper_test.go
MIBc: done

pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	})
+
+	t.Run("Test list obj and keys", func(t *testing.T) {
+		_, err = sWrapper.ListKeys("kubelet/pods/default")
+		if err != nil {
+			t.Errorf("failed to list keys, %v", err)
+		}
+		_, err = sWrapper.List("kubelet/pods/default")
+		if err != nil {
```
qclc, line 117: For a List request, if you get kubelet/pods, and there are no pods in the cache, it means that all pods may have been accidentally deleted. At this time, the `storageWrapper.List` function will return `storage.ErrStorageNotFound`. Maybe the unit test here needs to add this use case.
The specific logic code is here: https://github.com/openyurtio/openyurt/blob/146a74498f51781f1ec1f9a98055cf3fa4c02f87/pkg/yurthub/cachemanager/storage_wrapper.go#L161

pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		obj, err := sWrapper.Get("kubelet/pods/default/mypod1")
+		if err != nil {
+			t.Errorf("failed to create obj, %v", err)
+		}
+		accessor := meta.NewAccessor()
+		name, _ := accessor.Name(obj)
+		if name != "mypod1" {
+			t.Errorf("the name is not expected, expect mypod1, get %s", name)
+		}
```
qclc, line 73: One point different from storage is that storageWrapper will store eligible objects in the cache, so it may be need to test whether the eligible objects are successfully saved in the cache. At the same time, it need to test the objects that do not meet the conditions are not saved in the cache.
Determine whether the object needs to be cached into the cache logic here:
https://github.com/openyurtio/openyurt/blob/146a74498f51781f1ec1f9a98055cf3fa4c02f87/pkg/yurthub/cachemanager/storage_wrapper.go#L253

pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,154 @@
...
+		accessor := meta.NewAccessor()
+		labels, _ := accessor.Labels(obj)
+		if vaule, ok := labels["tag"]; ok {
+			if vaule != "test" {
+				t.Errorf("failed to get label, expect test, get %s", vaule)
+			}
+		} else {
+			t.Errorf("unexpected error occurs, the label `tag` is not existed")
+		}
+
```
qclc, line 108: Here may need to verify the object in memory, the reason is the same as above.

pkg/yurthub/cachemanager/storage_wrapper_test.go
```diff
@@ -0,0 +1,154 @@
...
+	t.Run("Test delete storage", func(t *testing.T) {
+		err = sWrapper.Delete("kubelet/pods/default/mypod1")
+		if err != nil {
+			t.Errorf("failed to delete obj, %v", err)
+		}
+		_, err = sWrapper.Get("kubelet/pods/default/mypod1")
+		if err != storage.ErrStorageNotFound {
+			t.Errorf("unexpected error occurs, %v", err)
+		}
```
qclc, line 152: Here may need to verify the existence of the object in the cache for the same reason as above.

pkg/yurthub/cachemanager/storage_wrapper_test.go
MIBc: I get your point. But the cache is a private  object. StorageWrapper is an abstract class. The test case just tests the general behavior. It is unnecessary to know whether it cache data in memory or not.

pkg/yurthub/cachemanager/storage_wrapper_test.go
MIBc: done

MIBc: Good idea. I have moved the case to last. Please review.

qclc: Should we keep the use case of normal List requests for non-empty resources?

MIBc: Done.

qclc:
@rambohe-ch Hi, this unti test looks good to me, please help review whether there are any problems with this unit test, if not, maybe we can merge this code.

rambohe-ch:
@MIBc Thank you for improving unit test cases for OpenYurt.

PR #373, opened 2021-07-01
proposal: add ingress feature support to nodepool

zzguang:
Signed-off-by: zhenggu1 <zhengguang.zhang@intel.com>

Fixes #

```release-note

```

### DIFF
--- a/docs/proposals/20210628-nodepool-ingress-support.md
+++ b/docs/proposals/20210628-nodepool-ingress-support.md
@@ -1,0 +1,230 @@
+---
+title: NodePool Ingress Support
+authors:
+  - "@zzguang"
+reviewers:
+  - "@gnunu"
+  - "@Walnux"
+  - "@rambohe-ch"
+  - "@wenjun93"
+creation-date: 2021-06-28
+last-updated: 2021-06-28
+status: provisional
+---
+
+# Add Ingress Feature Support to NodePool
+
+## Table of Contents
+
+- [Add Ingress Feature Support to NodePool](#add-ingress-feature-support-to-nodepool)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@zzguang Very appreciate for your proposal. and I am also agree with solution4.

wenjun93:
@zzguang  the high level nodepool workload api cannot support customized configmaps in current stage, we might consider using one shared nginx config instead of customized configs among nodepools.

docs/proposals/20210628-nodepool-ingress-support.md
```diff
@@ -0,0 +1,230 @@
...
+	- It should be granted the authorities to create/update/patch the related resources in kube-system namespace.
+	- It connects with Yurthub to monitor the resources(ingress/service/configmap/secret/endpoint) of its own NodePool.
+
+3). Ingress controller sidecar:
+	- When sidecar is launched, it gets the NodePool id it belongs to, and connects to Yurthub to list/watch the NodePool.
+	- When it detects the NodePool annotation below is changed, it will update externalIPs of the corresponding service:
+	  ```yaml
+            nodepool.openyurt.io/ingress-ips: "xxx"
+	  ```
+	- When it detects the NodePool annotation below is changed, it will update the corresponding configmap of the NodePool,
```
rambohe-ch, line 218: If the workloads in NodePool use different configmaps, NodePool daemonset resource need to support different configmap setting. so how about create different configmaps directly by user instead of setting NodePool ingress-config?

zzguang:
> @zzguang the high level nodepool workload api cannot support customized configmaps in current stage, we might consider using one shared nginx config instead of customized configs among nodepools.

Okay, we can consider to use a global nginx config for all the nodepools, then we need to figure out 2 problems:
1. Who is responsible to create the configmap
2. Whether it is popular that different nodepool needs some personalized configuration
Thanks!

docs/proposals/20210628-nodepool-ingress-support.md
zzguang: About the details design for implementation, it's just a draft version by now.  We will investigate more to make it clear.
Thanks for your suggestion, we will take it into account.

rambohe-ch:
/approve

and @zzguang will share this proposal in OpenYurt community meeting on July 21.

PR #375, opened 2021-07-02
improve tunnel availability

aholic:
#### What type of PR is this?
/kind bug

1. before this pr, all csr(pending/approved/other status) will try enqueue every 10 seconds, but the default queue is only 10qps, 100 bucket. tunnel-server is not able to approve csr when large number of csr comes.
2. the connection is not available before certificate signed, so tunnel-agent must wait for the certificate signed

None

/assign @rambohe-ch 

```release-note
None
```

### DIFF
--- a/cmd/yurt-tunnel-agent/app/start.go
+++ b/cmd/yurt-tunnel-agent/app/start.go
@@ -18,6 +18,7 @@ package app
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/openyurtio/openyurt/cmd/yurt-tunnel-agent/app/config"
 	"github.com/openyurtio/openyurt/cmd/yurt-tunnel-agent/app/options"
@@ -30,6 +31,8 @@ import (
 	"github.com/openyurtio/openyurt/pkg/yurttunnel/util"
 
 	"github.com/spf13/cobra"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/certificate"
 	"k8s.io/klog/v2"
 )
@@ -90,6 +93,17 @@ func Run(cfg *config.CompletedConfig, stopCh <-chan struct{}) error {
 		return err
 	}
 	agentCertMgr.Start()
+
+	// 2.1. waiting for the certificate is generated
+	_ = wait.PollUntil(5*time.Second, func() (bool, error) {
+		if agentCertMgr.Current() != nil {
+			return true, nil
+		}
+		klog.Infof("certificate %s not signed, waiting...",
+			projectinfo.GetAgentName())
+		return false, nil
+	}, stopCh)
+	klog.Infof("certificate %s ok", projectinfo.GetAgentName())
 
 	// 3. generate a TLS configuration for securing the connection to server
 	tlsCfg, err := pki.GenTLSConfigUseCertMgrAndCA(agentCertMgr,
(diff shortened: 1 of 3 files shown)


cmd/yurt-tunnel-server/app/options/options.go
```diff
@@ -150,7 +150,7 @@ func (o *ServerOptions) Config() (*config.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	cfg.SharedInformerFactory = informers.NewSharedInformerFactory(cfg.Client, 10*time.Second)
+	cfg.SharedInformerFactory = informers.NewSharedInformerFactory(cfg.Client, 24*time.Hour)
```
Fei-Guo, line 153: Why do you change the resync time?

cmd/yurt-tunnel-agent/app/start.go
```diff
@@ -91,6 +94,17 @@ func Run(cfg *config.CompletedConfig, stopCh <-chan struct{}) error {
 	}
 	agentCertMgr.Start()
 
+	// 2.1. waiting for the certificate is generated
+	_ = wait.PollUntil(15*time.Second, func() (bool, error) {
```
Fei-Guo, line 98: Is 15 second poll interval too long?

cmd/yurt-tunnel-server/app/options/options.go
aholic: In my situation, i see thousands tunnel certificate-sign-requests are in pending/approved status. All of them will be enqueued every 10 seconds, which full fills the work-queue. The pending certificate-sign-requests can't be approved. So i did two things:
1. make the re-sync period longer. I think there's no need for the re-sync period to be so short, it's just a method to fix some unexpected situation.  
2. filter before enqueue, only do enqueue for pending tunnel certificate-sign-requests

cmd/yurt-tunnel-agent/app/start.go
aholic: I did it to reduce qps pressure of api-server. I don't want to see an avalanche when thousands of node comes(or other situations). 5 second or 15 second will not hurt user-experience, but an avalanche will.

cmd/yurt-tunnel-server/app/options/options.go
Fei-Guo: Understood. Alternatively, we may set it to 0 to disable resync .

cmd/yurt-tunnel-agent/app/start.go
Fei-Guo: Everybody waits 15 seconds does not eliminated the bursty quests but only delay them. You will need to add random delay to smooth the QPS or just give a reasonable delay for getting a signed certificate.

cmd/yurt-tunnel-server/app/options/options.go
aholic: yes, 0 is also ok. let's just make it a relative long period, in case of some unexpected situation.

cmd/yurt-tunnel-agent/app/start.go
aholic: let's set it to 5 second first
i will update it to a random delay sometime later, when I'm free..

cmd/yurt-tunnel-server/app/options/options.go
rambohe-ch: @Fei-Guo Set a relative long period(like 24 * time.Hour) for resync is more reasonable than disabling resync, in case of some unexpected situation.

rambohe-ch:
@Fei-Guo If you have not any other comments, i'd like to approve this pr.

PR #389, opened 2021-07-22
Proposal: OpenYurt Convertor Operator for converting K8S to OpenYurt

gnunu:
Signed-off-by: nunu <shaoqiang.chen@intel.com>

Fixes #

```release-note

```

### DIFF
--- a/20210722-yurtcluster-operator.md
+++ b/20210722-yurtcluster-operator.md
@@ -1,0 +1,201 @@
+---
+title: YurtCluster Operator
+authors:
+  - "@gnunu"
+  - "@lindayu17"
+reviewers:
+  - "@rambohe-ch"
+  - "@guofei"
+  - "@kadisi"
+  - "@wenjun93"
+creation-date: 2021-07-22
+last-updated: 2021-07-22
+status: provisional
+---
+
+# Convert Vanilla Kubernetes cluster to OpenYurt cluster through Operator
+
+## Table of Contents
+* [Convert Vanilla Kubernetes cluster to OpenYurt cluster through Opertaor](#convert-vanilla-kubernetes-cluster-to-openyurt-cluster-through-operator)
+  * [Table of Contents](#table-of-contents)
+  * [Glossary](#glossary)
+  * [Summary](#summary)
+  * [Motivation](#motivation)
+    * [Goals](#goals)
+    * [Non-Goals/Future Work](#non-goalsfuture-work)
+  * [Proposal](#proposal)
+    * [User Stories](#user-stories)
+      * [Story 1](#story-1)
+      * [Story 2](#story-2)
+    * [Requirements](#requirements)
+      * [Functional Requirements](#functional-requirements)
+        * [FR1](#fr1)
+        * [FR2](#fr2)
+        * [FR3](#fr3)
+        * [FR4](#fr4)
+      * [Non-Functional Requirements](#non-functional-requirements)
+        * [NFR1](#nfr1)
+        * [NFR2](#nfr2)
+      * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+      * [Risks and Mitigations](#risks-and-mitigations)
+  * [Alternatives](#alternatives)
+  * [Upgrade Strategy](#upgrade-strategy)
+  * [Additional Details](#additional-details)
+    * [Test Plan](#test-plan)
+  * [Implementation History](#implementation-history)
+
+## Glossary
+
+Refer to the [Cluster API Book Glossary](https://cluster-api.sigs.k8s.io/reference/glossary.html).
+
+## Summary
+
+This YurtCluster Operator is to translate a vanilla Kubernetes cluster into an OpenYurt cluster, through a simple API (YurtCluster CRD).
+
+## Motivation
+
+To make OpenYurt translation easy for user, i.e., as automative as possible.
+
+### Goals
+* Convert nodes automatically
+* Revert nodes automatically
+* Translated OpenYurt cluster should be the same as through yurtctl or doing manually
+
+### Non-Goals/Future Work
+
+This is NOT a replacement of yurtctl command.
+
+## Proposal
+
+This is an operator, with CRD and Controller.
+
+CRD's definition:
+
+```go
+type ComponentConfig struct {
+        Enabled bool   `json:"enebled,omitempty"`
+        Version string `json:"version,omitempty"`
+}
+
+type NodeSpec struct {
+        Names           []string `json:"names,omitempty"`
+        Pattern         string   `json:"pattern,omitempty"`
+        ExcludedNames   []string `json:"excludedNames,omitempty"`
+        ExcludedPattern string   `json:"excludedPattern,omitempty"`
+}
+
+type FailedNode struct {
+        Name string `json:"name,omitempty"`
+        Info string `json:"info,omitempty"`
+}
+
+// YurtClusterSpec defines the desired state of YurtCluster
+type YurtClusterSpec struct {
+        CloudNodes            NodeSpec        `json:"cloudNodes,omitempty"`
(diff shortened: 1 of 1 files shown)


docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,175 @@
...
+This is NOT a replacement of yurtctl command.
+
+## Proposal
+
+This is an operator, with CRD and Controller.
+
+CRD's definition:
+
+// YurtconvertorSpec defines the desired state of Yurtconvertor
+type YurtconvertorSpec struct {
```
Fei-Guo, line 74: The "YurtConvertor" does not sound like a CRD name for declaring a "state". How about "YurtCluster"?

The CRD SPEC should include CloudNodes[] and EdgeNodes[].  `Convert` and `Revert` are actions, not states. The controller should figure out the nodes that need to be reverted, hence we don't need the Revert array in the Spec.

To describe the Cloud/Edge Node lists, we can try the following struct

```
type YurtClusterSpec struct{
       CloudNodes ItemList
       EdgeNodes ItemList

       TunnelService    ComponentConfig    
       AppManager  ComponentConfig
}

// You may consider different config for Tunnel and AppManager separately to allow customization for each component
type ComponentConfig{
    Enable bool 
    Version string
}


type ItemList {
     //options
     Names []string
     //optional
     Pattern string // For regular expression
     //optional
     ExcludedNames []string
     //optional
     ExcludedPattern string 
} 


```

docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,175 @@
...
+        YurtAppManagerVersion    string   `json:"yurt-app-manager-version,omitempty"`
+}
+
+// YurtconvertorStatus defines the observed state of Yurtconvertor
+type YurtconvertorStatus struct {
+        Converted []string `json:"converted,omitempty"`
+        Reverted  []string `json:"reverted,omitempty"`
+        Excepted  []string `json:"excepted,omitempty"`
+        Failed    []string `json:"failed,omitempty"`
+}
```
Fei-Guo, line 91: Please update the Status based on the SPEC modifications suggested above accordingly. 
For failed nodes, the names are not enough. We need to expose the error msgs in the Status as well.

Fei-Guo:
Also，please think about the following questions:
- Should YurtCluster CR be a singleton in each Cluster?
- What are the controller behaviors when the YurtCluster CR is deleted?

gnunu:
Hi Fei,
Very heuristic suggestions. Thanks!
I am going to refine the proposal on data struct and implementaion details such as CR addition (if singleton) and deleteion response.

liangyuanpeng:
If the cluster nodes have multiple different deployment methods, what will happen at this time? E.g. kubeadm binary.

And does it mean that there will be two sets of code for convert/revert?

gnunu:
> If the cluster nodes have multiple different deployment methods, what will happen at this time? E.g. kubeadm binary.
> 
> And does it mean that there will be two sets of code for convert/revert?

I think the converted effect is added necessary changes to the node to be an OpenYurt node. If nodes are added through different methods, *maybe* there would be specific code for each case. This should be transparent to user. The Operator needs do all dirty works necessary. :)

gnunu:
Hi @Fei-Guo,
I refined the proposal based on your suggestions. Two points worth mentioning:
1) Now that CR would be a singleton, the ConfigMap mentioned in the original proposal for merging CRs' requests is not necessary any more.
2) On CR deletion, the whole cluster would be recovered back to K8S.

Please see if anything is missing.

Thanks!

gnunu:
the refinement link is here:
https://github.com/openyurtio/openyurt/pull/389/commits/54f227d22fc2f6a95fb3fc8e24acb78e38871126

gnunu:
sorry, one more refinement: use camelCase for json tags:
https://github.com/openyurtio/openyurt/pull/389/commits/f84094e75f77d38b0fc033a74d5d0c38aea617a1

gnunu:
> Please fix markdownlint error. Overall, LG.
> 
> I don't worry the performance too much since the node components installation can be done in parallel. You probably need to maintain an internal state machine to track the covert progress for each node to deal with frequent switches between edge mode and cloud mode for any reason.
> 
> @rambohe-ch Please take a look as well.

Thanks Fei!
markdownlint errors fixed.
Really nice mention about cloud-edge transition. It's necessary to design for such corner cases for overall right function and performance.

- Shaoqiang

docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,197 @@
...
+This YurtCluster Operator is to translate a vanilla Kubernetes cluster into an OpenYurt cluster, through a simple API (YurtCluster CRD).
+
+## Motivation
+
+To make OpenYurt translation easy for user, i.e., as automative as possible.
+
+### Goals
+* Convert nodes automatically
+* Revert nodes automatically
+* Translated OpenYurt cluster should be the same as through Yurtcutl or doing manually
```
rambohe-ch, line 62: Yurtcutl --> yurtctl

docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,197 @@
...
+
+This is NOT a replacement of yurtctl command.
+
+## Proposal
+
+This is an operator, with CRD and Controller.
+
+CRD's definition:
+
+type ComponentConfig struct {
```
rambohe-ch, line 74: please define all of structs in code block(please use ```)

docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// YurtClusterStatus defines the observed state of YurtCluster
+type YurtClusterStatus struct {
+        ConvertedCloudNodes []string        `json:"convertedCloudNodes,omitempty"`
+        ConvertedEdgeNodes  []string        `json:"convertedEdgeNodes,omitempty"`
+        YurtTunnelVersion   ComponentConfig `json:"yurtTunnelVersion,omitempty"`
+        AppManagerVersion   ComponentConfig `json:"appMangerVersion,omitempty"`
+        ExcludedCloudNodes  []string        `json:"excludedCloudNodes,omitempty"`
+        ExcludedEdgeNodes   []string        `json:"excludedEdgeNodes,omitempty"`
```
rambohe-ch, line 106: How about merge `ExcludedCloudNodes` and `ExcludedEdgeNodes` as `ExcludedNodes` filed. because both fields maybe contain same excluded nodes.

docs/proposals/20210722-openyurt-convertor-operator.md
```diff
@@ -0,0 +1,197 @@
...
+        ExcludedPattern string   `json:"excludedPattern,omitempty"`
+}
+
+type FailedNode struct {
+        Name string `json:"name,omitempty"`
+        Info string `json:"info,omitempty"`
+}
+
+// YurtClusterSpec defines the desired state of YurtCluster
+type YurtClusterSpec struct {
```
rambohe-ch, line 92: User maybe want to define the version of yurthub and yurt-controller-manager component, how about add fields to  define yurthub and yurt-controller-manager?

rambohe-ch:
@gnunu  please merge all of six commits into one commit.

gnunu:
> @gnunu please merge all of six commits into one commit.

Hi @rambohe-ch, thanks, have made revision on your suggestions.

will merge all the commits nto one.

gnunu:
@Fei-Guo @rambohe-ch,

I created a new pull request, for my bad not created a branch in the first place, sorry!
https://github.com/openyurtio/openyurt/pull/395

gnunu:
just pushed the merged proposal.

thanks @rambohe-ch for guide to do this!

PR #399, opened 2021-07-29
Yurtctl modify  kube-controllersetting to close the nodelifecycle-controller

### LINKED ISSUE #391: [BUG]How `yurtctl convert` can better close the `nodelifecycle controller` (`kube-controller-manager`)?
**What happened**:
When excuting the command `yurtctl convert`,  the clusterrolebinding `system:controller:node-controller` will be deleted to turn off the `nodelifecycle controller` (in `kube-controller-manager`).
However, the flowing questions will appear afterwards:
- Other controllers using clusterrolebinding `system:controller:node-controller` can't work normally.
  eg: issue: https://github.com/openyurtio/openyurt/issues/380
- In a high availability scenario, when kube-controller-manager changes the leader, kube-controller-manager can't start successfully because of lack of clusterrolebinding `system:controller:node-controller` .
  Issue: https://github.com/openyurtio/openyurt/issues/347

`yurtctl convert`  doesn't  turn off `nodelifecycle controller` by deleting clusterrolebinding `system:controller:node-controller`. And it can support the high availability of `kube-controller-manager`.

*Solutions currently proposed:*
When executing `yurtctl convert`, 1) If `kube-controller-manager` is deployed through static file `/etc/kubernetes/manifests/kube-controller-manager.yaml`: Create jobs for all candidates to modify this file to turn off `nodelifecycle controller` . 2) If `kube-controller-manager` is not deployed through a static file: Modify `kube-controller-manager` template directly.
*And this scheme has the following problems:*
If the kube-controller-manager is not deployed through a static file (/etc/kubernetes/manifests/), modify `kube-controller-manager` template directly will lead to the runtime template is inconsistent with the user's local yaml file.

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**
/kind bug

Peeknut:
/kind bug

If kube-controlller-manager is deployed as a static pod, `yurtctl convert` turn off `nodelifecycle controller` by modifying `kube-controller-manager.yaml` instead of deleting clusterrolebinding `system:controller:node-controller`. So OpenYurt cluster created by yurtctl can run normally and support the high availability of `kube-controller-manager`.

Fixes #391

/assign @rambohe-ch

```release-note

```

### DIFF
--- a/pkg/yurtctl/cmd/revert/revert.go
+++ b/pkg/yurtctl/cmd/revert/revert.go
@@ -23,17 +23,17 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	v1 "k8s.io/api/core/v1"
-	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog"
+	nodeutil "k8s.io/kubernetes/pkg/controller/util/node"
 
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
 	"github.com/openyurtio/openyurt/pkg/yurtctl/constants"
 	"github.com/openyurtio/openyurt/pkg/yurtctl/lock"
 	kubeutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/kubernetes"
-	nodeutil "k8s.io/kubernetes/pkg/controller/util/node"
+	strutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/strings"
 )
 
 // RevertOptions has the information required by the revert operation
@@ -125,7 +125,13 @@ func (ro *RevertOptions) RunRevert() (err error) {
 	}
 	klog.V(4).Info("the server version is valid")
 
-	// 1.1. check the state of worker nodes
+	// 1.1. get kube-controller-manager HA nodes
+	kcmNodeNames, err := kubeutil.GetKubeControllerManagerHANodes(ro.clientSet)
+	if err != nil {
+		return
+	}
+
+	// 1.2. check the state of worker nodes
 	nodeLst, err := ro.clientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 	if err != nil {
 		return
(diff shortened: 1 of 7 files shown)


pkg/yurtctl/constants/constants.go
```diff
@@ -251,5 +251,57 @@ spec:
...
+      nodeName: {{.nodeName}}
+      containers:
+      - name: yurtctl-enable-node-controller
+        image: {{.yurtctl_servant_image}}
+        imagePullPolicy: IfNotPresent
+        command:
+        - /bin/sh
+        - -c
+        args:
+        - "nsenter -t 1 -m -u -n -i -- sed -i 's/--controllers=-nodelifecycle,/--controllers=/g' {{.pod_manifest_path}}/kube-controller-manager.yaml"
```
rambohe-ch, line 303: if no `--controllers` parameter is set in `kube-controller-manager.yaml`, the above command can not disable `nodelifecycle` controller.

pkg/yurtctl/cmd/revert/revert.go
```diff
@@ -213,30 +219,18 @@ func (ro *RevertOptions) RunRevert() (err error) {
...
-			{
-				Kind:      "ServiceAccount",
-				Name:      "node-controller",
-				Namespace: "kube-system",
-			},
+	// 6. enable node-controller
+	if err = kubeutil.RunServantJobs(ro.clientSet,
+		map[string]string{
+			"action":                "enable",
+			"yurtctl_servant_image": ro.YurtctlServantImage,
```
rambohe-ch, line 226: why use `ro.YurtctlServantImage` to enable or disable nodelifecycle controller?

pkg/yurtctl/constants/constants.go
Peeknut: Oh, yes. In the cluster installed by kubeadm, the static file of kube-controller-manager contains this parameter `--controllers` by default. In order to deal with other situations, the situation without this parameter should indeed be considered.
And the method of modifying files will be optimized in the future to make string matching more accurate instead of using `sed`.

pkg/yurtctl/cmd/revert/revert.go
Peeknut: Because the system architecture may be amd64, arm64, arm, using a unified basic image may cause problems. So the parameter of yurtctl convert --yurtctl-servant-image is reused here.
The yurtctl-servant image may be separated from yurtctl in the future for more elegant use.

PR #407, opened 2021-08-06
fix pod access yurthub https endpoint error

rambohe-ch:
/kind bug

1. If pods access dummy ip address of yurthub https server, pods get the error info: x509: certificate is valid for 127.0.0.1, not 169.254.2.1, so we need to add ip address of dummy network interface into server certificate of yurthub.

2. Move server certificate dir and files under `pkg/yurthub/certificate` dir,  so all certificates(include both server and client certificate) can be managed under same dir. 

Fixes #

```release-note

```


huangyuqi:
LGTM

PR #437, opened 2021-08-29
enhancement: proxy request with bearer token in header without client certificate

rambohe-ch:
/kind enhancement

- background:
when pods on edge nodes access kube-apiserver through yurthub, yurthub will use client certificate to proxy requests to kube-apiserver. so in order to get authorization by kube-apiserver, we need to add rbac setting of pods to client certificate of yurthub(system:nodes group).

- enhancement:
In order to prevent add additional rbac setting to `system:nodes` group, we use the original bearer token to proxy request to kube-apiserver instead of using client certificate to proxy requests, so kube-apiserver can get authorization for pods based on the bearer token of pods instead of client certificate of yurthub.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/proxy/remote/remote.go
+++ b/pkg/yurthub/proxy/remote/remote.go
@@ -23,6 +23,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"strings"
 
 	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
 	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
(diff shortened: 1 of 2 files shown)


pkg/yurthub/proxy/remote/remote.go
```diff
@@ -80,6 +89,15 @@ func (rp *RemoteProxy) Name() string {
 }
 
 func (rp *RemoteProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	rp.reverseProxy.Transport = rp.currentTransport
+	auth := strings.TrimSpace(req.Header.Get("Authorization"))
+	if auth != "" {
```
Fei-Guo, line 94: Can you add a comment here to explain when the auth is nil and when it is not nil?

rambohe-ch: comments were added.

PR #440, opened 2021-08-31
feature: add discard cloud service filter in yurthub

rambohe-ch:
/kind feature

- background:
    Kube-proxy(ipvs mode) will configure ipvs rules for `LoadBalancer` service on the edge node, so if pods(like yurt-tunnel-agent) on edge nodes use ingress ip of `LoadBalancer` service to access the cloud pods(like yurt-tunnel-server),  the connection will be refused for ipvs rule will dnat the ingress ip to pod ip on the edge node.
    And  end user have come across the same error that yurt-tunnel-agent failed to use ingress ip of `x-tunnel-server-svc` service to access yurt-tunnel-server. the detailed info is here: https://github.com/openyurtio/openyurt/issues/447

- solution:
    In order to make sure pods on edge nodes can use LoadBalancer service to access pods on cloud nodes, we need disable the kube-proxy dnat rule for LoadBalancer service. so we add a filter named `discardCloudService` for yurthub to discard LoadBalancer service for kube-proxy component.
At the same time, some ClusterIP services(like kube-system/x-tunnel-server-internal-svc) are not need to aware by edge nodes, so we also discard these ClusterIP service in the new filter.

Fixes #

```release-note

```

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -28,6 +28,7 @@ import (
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
 	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
 	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/discardcloudservice"
 	"github.com/openyurtio/openyurt/pkg/yurthub/filter/masterservice"
 	"github.com/openyurtio/openyurt/pkg/yurthub/filter/servicetopology"
 	"github.com/openyurtio/openyurt/pkg/yurthub/kubernetes/meta"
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@Fei-Guo I have updated the context of this feature. and `internalTrafficPolicy` feature of `Service` in k8s v1.21 can not solve the above problem, so after v1.21 version, we need to keep `discardcloudservice` filter enabled.

DrmagicE:
@rambohe-ch Hi, I am working on "adding yurthub on cloud-side feature" recently. In order to enable yurt-tunnel DNS mode, the kube-proxy on the cloud-side needs the "kube-system/x-tunnel-server-internal-svc" endpoints, so maybe we shoud add an option to let the cloud-side components can still get the endpoint?

rambohe-ch:
@DrmagicE On cloud nodes, you can add `--disabled-resource-filters=discardcloudservice` parameter for yurthub to disable `discardcloudservice` filter.

PR #442, opened 2021-08-31
feature: support proxy https request from cloud to edge through yurt-tunnel

### LINKED ISSUE #414: [yurt tunnel]support proxy https request from cloud to edge through yurt-tunnel
Yurt-tunnel components can proxy https request from cloud to edge nodes, so user from the cloud can access the https service at the edge through yurt-tunnel.

**Why is this needed**:
Some cloud-native ecosystem components(like Node Exporter) serve https requests at the edge, in order to support prometheus collect metrics of node-exporter, we need to support proxy https requests through yurt-tunnel.

**others**
/kind feature

rambohe-ch:
/kind feature

1. Yurt-tunnel can proxy https request from cloud to edge nodes, so components(like prometheus) on the cloud nodes can access the https service(like node-exporter) on the edge node through yurt-tunnel.
2. Beside of `10250` and `10255` Ports, In order to support more tcp ports of https and http service on edge nodes. use can configure the `yurt-tunnel-server-cfg` configmap, and add specified tcp ports in `https-proxy-ports` and `http-proxy-port` field of configmap.
3. for example, prometheus on cloud node likes to collect metrics at endpoint `https://{nodeIP}:9051` and `http://{ndoeIP}:9052` on edge nodes,  you need to configure as following:

```
http-proxy-ports: "9052"
https-proxy-ports: "9051"
```

Fixes #414 

```release-note

```

### DIFF
--- a/config/yaml-template/yurt-tunnel-server.yaml
+++ b/config/yaml-template/yurt-tunnel-server.yaml
@@ -130,6 +130,8 @@ metadata:
   namespace: kube-system
 data:
   localhost-proxy-ports: "10266, 10267"
+  http-proxy-ports: ""
+  https-proxy-ports: ""
   dnat-ports-pair: ""
 ---
 apiVersion: apps/v1
--- a/pkg/yurttunnel/util/util.go
+++ b/pkg/yurttunnel/util/util.go
@@ -1,20 +1,39 @@
+/*
+Copyright 2021 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package util
 
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
+	"strconv"
 	"strings"
 
-	"github.com/gorilla/mux"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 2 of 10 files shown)


pkg/yurttunnel/util/util.go
```diff
@@ -52,41 +60,113 @@ func RunMetaServer(addr string) {
...
+		return proxyPorts, portMappings, err
+	}
+
+	_, securePort, err := net.SplitHostPort(secureListenAddr)
+	if err != nil {
+		return proxyPorts, portMappings, err
+	}
+
+	// field dnat-ports-pair will be deprecated in future version
+	pairStr, ok := cm.Data[yurttunnelServerDnatDataKey]
```
Fei-Guo, line 101: It seems that you can dedup the code 102-115 with 119-134 by introducing at local function.

config/yaml-template/yurt-tunnel-server.yaml
```diff
@@ -129,6 +129,8 @@ metadata:
   name: __project_prefix__-tunnel-server-cfg
   namespace: kube-system
 data:
+  http-proxy-ports: ""
+  https-proxy-ports: ""
```
DrmagicE, line 133: What about using an array rather than a comma-separated string?

rambohe-ch: I think comma-separated string is more simpler than array string for user to edit fields in configmap resource. because array string like the following setting:
```
http-proxy-ports: '["aaa", "aaa", "aaa"]'
https-proxy-ports: '["bbb", "bbb", "bbb"]'
```
so how about keep this comma-separated string design?

DrmagicE: Ok, it is reasonable.

pkg/yurttunnel/util/util.go
rambohe-ch: @Fei-Guo It's a good idea, fixed it.

Fei-Guo:
Please add the copyright headers for all the files that miss it in this change. Otherwise, LGTM.

rambohe-ch:
> Please add the copyright headers for all the files that miss it in this change. Otherwise, LGTM.

ok, copyright headers are added.

SataQiu (commented):
/lgtm

PR #446, opened 2021-09-02
remove excess chan that will block the program

zc2638:
/kind bug

Fixes #

```release-note

```


rambohe-ch:
@zc2638 Very appreciate you for fixing the chan block problem.
Would you like to apply for OpenYurt community member?  and you reference the apply example here: https://github.com/openyurtio/community/issues/19

PR #455, opened 2021-09-07
enhancement: add name info for filters

rambohe-ch:
/kind enhancement

Add name info for filters in yurthub, meanwhile, add filter name in the log message in order to monitor status of the filter.

Fixes #

```release-note

```


pkg/yurthub/filter/interfaces.go
```diff
@@ -26,6 +26,7 @@ import (
 )
 
 type Interface interface {
+	Name() string
```
Fei-Guo, line 29: Is the change in the interface necessary?

rambohe-ch: @Fei-Guo ok, i have removed the new func `Name()`.

PR #457, opened 2021-09-08
bugfix: prepare yurthub server tls config panic

rambohe-ch:
/kind bug

- bug:
   if yurthub re-startup when network between cloud and edge disconnected, a panic will happen for preparing server tls config. and the panic like as following:
  
image

- reason:
  yurthub will use healthy server address and certificate under `/var/lib/yurthub/pki` dir to create `rest.Config` for https server. the code is here: https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/kubernetes/rest/config.go#L83-113
  and when cloud-edge network breaks down, healthy server will become true, so `rest.Config` will be nil, then panic happened.

- solution:
  we only need to prepare for `rest.Config`, and not need to connect remote server, so healthy server is not a requirement, so we add a `needHealthyServer` parameter for  `GetRestConfig()` to create `rest.Config` whether healthy server should be set or not. 

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/gc/gc.go
+++ b/pkg/yurthub/gc/gc.go
@@ -73,7 +73,7 @@ func (m *GCManager) Run() {
 	go wait.JitterUntil(func() {
 		klog.V(2).Infof("start gc events after waiting %v from previous gc", time.Since(m.lastTime))
 		m.lastTime = time.Now()
-		cfg := m.restConfigManager.GetRestConfig()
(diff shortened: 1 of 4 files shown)


zc2638:
/lgtm
I have passed the test in my environment.

rambohe-ch:
> /lgtm
> I have passed the test in my environment.

@zc2638 Thanks for testing.

PR #458, opened 2021-09-08
enhancement: skip ensure iptables rule error for yurthub

rambohe-ch:
/kind enhancement

Iptables rules are not requirement for clients(like kubelet, kube-proxy etc.) access yurthub, we can skip the iptables ensure error and continue to ensure remained iptables rules.

Fixes #

```release-note

```


Fei-Guo:
Can you explain the problem of current code which fails earlier?  The error seems indicating the host is not in expected state.

rambohe-ch:
> Can you explain the problem of current code which fails earlier? The error seems indicating the host is not in expected state.

@Fei-Guo Iptables rules in [makeupIptablesRules](https://github.com/openyurtio/openyurt/blob/446daa83ee3cba8a2e5173e810199f47685eecd4/pkg/yurthub/network/iptables.go#L53-L74), Each rule has its own role, and there is no correlation between the rules. so we can skip the error and continue the later iptables rule setting.

PR #465, opened 2021-09-10
bugfix: add rabc setting for yurthub

rambohe-ch:
/kind bug

yurthub need to list/watch nodepool resource, so we need add these `RBAC` rights to `system:nodes` group because a node certificate have been applied by yurthub.

Fixes #

```release-note

```

### DIFF
--- a/config/setup/yurthub.yaml
+++ b/config/setup/yurthub.yaml
@@ -7,10 +7,10 @@ metadata:
   namespace: kube-system
 spec:
   volumes:
-  - name: pki
+  - name: hub-dir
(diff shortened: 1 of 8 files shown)


Fei-Guo:
Should you also change `openyurt/hack/lib/build.sh` file as well to replace `__pki_path__' ?

rambohe-ch:
> Should you also change `openyurt/hack/lib/build.sh` file as well to replace `**pki_path**' ?

@Fei-Guo build.sh has been updated, please check it.

PR #466, opened 2021-09-10
enhancement: add configmap to configure user agents for specify response cache.

rambohe-ch:
/kind feature

- background:
    At present, only response for requests from fixed user agents(like kube-proxy, flanneld, coredns, kubelet, yurthub, yurt-tunnel-agent) can be cached in local disk by yurthub. if user's pod access kube-apiserver through yurthub, the response can not be cached on local disk for user agents can not be recognized.

- solution:
  1. It's not good idea to enable cache all response for requests through yurthub, because some clients only want to access kube-apiserver or some list requests may get large volume data and make pressure to local disk.
  2. So we add an configmap named yurt-hub-cfg with `cache_agents` field in kube-system namespace, user can add request user agent in this field.
  3. For example, `calico` components want to access kube-apiserver through yurthub and want to use yurthub cache ability, you can configure as following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: yurt-hub-cfg
  namespace: kube-system
data:
  cache_agents: "calico"
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/cachemanager/cache_agent.go
+++ b/pkg/yurthub/cachemanager/cache_agent.go
@@ -18,94 +18,65 @@ package cachemanager
 
 import (
 	"strings"
+	"time"
 
-	"github.com/openyurtio/openyurt/pkg/projectinfo"
-
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/sets"
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+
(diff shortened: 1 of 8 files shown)


pkg/yurthub/cachemanager/cache_agent.go
```diff
@@ -18,94 +18,62 @@ package cachemanager
...
-		for _, agent := range agents {
-			if _, ok := cm.cacheAgents[agent]; !ok {
-				hasUpdated = true
-				break
-			}
-		}
+	oldAgents := cm.cacheAgents.Delete(util.DefaultCacheAgents...)
+
+	if oldAgents.Equal(newAgents) {
+		return sets.String{}
```
Fei-Guo, line 64: Should you add the DefaultCacheAgents back before returning sets.String{} here and for constructing new cache agents below?

rambohe-ch: > Should you add the DefaultCacheAgents back before returning sets.String{} here and for constructing new cache agents below?

fixed, DefaultCacheAgents have been added.

PR #479, opened 2021-09-17
UPDATE config.go

guoguodan:
it may be null point

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -159,7 +159,7 @@ func Complete(options *options.YurtHubOptions) (*YurtHubConfiguration, error) {
 
 func parseRemoteServers(serverAddr string) ([]*url.URL, error) {
 	if serverAddr == "" {
-		return make([]*url.URL, 0, 0)
+		return make([]*url.URL, 0)
 	}
 	
 	servers := strings.Split(serverAddr, ",")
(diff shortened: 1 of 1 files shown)


cmd/yurthub/app/config/config.go
```diff
@@ -158,6 +158,10 @@ func Complete(options *options.YurtHubOptions) (*YurtHubConfiguration, error) {
 }
 
 func parseRemoteServers(serverAddr string) ([]*url.URL, error) {
+	if serverAddr == "" {
+		return make([]*url.URL, 0)
```
rambohe-ch, line 162: how about change to `return []*url.URL{}, nil`?

rambohe-ch: one return value(`error`) is missed.

guoguodan: Thinks！ I've added one return value(error).

rambohe-ch:
@guoguodan please merge these four commits into one commit.
image

guoguodan:
thinks！ this pr close. I will create a new pr.

rambohe-ch:
> thinks！ this pr close. I will create a new pr.

@guoguodan It's not need to create new pr, and you can git rebase these four commits into one commit, then git push --force  to overwrite the original branch.

PR #482, opened 2021-09-18
yurthub: stop renew node lease when kubelet's heartbeat is stopped

### LINKED ISSUE #420: [yurthub]improve the healthy check mechanism when kubelet's heartbeat stopped
**Healthy check background**
In order to improve the healthy check efficiency, yurthub makes `NodeLease` request to check, at the same time `NodeLease` request is used as heartbeat of edge node. so kubelet's `NodeLease` request is intercepted by yurthub and return local nodeLease back to kubelet.

If kubelet component down, yurthub should stop to send `NodeLease` request to kube-apiserver to check server's healthy status. 

**Why is this needed**:
The master was unable to realize that Kubelet had died
When kubelet component down, the Node status should become NotReady. If yurthub continue to send `NodeLease` request to kube-apiserver,  k8s master was unable to aware of that kubelet had down. 

**others**
/kind enhancement

SataQiu:
#### What type of PR is this?
 /kind bug
> Uncomment only one ` /kind <>` line, hit enter to put that in a new line, and remove leading whitespace from that line:
> /kind bug
> /kind documentation
> /kind enhancement
> /kind good-first-issue
> /kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

yurthub: stop renew node lease when kubelet's heartbeat is stopped

Fixes #420

/assign @rambohe-ch @Fei-Guo 

```release-note
yurthub: improve the healthy check mechanism when kubelet's heartbeat stopped
```

### DIFF
--- a/pkg/yurthub/healthchecker/health_checker.go
+++ b/pkg/yurthub/healthchecker/health_checker.go
@@ -41,6 +41,7 @@ const (
 // HealthChecker is an interface for checking healthy stats of server
 type HealthChecker interface {
 	IsHealthy(server *url.URL) bool
+	UpdateLastKubeletLeaseReqTime(time.Time)
 	Run()
 }
 
@@ -48,12 +49,15 @@ type setNodeLease func(*coordinationv1.Lease) error
 type getNodeLease func() *coordinationv1.Lease
 
 type healthCheckerManager struct {
-	remoteServers     []*url.URL
-	checkers          map[string]*checker
-	latestLease       *coordinationv1.Lease
-	sw                cachemanager.StorageWrapper
-	remoteServerIndex int
-	stopCh            <-chan struct{}
+	sync.RWMutex
+	remoteServers           []*url.URL
+	checkers                map[string]*checker
+	latestLease             *coordinationv1.Lease
+	sw                      cachemanager.StorageWrapper
+	remoteServerIndex       int
+	lastKubeletLeaseReqTime time.Time
+	healthCheckGracePeriod  time.Duration
+	stopCh                  <-chan struct{}
 }
 
 type checker struct {
@@ -76,11 +80,12 @@ func NewHealthChecker(cfg *config.YurtHubConfiguration, tp transport.Interface, 
 	}
 
 	hcm := &healthCheckerManager{
-		checkers:          make(map[string]*checker),
-		remoteServers:     cfg.RemoteServers,
-		remoteServerIndex: 0,
-		sw:                cfg.StorageWrapper,
-		stopCh:            stopCh,
+		checkers:               make(map[string]*checker),
+		remoteServers:          cfg.RemoteServers,
+		remoteServerIndex:      0,
+		sw:                     cfg.StorageWrapper,
+		healthCheckGracePeriod: cfg.KubeletHealthGracePeriod,
+		stopCh:                 stopCh,
 	}
 
 	for _, remoteServer := range cfg.RemoteServers {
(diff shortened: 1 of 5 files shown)


pkg/yurthub/proxy/proxy.go
```diff
@@ -98,7 +102,20 @@ func (p *yurtReverseProxy) buildHandlerChain(handler http.Handler) http.Handler
 }
 
 func (p *yurtReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	if !hubutil.IsKubeletLeaseReq(req) && p.loadBalancer.IsHealthy() {
+	now := time.Now()
+	isKubeletLeaseReq := hubutil.IsKubeletLeaseReq(req)
+	if isKubeletLeaseReq {
+		p.loadBalancer.Resume()
+		p.lastKubeletLeaseReqTime = now
+	}
```
Fei-Guo, line 110: Please use `else if` here.
Also please combine three conditions together in `else if` clause

Fei-Guo (commented):
This seems to be a heavy solution.  
Can we just log the `lastKubeletLeaseReqTime` in ServerHTTP and just check the timestamp when we send heartbeat to APIServer?

pkg/yurthub/proxy/proxy.go
```diff
@@ -98,7 +102,20 @@ func (p *yurtReverseProxy) buildHandlerChain(handler http.Handler) http.Handler
...
+
+	if !p.lastKubeletLeaseReqTime.IsZero() && p.healthCheckGracePeriod > 0 {
+		if now.After(p.lastKubeletLeaseReqTime.Add(p.healthCheckGracePeriod)) {
+			p.loadBalancer.Pause()
```
rambohe-ch, line 114: we can not use lease request or other http requests to trigger to pause healthyChecker, because no requests except watch requests can be sent to yurthub when kubelet stopped.

btw: maybe we only need to update `lastKubeletLeaseReqTime` here, and then check kubelet stopped or not in HealthyChecker `sync` interval.

rambohe-ch:
@SataQiu please fix the conflicting files.

SataQiu:
Thanks @rambohe-ch @Fei-Guo 
Updated!

pkg/yurthub/healthchecker/health_checker.go
```diff
@@ -181,50 +190,66 @@ func newChecker(
...
 	return c, nil
 }
 
 func (c *checker) check() bool {
-	baseLease := c.getLastNodeLease()
-	lease, err := c.nodeLease.Update(baseLease)
-	if err == nil {
-		if err := c.setLastNodeLease(lease); err != nil {
-			klog.Errorf("set last node lease fail: %v", err)
+	var err error
```
rambohe-ch, line 208: `check()` is used for sending lease to remote server and checking healthy status. and we need to stop `check()` when kubelet stopped renew node lease.  so i think add `kubelet stopped renew node lease verification` in `func (hcm *healthCheckerManager) sync()` maybe more readable. like as following:

```
func (hcm *healthCheckerManager) sync() {
        // check kubelet stopped renew node lease or not
	isKubeletStoped, now := false, time.Now()
	lastKubeletLeaseReqTime := c.getLastKubeletLeaseReqTime()
	if !lastKubeletLeaseReqTime.IsZero() && c.healthCheckGracePeriod > 0 {
		isKubeletStoped = now.After(lastKubeletLeaseReqTime.Add(c.healthCheckGracePeriod))
	}

	// Ensure that the node heartbeat can be reported when there is a healthy remote server.
	//try detect all remote server in a loop, if there is an remote server can update nodeLease, exit the loop.
	for i := 0; i < len(hcm.remoteServers); i++ {
		c := hcm.getChecker()
                if isKubeletStoped {
                       klog.Infof("kubelet does not post lease request for more than %v, stop renew node lease and assume remote server %s is unhealthy", now.Sub(lastKubeletLeaseReqTime), c.remoteServer.String())
                      // do other settings.
                      continue              
                }

		if c.check() {
			break
		}
	}
}
```

pkg/yurthub/proxy/remote/loadbalancer.go
```diff
@@ -175,6 +177,12 @@ func (lb *loadBalancer) IsHealthy() bool {
 	return false
 }
 
+func (lb *loadBalancer) UpdateLastKubeletLeaseReqTime(lastKubeletLeaseReqTime time.Time) {
+	for i := range lb.backends {
+		lb.backends[i].checker.UpdateLastKubeletLeaseReqTime(lastKubeletLeaseReqTime)
```
rambohe-ch, line 182: It's not need to set `lastKubeletLeaseReqTime` for every backend, and maybe we can use checker to call `UpdateLastKubeletLeaseReqTime` only once.

btw: we need add a new filed `checker` for `loadBalancer`

rambohe-ch: please skip the above comment. 
we don't need to add `UpdateLastKubeletLeaseReqTime()` func for `loadBalancer`. and we can add a new field `checker` for `yurtReverseProxy`, so we can call `UpdateLastKubeletLeaseReqTime` directly in `ServeHTTP`

SataQiu: Updated

rambohe-ch (approved):

PR #491, opened 2021-09-22
Update config.go

guoguodan:
it maybe null point

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -164,6 +164,9 @@ func Complete(options *options.YurtHubOptions) (*YurtHubConfiguration, error) {
 }
 
 func parseRemoteServers(serverAddr string) ([]*url.URL, error) {
+	if serverAddr == "" {
+		return make([]*url.URL, 0), fmt.Errorf("--server-addr should be set for hub agent")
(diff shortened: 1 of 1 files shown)


cmd/yurthub/app/config/config.go
```diff
@@ -164,6 +164,9 @@ func Complete(options *options.YurtHubOptions) (*YurtHubConfiguration, error) {
 }
 
 func parseRemoteServers(serverAddr string) ([]*url.URL, error) {
+	if serverAddr == "" {
+		return make([]*url.URL, 0), fmt.Errorf("--server-addr should be set for hub agent")
```
Fei-Guo, line 168: We have line 189 below which should protect the case that serverAddr is "".

rambohe-ch: I think that line 189 will check resolved serverAddress, and this pr only check the parameter at the func entry. it's make sense.

rambohe-ch: > We have line 189 below which should protect the case that serverAddr is "".

@Fei-Guo I think that line 189 will check resolved serverAddress, and this pr only check the parameter at the func entry. it's make sense.

Issue #384, opened 2021-07-15
[BUG] cannot create certificate signing request: Post https://xx/apis/certificates.k8s.io/v1beta1/certificatesigningrequests: x509: certificate signed by unknown authority


konghui:
**What happened**:
manual install yurthub  it display this error:
```
waiting for preparing client certificate
```
after solve this. it show
```
0625 11:10:21.580035       1 certificate_manager.go:434] Failed while requesting a signed certificate from the master: cannot create certificate signing request: Post https://xxx:6443/apis/certificates.k8s.io/v1beta1/certificatesigningrequests: x509: certificate signed by unknown authority 
```

**What you expected to happen**:
yurthub start correctly
**How to reproduce it (as minimally and precisely as possible)**:
kubectl apply -f config/setup/yurthub.yaml
**Anything else we need to know?**:

**Environment**:
- OpenYurt version:
- Kubernetes version (use `kubectl version`):
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

konghui:
Problem one is:  user should `--join-token=$bootstraptoken` on the yourthub.
Problem two is :If user chose manual install `openyurt` with self sign certificate. User should make sure `CA` certificate locate in  `/var/lib/yurthub/pki/ca.crt`.

DreamEagle0:
Hi @konghui 
Since yurthub is deployed as static pod, I think you should not deploy it by running `kubectl apply -f` or change it to daemonset. Instead, you should put the yaml file in edgenode static pod path (default: /etc/kubernetes/manifests).

konghui:
Ok thanks.

rambohe-ch:
> Ok thanks. But I think if yurthub deployed as a static pod. It still need enough `RBAC` permissions

We have configured `RBAC` setting in default `config/setup/yurthub.yaml`. if you have any problems with yurthub, please let me in any way.

DrmagicE:
Same issues here when I try to convert a pre-installed kubernetes cluster. It seams there were some problems for yurthub to get the certificate.
```
1 cert_mgr.go:398] avoid tcp conn leak, close old tcp conn that used to rotate certificate
I0908 13:52:08.426593       1 connrotation.go:110] forcibly close 1 connections on 192.168.33.220:6443 for hub certificate manager dialer
[...]

PR #510, opened 2021-09-29
bugfix(yurtctl): ensure the join token of yurthub has right usages and groups

DrmagicE:
/kind bug

When converting cluster with `yurtctl convert`, `yurtctl` will try to get the join-token for yurthub from Kubernetes secrets. 
We should ensure that the token can be used as a bootstrap token, otherwise the yurthub cannot startup.
For example, in my setup, `yurtctl` select a wrong token(the first one), so it can not get the certificate correctly.
```
[root@master-1 ~]# kubeadm token list
TOKEN                     TTL         EXPIRES                     USAGES                   DESCRIPTION                                                EXTRA GROUPS
8b9x1o.5ranibmuunw34b5v   55m         2021-09-08T23:04:10+08:00   <none>                   Proxy for managing TTL for the kubeadm-certs secret        <none>
argmko.fxklti113u7t2gvf   22h         2021-09-09T21:04:10+08:00   authentication,signing   The default bootstrap token generated by 'kubeadm init'.   system:bootstrappers:kubeadm:default-node-token
latkt1.a2d6m1zi46gvuflz   22h         2021-09-09T21:04:16+08:00   authentication,signing   <none>                                                     system:bootstrappers:kubeadm:default-node-token
```
Ref: https://github.com/openyurtio/openyurt/issues/384#issuecomment-915347725

Fixes #

/assign @rambohe-ch 

```release-note

```

### DIFF
--- a/pkg/yurtctl/util/kubernetes/util.go
+++ b/pkg/yurtctl/util/kubernetes/util.go
@@ -26,11 +26,12 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
-	v1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	"k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/yurtctl/util/kubernetes/util.go
```diff
@@ -608,12 +614,36 @@ func GetOrCreateJoinTokenString(cliSet *kubernetes.Clientset) (string, error) {
...
+			Usages: kubeadmcontants.DefaultTokenUsages,
+			Groups: kubeadmcontants.DefaultTokenGroups,
+		}}); err != nil {
 		return "", err
 	}
 	return tokenStr, nil
 }
 
+// isBootstrapToken checks if the token can be used as a bootstrap token.
+func isBootstrapToken(token *kubeadmapi.BootstrapToken) bool {
```
rambohe-ch, line 629: isBootstrapToken is used to check token has the right usages and groups, so how about define the func name to `usagesAndGroupsAreValid`?

DrmagicE: @rambohe-ch thanks for the comment, I agree that`usagesAndGroupsAreValid` is better, updated.

PR #563, opened 2021-11-09
When yurtctl revert executes yurtappmanager to delete automatically, no crd resource error is reported.

### LINKED ISSUE #557: When yurtctl revert executes yurtappmanager to delete automatically, no crd resource error is reported.
**What happened**:
If yurtappmanager is not deployed and yurtctl revert is executed, a notfound error will be reported when yurtappmanager is deleted.

A not found error occurred

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

yanyhui:
…D resource error

/kind bug

Yurtappmanager does not use the method of deleting crd resources when crd is not deployed.

Fixes #557 

```release-note

```

### DIFF
--- a/pkg/yurtctl/util/kubernetes/util.go
+++ b/pkg/yurtctl/util/kubernetes/util.go
@@ -31,12 +31,14 @@ import (
 	"sync"
 	"time"
 
+	"github.com/spf13/pflag"
 	"k8s.io/api/admissionregistration/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


pkg/yurtctl/util/kubernetes/util.go
```diff
@@ -31,12 +31,14 @@ import (
 	"sync"
 	"time"
 
+	"github.com/spf13/pflag"
```
rambohe-ch, line 34: please make go-imports.

rambohe-ch (approved):

PR #595, opened 2021-11-16
fix: cloud-node role is not master

cuisongliu:
` /kind bug ` 

 I use kubeadm install k8s is has role is master ,but for yurtctl init k8s not master role . For user ,this is not better display.
I think this feature should be retained.

```release-note


```

### DIFF
--- a/pkg/yurtctl/cmd/yurtinit/phases/labelCloudNode.go
+++ b/pkg/yurtctl/cmd/yurtinit/phases/labelCloudNode.go
@@ -19,6 +19,8 @@ package phases
 import (
 	"fmt"
 
+	"k8s.io/kubernetes/cmd/kubeadm/app/constants"
+
 	v1 "k8s.io/api/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/cmd/kubeadm/app/cmd/phases/workflow"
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@Peeknut @adamzhoul

rambohe-ch:
@cuisongliu please fix golangci lint error.

adamzhoul:
I see here adding the missing part of `traint` and label, after checking kubeadm implementation.

is this indeed a missing bug or something else is under consideration in the first place?
@rambohe-ch

rambohe-ch:
> I see here adding the missing part of `traint` and label, after checking kubeadm implementation.
> 
> is this indeed a missing bug or something else is under consideration in the first place? @rambohe-ch

@adamzhoul Thank you for your feedback. and i think that some introduction or background message should be added for reviewing this pull request. @cuisongliu would you be able to add some background about this pull request?

rambohe-ch:
@DrmagicE Do you have any comments about this pull request?

DrmagicE:
@rambohe-ch Seems this PR is intended to ensure the cloud nodes created by `yurtctl init` will preserve the same labels and taints as the nodes created by `kubeadm init`.  

I personally would agree to do so if there is no side-effect on the current implementation. I think preserving those behavior in `kubeadm` is more comfortable for the user.  Although I had not experienced with `yurtctl init` yet, as a user, I would assume that the behavior of `yurtctl init` is very close to `kubeadm init`, so I would expect to see those labels and taints after `yurtctl init`. 

Need more information and background for reviewing+1.

cuisongliu:
> @cuisongliu please fix golangci lint error.

done

zyjhtangtang:
Thanks @cuisongliu , I briefly explain the purpose of `yurtctl init` . In `yurtctl init`, kubernetes controlplane uses all-in-one static pod deployment.  The initial design of `yurtctl init` was to shield the complexity of Kubernete,  so that more developers in the field of edge computing could quickly set up an OpenYurt test cluster. The concept of `master node`  is not revealed here. On the one hand, the `master node` is used as a `cloud node`, and users do not need to consider the relationship between the `master node` and the `cloud node`. On the other hand, the deployment of OpenYurt components is easier. If users are familiar with OpenYurt's architecture and want to have more demands, they can use convert.

adamzhoul:
> Thanks @cuisongliu , I briefly explain the purpose of `yurtctl init` . In `yurtctl init`, kubernetes controlplane uses all-in-one static pod deployment. The initial design of `yurtctl init` was to shield the complexity of Kubernete, so that more developers in the field of edge computing could quickly set up an OpenYurt test cluster. The concept of `master node` is not revealed here. On the one hand, the `master node` is used as a `cloud node`, and users do not need to consider the relationship between the `master node` and the `cloud node`. On the other hand, the deployment of OpenYurt components is easier. If users are familiar with OpenYurt's architecture and want to have more demands, they can use convert.

thanks for introducing the background @cuisongliu 


as a design purpose, `yurtctl xxx` is used for the test, for users to experience openYurt quickly.
it is not designed for production.
so it removes many many things during installation.

compare with kubeadm, not only `yurtctl init` is simpler but  `yurtctl join` is simpler too.
it only focuses on what needs for openYurt to run to experience to develop.

adding this may affect nothing but more like origin kubeadm operate as I can see
it's good and feels better for developers, but introduce two questions to us :
1. do we have to make it looks like origin kubeadm operates? if so, is other parts missing ? fix them?
2. users may never notice that yurtctl is used for an experience, not for production. anything we should do?

rambohe-ch:
> 2\. . anything we should do?

@adamzhoul I think that to construct a production level K8s cluster is not OpenYurt job, and we need to point that out clearly. we will update OpenYurt readme sson.

adamzhoul:
by the way, 
do you think only using `kubeadm` is enough to construct a production-level edge K8s cluster?
say,  we have an edge node in a totally isolated network and `apiServer` has a public IP.

`kube-proxy` may not work well or some other problem
can operator cluster work on such situation and get the job done perfectly?

rambohe-ch:
> by the way, do you think only using `kubeadm` is enough to construct a production-level edge K8s cluster? say, we have an edge node in a totally isolated network and `apiServer` has a public IP.

@adamzhoul I think we can use kubeadm to install K8s control plane, and then install OpenYurt, then use `yurtctl join` to join node.

by the way,  we will share `OpenYurt Experience Center` demo in the next community meeting, and which is used for reducing the complexity of experience OpenYurt.

> `kube-proxy` may not work well or some other problem can operator cluster work on such situation and get the job done perfectly?

@SataQiu Would you be able to give comments about `yurtcluster` ability?

adamzhoul:
thanks for the reply @rambohe-ch 

only want to make sure, user can easily build an edge cluster
if there is trouble, we should help.

and, what about this pr?  
it there still a reason to merge?

adamzhoul:
oh about `yurtctl join`,  is this ok to be used to add production node?
since `yurtctl xxx` is not used for production.

SataQiu:
> > by the way, do you think only using `kubeadm` is enough to construct a production-level edge K8s cluster? say, we have an edge node in a totally isolated network and `apiServer` has a public IP.
> 
> @adamzhoul I think we can use kubeadm to install K8s control plane, and then install OpenYurt, then use `yurtctl join` to join node.
> 
> by the way, we will share `OpenYurt Experience Center` demo in the next community meeting, and which is used for reducing the complexity of experience OpenYurt.
> 
> > `kube-proxy` may not work well or some other problem can operator cluster work on such situation and get the job done perfectly?
> 
> @SataQiu Would you be able to give comments about `yurtcluster` ability?

To provide a better experience, we are building the [YurtCluster-Operator](https://github.com/openyurtio/yurtcluster-operator). It provides a declarative API to help you automatically convert a normal cluster (for example, created by kubeadm init/join) into an OpenYurt cluster. In this way, the OpenYurt cluster and the normal cluster nodes are guaranteed to have exactly the same role, and the whole conversion process is automated. We hope that this solution can be applied to the production environment in the feature.

PR #598, opened 2021-11-16
use https[s] proxy in docker for release build if necessary

gnunu:
/kind bug

When build release from a proxy env, http[s]_proxy is needed within the build container.

Fixes #

/assign @rambohe-ch 

```release-note
NONE
```


rambohe-ch:
@gnunu would you be able to add examples how to use these two new envs?

gnunu:
> @gnunu would you be able to add examples how to use these two new envs?

in environment behind proxy like our company, normally these env's would be set:
http_proxy=http://XXXX:PPP
https_proxy=https://XXXX:PPP

in building docker container, these env's are also needed for network access.

if no proxy is needed (no env's set), then nothing to worry due to [[ -n ${http_proxy+x} ]] guard.

PR #599, opened 2021-11-17
Refactor: tunnel-agent can connect new address of tunnel server

### LINKED ISSUE #412: [yurt tunnel]yurt-tunnel-agent can connect the new address of yurt-tunnel-server when service address changed.
when the yurt-tunne-server service address changed, yurt-tunnel-agent can use the new service address to re-connect yurt-tunnel-server automatically.

**Why is this needed**:
This issue comes from #411 , when the service address of yurt-tunnel-server changed, connection between yurt-tunnel-agent and yurt-tunnel-server will be broken. At present, we need to restart yurt-tunnel-agent pod manually to recover the connection.

**others**
/kind feature

YRXING:
/kind enhancement
/kind good-first-issue
/kind feature

when the service address of yurt-tunnel-server changed, connection between yurt-tunnel-agent and yurt-tunnel-server will be broken.

Fixes #412 

NONE
```release-note

```

### DIFF
--- a/pkg/yurttunnel/agent/anpagent.go
+++ b/pkg/yurttunnel/agent/anpagent.go
@@ -18,12 +18,16 @@ package agent
 
 import (
 	"crypto/tls"
+	"net"
+	"reflect"
 	"time"
 
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
-
+	"github.com/openyurtio/openyurt/pkg/yurttunnel/server/serveraddr"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 	anpagent "sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 )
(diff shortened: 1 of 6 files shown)


rambohe-ch:
@YRXING I think we need to check resources(like conn) have been garbaged or not in both side(tunnel server/client) when tunnel-server address changed, in order to avoid resource leak.

YRXING:
> @YRXING I think we need to check resources(like conn) have been garbaged or not in both side(tunnel server/client) when tunnel-server address changed, in order to avoid resource leak. maybe it's important to add some metrics for tunnel-server/agent for verifying resource(like conn, registered agents etc.) usage.

I learned that when the stop channel can be read, clientSet will shutdown all the connections.

pkg/yurttunnel/agent/anpagent.go
```diff
@@ -52,8 +61,41 @@ func (ata *anpTunnelAgent) Run(stopChan <-chan struct{}) {
...
+			case <-stopChan:
+				close(csStopChan)
+				return
+			case <-time.After(ata.probInterval):
+				addr, err := serveraddr.GetTunnelServerAddr(ata.client)
+				if err != nil {
+					klog.Infof("get tunnel server addr err: %+v", err)
+					continue
+				}
+				if !reflect.DeepEqual(addr, ata.tunnelServerAddr) {
```
adamzhoul, line 81: what about :

```
if reflect.DeepEqual(addr, ata.tunnelServerAddr) {
    continue
}
klog.Infof("tunnel server's ip has changed")
...
```

rambohe-ch:
> > @YRXING I think we need to check resources(like conn) have been garbaged or not in both side(tunnel server/client) when tunnel-server address changed, in order to avoid resource leak. maybe it's important to add some metrics for tunnel-server/agent for verifying resource(like conn, registered agents etc.) usage.
> 
> I learned that when the stop channel can be read, clientSet will shutdown all the connections.

@YRXING how about some metrics for checking source leak or not?

YRXING:
the konnectivity server already has some metrics for connections

// ServerMetrics includes all the metrics of the proxy server.
type ServerMetrics struct {
	latencies         *prometheus.HistogramVec
	frontendLatencies *prometheus.HistogramVec
	connections       *prometheus.GaugeVec
	httpConnections   prometheus.Gauge
	backend           *prometheus.GaugeVec
	pendingDials      *prometheus.GaugeVec
}

And in the agent side,  it is unnecessary to record connections. As we will create a new ClientSet to reconnect the server.

PR #603, opened 2021-11-18
Update node.go

guoguodan:
it may be null point,  add null point check.

### DIFF
--- a/pkg/yurtctl/cmd/convert/node.go
+++ b/pkg/yurtctl/cmd/convert/node.go
@@ -146,18 +146,24 @@ func (c *ConvertNodeOptions) SetupYurthub(workingMode util.WorkingMode) error {
 	if err != nil {
 		return err
 	}
-	kubeletConfPath = strings.Split(kubeletConfPath, "=")[1]
-	apiserverAddr, err := enutil.GetSingleContentFromFile(kubeletConfPath, apiserverAddrRegularExpression)
+	kubeletConfPathArr := strings.Split(kubeletConfPath, "=")
+	if len(kubeletConfPathArr) < 2 {
+		return fmt.Errorf("--kubeconfig-path format error")
+	}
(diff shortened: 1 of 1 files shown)


pkg/yurtctl/cmd/convert/node.go
```diff
@@ -146,18 +146,24 @@ func (c *ConvertNodeOptions) SetupYurthub(workingMode util.WorkingMode) error {
 	if err != nil {
 		return err
 	}
-	kubeletConfPath = strings.Split(kubeletConfPath, "=")[1]
-	apiserverAddr, err := enutil.GetSingleContentFromFile(kubeletConfPath, apiserverAddrRegularExpression)
+	kubeletConfPathArr := strings.Split(kubeletConfPath, "=")
+	if len(kubeletConfPathArr) < 2 {
```
Peeknut, line 150: I think  const var `kubeletConfigRegularExpression` used in `kubeletConfPath, err := enutil.GetSingleContentFromFile(c.KubeadmConfPath, kubeletConfigRegularExpression)` can guarantee this situation.

pkg/yurtctl/cmd/convert/node.go
```diff
@@ -146,18 +146,24 @@ func (c *ConvertNodeOptions) SetupYurthub(workingMode util.WorkingMode) error {
...
+		return fmt.Errorf("--kubeconfig-path format error")
+	}
+	apiserverAddr, err := enutil.GetSingleContentFromFile(kubeletConfPathArr[1], apiserverAddrRegularExpression)
 	if err != nil {
 		return err
 	}
-	apiserverAddr = strings.Split(apiserverAddr, " ")[1]
+	apiserverAddrArr := strings.Split(apiserverAddr, " ")
+	if len(apiserverAddrArr) < 2 {
```
Peeknut, line 158: same as before.

guoguodan: ok  thinks

PR #613, opened 2021-11-22
Condition is always 'false'

wangchenglong01:
/kind enhancement

Condition is always 'false'. Variable `commit` is assigned `gitCommit` whose value is fixed length.

Fixes #

```release-note

```

### DIFF
--- a/pkg/projectinfo/projectinfo.go
+++ b/pkg/projectinfo/projectinfo.go
@@ -31,19 +31,11 @@ var (
 )
 
 func ShortAgentVersion() string {
-	commit := gitCommit
-	if len(gitCommit) > 7 {
-		commit = gitCommit[:7]
-	}
-	return GetAgentName() + "/" + gitVersion + "-" + commit
+	return GetAgentName() + "/" + gitVersion + "-" + normalizeGitCommit(gitCommit)
 }
 
 func ShortServerVersion() string {
-	commit := gitCommit
-	if len(gitCommit) > 7 {
-		commit = gitCommit[:7]
-	}
-	return GetServerName() + "/" + gitVersion + "-" + commit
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@wangchenglong01 Thank you for raising pull request. how about add more information that condition is always `false`, like where `commit` is setup and length is fixed.

wangchenglong01:
> @wangchenglong01 Thank you for raising pull request. how about add more information that condition is always `false`, like where `commit` is setup and length is fixed.

Ok, I will add more  details.

wangchenglong01:
The more details has been added.

pkg/projectinfo/projectinfo.go
```diff
@@ -32,17 +32,11 @@ var (
 
 func ShortAgentVersion() string {
 	commit := gitCommit
-	if len(gitCommit) > 7 {
-		commit = gitCommit[:7]
-	}
 	return GetAgentName() + "/" + gitVersion + "-" + commit
```
rambohe-ch, line 35: how about use `normalizeGitCommit()` to optimize this snippet, then as following:
```
return GetAgentName() + "/" + gitVersion + "-" + normalizeGitCommit(gitCommit)
```

rambohe-ch:
> @rambohe-ch The more details has been added.

@wangchenglong01 Thanks for your reply. how about use `normalizeGitCommit` func to unify `gitCommit` usage. I have add a new comment, please take a look.

pkg/projectinfo/projectinfo.go
wangchenglong01: The latest code has been commited. Optimize this snippet.

rambohe-ch (approved):

PR #620, opened 2021-11-23
fix node autonomy annotation key

rambohe-ch:
/kind enhancement

fix node autonomy annotation key that use label prefix of project info

Fixes #

```release-note

```

#### other Note

other docs that use this annotation key will be updated by another pull request.


Peeknut:
The `AnnotationAutonomy` is used in another file: `pkg/yurtctl/constants/constants.go`.
It also needs to be modified.

rambohe-ch:
> The `AnnotationAutonomy` is used in another file: `pkg/yurtctl/constants/constants.go`. It also needs to be modified.

@Peeknut Thank you for pointing out the problem.

Peeknut:
As autonomy annotation key is used more than one place, how about define it in projectinfo and we can just use `GetAutonomyAnnotation()` ?

rambohe-ch:
> As autonomy annotation key is used more than one place, how about define it in projectinfo and we can just use `GetAutonomyAnnotation()` ?

@Peeknut It's a good idea, I will add autonomy annotation key definition in ProjectInfo.

PR #621, opened 2021-11-23
Remove k8s.io/kubernetes dependency for yurt-tunnel

### LINKED ISSUE #453: Eliminate dependency on k8s.io/kubernetes
https://github.com/kubernetes/kubernetes#to-start-using-k8s

> Use of the k8s.io/kubernetes module or k8s.io/kubernetes/... packages as libraries is not supported.

Relying directly on `k8s.io/kubernetes` didn't seem like a good idea, which made our project difficult to upgrade and maintain. I propose that eliminate dependency on `k8s.io/kubernetes`. 

This is likely to be a long-term task.

JcJinChen:
From kubernetes introduction: [ttps://github.com/kubernetes/kubernetes#to-start-using-k8s](https://github.com/kubernetes/kubernetes#to-start-using-k8s)

Fixes #453

```
none
```

#### other Note

```
none
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/stretchr/testify v1.4.0
 	github.com/vishvananda/netlink v1.0.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
+	golang.org/x/sys v0.0.0-20200519105757-fe76b779f299
 	golang.org/x/time v0.0.0-20200416051211-89c76fbcd5d1 // indirect
 	google.golang.org/grpc v1.27.0
 	gopkg.in/cheggaaa/pb.v1 v1.0.25
--- a/pkg/util/iptables/iptables.go
+++ b/pkg/util/iptables/iptables.go
@@ -1,0 +1,755 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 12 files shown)


rambohe-ch:
@JcJinChen maybe we need keep `iptables` from `k8s.io/kuberentes` to locate in a util directory.

JcJinChen:
> @JcJinChen maybe we need keep `iptables` from `k8s.io/kuberentes` to locate in a util directory.

ok, i will submit again after modification.

rambohe-ch:
@JcJinChen please fix license check and lint error.

pkg/yurthub/network/dummyif_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux
```
rambohe-ch, line 1: why add this comment?

pkg/yurthub/network/dummyif_unsupported.go
```diff
@@ -1,3 +1,4 @@
+//go:build !linux
```
rambohe-ch, line 1: the same as above comment

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/stretchr/testify v1.4.0
 	github.com/vishvananda/netlink v1.0.0
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d // indirect
+	golang.org/x/sys v0.0.0-20200519105757-fe76b779f299
```
rambohe-ch, line 27: why add this package?

pkg/yurthub/network/dummyif_test.go
JcJinChen: After coding, to format my code, so I execute gofmt -w .,then there is an extra line(go:build linux) in some files located under the path pkg/uitl/iptables. But I found that there is no such line in the kubernetes codekubernetes, so i delete this extre line(go:build linux) in those files, forgot to delete this line in pkg/yurthub/network/dummyif_test.go. I think this is caused by gofmt.

rambohe-ch (approved):

PR #628, opened 2021-11-25
ci(master): ln hotfix to build in macOS

cuisongliu:
/kind feature

hotfix build linux arch in macOS  'ln' not in this os

```release-note

in macOS exec `GOOS=linux GOARCH=amd64 make build WHAT=cmd/yurtctl`,but build after  the local dir ln  into  linux dir.
so I fix the `HOST_PLATFORM` default value is `$(go env GOOS)/$(go env GOARCH)`

```

### DIFF
--- a/hack/make-rules/build.sh
+++ b/hack/make-rules/build.sh
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 
 # Copyright 2020 The OpenYurt Authors.
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #     http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,8 +17,8 @@
 set -x
 
 YURT_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
-source "${YURT_ROOT}/hack/lib/init.sh" 
+source "${YURT_ROOT}/hack/lib/init.sh"
 
-HOST_PLATFORM=${HOST_PLATFORM:-"$(host_platform)"}
+HOST_PLATFORM=${HOST_PLATFORM:-"$(go env GOOS)/$(go env GOARCH)"}
 
 build_binaries "$@"


rambohe-ch:
@cuisongliu Thank you for raising pull request.
 In fact, os=linux and arch=amd64/arm64/arm have been supported now. 

@Congrool Would you like to take a look of this pull request?

Congrool:
> @cuisongliu Thank you for raising pull request. In fact, os=linux and arch=amd64/arm64/arm have been supported now.
> 
> @Congrool Would you like to take a look of this pull request?

Sure.

Congrool:
@cuisongliu , Hi, I saw
>this feature can let user at person computer build the linux binary and scp to linux server  

I think this will cause some protential problems in golang. We know `go build` is based on glibc. The binary built based on glibc of Mac may not work on Linux platform. Even between two Linux hosts, it still has the problem for Glibc version incompatibility and other reasons. Maybe it's a typical pratice in Java, but we don't recommand it in Go.  

And what do you think? @rambohe-ch  
Maybe we can provide a way to build a binary on other platform, but it's not recommended.

cuisongliu:
I think we can support it . like kubernetes can cross compile.

Congrool:
> I think we can support it . like kubernetes can cross compile.

@cuisongliu How does kubernetes solve the glibc incompatibility problem? Is it the responsibility users should take?

Congrool:
@cuisongliu Currently, openyurt support multi-arch cross compilation, and will copy the built binaries into docker images. And the docker image is based on linux. You can see the script [here](https://github.com/openyurtio/openyurt/blob/master/hack/lib/release-images.sh). So I think we should run the openyurt components into containers, and there's no need to support multi-platform.

cuisongliu:
we use `CGO_ENABLED=0`  can resolve cross compilation. we use `yurtctl`  binary better, same `kubeadm` is not container. because `yurtctl` binary  file is better.

Congrool:
@cuisongliu So, do you mean we should support cross compliation just for yurtctl?

cuisongliu:
> So, do you mean we should support cross compliation just for yurtctl?

I think like kubernetes , support binary and container. but  we use container is more.

cuisongliu:
There is really no need to support macOS.

Congrool:
>I think like kubernetes , support binary and container. but we use container is more.  

How does kubernetes run barely on host OS? As far as I know, only `kubeadm`, `kubectl` and `kubelet` run on host os.
I agree with support multi-platform cross compilation of yurtctl, that will enable users built yurtctl of other platform, such as buildting yurtctl linux on Mac.

rambohe-ch:
> > I think like kubernetes , support binary and container. but we use container is more.
> 
> How does kubernetes run barely on host OS? As far as I know, only `kubeadm`, `kubectl` and `kubelet` run on host os. I agree with support multi-platform cross compilation of yurtctl, that will enable users built yurtctl of other platform, such as buildting yurtctl for linux on Mac.

@cuisongliu @Congrool supoort multi-platform cross compilation of yurctl, agree +1.

cuisongliu:
> > > I think like kubernetes , support binary and container. but we use container is more.
> > 
> > 
> > How does kubernetes run barely on host OS? As far as I know, only `kubeadm`, `kubectl` and `kubelet` run on host os. I agree with support multi-platform cross compilation of yurtctl, that will enable users built yurtctl of other platform, such as buildting yurtctl for linux on Mac.
> 
> @cuisongliu @Congrool supoort multi-platform cross compilation of yurctl, agree +1. but other components like yurthub, yurt-tunnel etc. we can keep the current state.

yurhub we running on host os 。 hhhh. so , we can support `yurctl `,`yurthub` ,`yurt-node-servant` , agree it ?

rambohe-ch:
> > > > I think like kubernetes , support binary and container. but we use container is more.
> > > 
> > > 
> > > How does kubernetes run barely on host OS? As far as I know, only `kubeadm`, `kubectl` and `kubelet` run on host os. I agree with support multi-platform cross compilation of yurtctl, that will enable users built yurtctl of other platform, such as buildting yurtctl for linux on Mac.
> > 
> > 
> > @cuisongliu @Congrool supoort multi-platform cross compilation of yurctl, agree +1. but other components like yurthub, yurt-tunnel etc. we can keep the current state.
> 
> yurhub we running on host os 。 hhhh. so , we can support `yurctl `,`yurthub` ,`yurt-node-servant` , agree it ?

@cuisongliu If you want to support other os like darwin, you can add an env(like YURT_BUILD_PLATFORM) so end user can  define the OS and ARCH. and you only need to modify func `get_binary_dir_with_arch` as following:
```
get_binary_dir_with_arch(){
     if [ ! -z ${YURT_BUILD_PLATFORM} ]; then
           echo $1/$(YURT_BUILD_PLATFORM)/
     else
          echo $1/$(go env GOOS)/$(go env GOARCH)/
     fi  
}
```

and when you want build yurthub or yurtctl on darwin, you can use `make build WHAT=cmd/yurtctl  YURT_BUILD_PLATFORM=linux/amd64` to make linux binary.

cuisongliu:
> > > > > I think like kubernetes , support binary and container. but we use container is more.
> > > > 
> > > > 
> > > > How does kubernetes run barely on host OS? As far as I know, only `kubeadm`, `kubectl` and `kubelet` run on host os. I agree with support multi-platform cross compilation of yurtctl, that will enable users built yurtctl of other platform, such as buildting yurtctl for linux on Mac.
> > > 
> > > 
> > > @cuisongliu @Congrool supoort multi-platform cross compilation of yurctl, agree +1. but other components like yurthub, yurt-tunnel etc. we can keep the current state.
> > 
> > 
> > yurhub we running on host os 。 hhhh. so , we can support `yurctl `,`yurthub` ,`yurt-node-servant` , agree it ?
> 
> @cuisongliu If you want to support other os like darwin, you can add an env(like YURT_BUILD_PLATFORM) so end user can define the OS and ARCH. and you only need to modify func `get_binary_dir_with_arch` as following:
> 
> ```
> get_binary_dir_with_arch(){
>      if [ ! -z ${YURT_BUILD_PLATFORM} ]; then
>            echo $1/$(YURT_BUILD_PLATFORM)/
>      else
>           echo $1/$(go env GOOS)/$(go env GOARCH)/
>      fi  
> }
> ```
> 
> and when you want build yurthub or yurtctl on darwin, you can use `make build WHAT=cmd/yurtctl YURT_BUILD_PLATFORM=linux/amd64` to make linux binary.

fix ci error. approved it . this ci not auto ???

Congrool:
Well, the ci is blocked because you're a first-time contributor. So it needs approval of maintainers.

hack/lib/build.sh
```diff
@@ -20,9 +20,6 @@ readonly YURT_ALL_TARGETS=(
     yurtctl
     yurt-node-servant
     yurthub
-    yurt-controller-manager
-    yurt-tunnel-server
-    yurt-tunnel-agent
```
Congrool, line 25: `release-images.sh` depends on the `build.sh`. Deleting targets here will make `release-images.sh` fail.

hack/lib/build.sh
```diff
@@ -78,20 +81,23 @@ build_binaries() {
...
-          -gcflags "${gcflags:-}" ${goflags} $YURT_ROOT/cmd/$(canonicalize_target $binary)
+    for platform in "${SUPPORTED_PLATFORMS[@]}"; do
+      for binary in "${targets[@]}"; do
+           OS=${platform%/*}
+           ARCH=${platform##*/}
+           local target_bin_dir=${YURT_LOCAL_BIN_DIR}/$OS/$ARCH/
+           mkdir -p ${target_bin_dir}
+           cd ${target_bin_dir}
+           echo "Building ${binary} $OS $ARCH"
+           GOOS=$OS GOARCH=$ARCH CGO_ENABLED=0 go build -o $(get_output_name $binary) \
```
Congrool, line 92: `release-images.sh` also specifies `go envs` and calls the `build.sh` once per arch. The loop here will result in building redundant binaries. In addition, setting `GOARCH` and other go env here will be in conflict with what `release-images.sh` has set.

Congrool:
You can test your code locally to check if it's correct. Please see the [tutorial](https://github.com/openyurtio/openyurt/blob/master/docs/tutorial/yurt-e2e-test.md). CI also runs in this way.

hack/lib/build.sh
Congrool: Well，according to CI logs, there's no problem here.

hack/lib/build.sh
Congrool: Please fix this problem.

Congrool:
@cuisongliu Oh, sorry, it seems that original `build.sh` has already supportted cross compliation. See the [developer-guide.md](https://github.com/openyurtio/openyurt/blob/master/docs/developer-guide.md). 
We can build yurtctl for linux on Mac with command:
```
GOOS=linux GOARCH=amd64 make build WHAT=cmd/yurtctl
```
Maybe we just need to add `CGO_ENABLED=0` before `go build`.

Congrool (commented):
good, lgtm

Issue #597, opened 2021-11-16
[feature request] Add an option to "yurtctl join" to enable "node autonomy" automatically.


DrmagicE:
**What would you like to be added**:
Currently, if the user wants to enable node autonomy for a new node joined by `yurtctl join`. The user had to call `yurtctl markautonomous` after `yurtctl join`, which is inconvenient.

I think the whole process can be simplified by adding an option to `yurtctl join` to enable ”node autonomy“ automatically.

**Why is this needed**:

**others**
/kind feature

rambohe-ch:
@DrmagicE It's a good idea to add parameter for `yurtctl join` for marking autonomous. @zyjhtangtang Do you have any comments?

zyjhtangtang:
Thanks @DrmagicE , I agree. Can you submit a PR to  add this parameter.

DrmagicE:
@zyjhtangtang Hi, I would like to, but I am afraid I had to finish some jobs in hands (https://github.com/openyurtio/openyurt/issues/546 and https://github.com/openyurtio/openyurt/issues/527) before starting this one. 
@rambohe-ch How about we label this issue as a good-first issue?

rambohe-ch:
> @zyjhtangtang Hi, I would like to, but I am afraid I had to finish some jobs in hands (#546 and #527) before starting this one. @rambohe-ch How about we label this issue as a good-first issue?

@DrmagicE done

adamzhoul:
I am working on https://github.com/openyurtio/openyurt/issues/545 
I can pick up this by the way 

/assign @adamzhoul

Issue #545, opened 2021-11-01
[feature request] yurtctl join use node-servant to finish some job


adamzhoul:
it's a job after https://github.com/openyurtio/openyurt/pull/516

in `yurtctl join`, what it does to the node itself is included in the `node-servant`
so we should remove that part of the code and use `node-servant` package instead

/help

zyjhtangtang:
`yurtctl join` uses `yurt-hub` to generate node certificates, and does not include the logic of `node-servant`. If `node-servant` is introduced, the `yurtctl join` process will become complicated. Therefore, I don’t think it is necessary to introduce `node-servant` in `yurtctl join`.

adamzhoul:
thanks for pointing this out @zyjhtangtang 
I start to read this part of the code recently.

the core thought is,  any operating directly to nodes like `revise kubelet` or `install yurthub` can use node-servant.
maybe by import pkg.
other controller logic parts can remain unchanged.

zyjhtangtang:
Oh, `yurtctl join` has the logic to `install yurthub`, this can use `node servant`.  But, the configuration of `kubelet` is written directly and does not need to be revised. @adamzhoul

adamzhoul:
by the way, this may introduce a special case:

    node  operated by  `yurtctl convert` can retrieve by ` yurtctl revert`
    node  operated by  `yurtctl join` can retrieve by `yurtctl reset`

in all, both `convert` and `join` will create an openyurt kind node.
but their creation may be different.

zyjhtangtang:
Are the configuration parameters of `kubelet` different? @adamzhoul

adamzhoul:
Here is now what they do to kubelet：

yurtctl convert

1.  create `kubeconfig` file  with full path:  /var/lib/openyurt/kubelet.conf
2.  backup `KubeadmConfPath` conf file:  `cp 10-kubeadm.conf 10-kubeadm.conf.bk`
3.  update `10-kubeadm.conf`, set --kubeconfig=/var/lib/openyurt/kubelet.conf

node-servant

1.  create `kubeconfig` file  with full path:  /var/lib/openyurt/kubelet.conf
2.  append file `/var/lib/kubelet/kubeadm-flags.env`, add `--kubeconfig=/var/lib/openyurt/kubelet.conf` 

append content effect as  overwrite the params

yurtctl join

1. create `kubeconfig` file  with full path:  /etc/kubernetes/kubelet.conf

adamzhoul:
[...]

PR #630, opened 2021-11-25
[WIP] yurtcl join use node-servant

adamzhoul:
> /kind enhancement

https://github.com/openyurtio/openyurt/issues/597 [feature request] Add an option to "yurtctl join" to enable "node autonomy" automatically.

https://github.com/openyurtio/openyurt/issues/545 [feature request] yurtctl join use node-servant to finish some job 
https://github.com/openyurtio/openyurt/issues/516  add node-servant


pkg/yurtctl/cmd/join/phases/postcheck.go
```diff
@@ -103,14 +103,21 @@ func checkYurthubHealthz() error {
...
 	}
+
+	if markAutonomous {
+		if err := apiclient.PatchNode(client, cfg.NodeRegistration.Name, annotateNodeWithAutonomousNode); err != nil {
+			return err
+		}
+	}
+
 	if err := apiclient.PatchNode(client, cfg.NodeRegistration.Name, func(n *v1.Node) {
 		n.Labels[projectinfo.GetEdgeWorkerLabelKey()] = "true"
```
rambohe-ch, line 122: how about move `makeAutonomous` at here?

adamzhoul: Do you mean using `MarkAutonomousOptions` int `pkg/yurtctl/cmd/markautonomous/markautonomous.go`
```
	co := NewMarkAutonomousOptions()
        o.RunMarkAutonomous()
```
it's definitely better using this.

But, it has serval questions we have to answer:
1. a `kubeconf` file is required, and we may use `kubelet.conf`. Is this ok?
2.  join package import markautonomous package is ok but not advised I think.

rambohe-ch: @adamzhoul I'm sorry that the comment is not clear. i mean that move the code line 114~120 after line 122, so reduce one node patch request.

adamzhoul:
don't have enough time to have a full test.
besides, the community is going to remove k8s.io/kubernetes.
this will be affected.

so close this pr for now.

PR #631, opened 2021-11-26
yurthub verify bootstrap ca on start

gnunu:
/kind bug

verify existed ca file if any, and do cleanup if necessary (mismatch). This will fix the openyurt convert failure due to stale contents left over in the yurthub root dir (/var/lib/yurthub).

Fixes #

/assign @rambohe-ch 

```release-note
NONE
```

### DIFF
--- a/pkg/yurthub/certificate/hubself/cert_mgr.go
+++ b/pkg/yurthub/certificate/hubself/cert_mgr.go
@@ -17,6 +17,7 @@ limitations under the License.
 package hubself
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
@@ -24,6 +25,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/yurthub/certificate/hubself/cert_mgr.go
```diff
@@ -258,6 +274,24 @@ func (ycm *yurtHubCertManager) initCaCert() error {
...
 	}
 
+	if caExisted {
+		var existedCABytes []byte
+		var err error
+		existedCABytes, err = ioutil.ReadFile(caFile)
+		if err != nil {
+			klog.Infof("could not read existed %s ca cert, %v", caFile, err)
+		}
+		if (bytes.Compare(clusterCABytes, existedCABytes) == 0) {
```
rambohe-ch, line 284: when yurthub startup if cloud-edge network disconnected, so yurthub can not get `clusterCABytes` from kube-apiserver, leading to yurthub startup failure. so we can not use `CABytes` to verify whether remove certificates or not, and maybe we need to use `bootstrap token` from `yurthub parameter --join-token` and `/var/lib/yurthub/bootstrap-hub.conf` to verify.

gnunu: ok. Let me do bootstrap token way.

rambohe-ch:
@gnunu A good job to remove pki based on verifying ca file. please fix the above golangci lint error.

gnunu:
@rambohe-ch thanks!

rambohe-ch:
@gnunu I think that maybe we can compare the remote server address from yurthub startup parameter `--serveraddr` and server address of `/var/lib/yurthub/bootstrap-hub.conf`. if yurthub is used to connect to a new kube-apiserver, the remote server address will be changed, so the old certificates should be removed.

rambohe-ch (approved):

PR #637, opened 2021-11-29
Proposal: enhance cluster networking capabilities.

DrmagicE:
/kind feature

Fixes #

```release-note

```


Congrool:
Hi, it's really an exciting feature, but I still have some questions here.
1. How do nodepools get their podCIDR? Is it the responsibility of this network solution or dependent on other components, such as yurt-app-manager?  
2. We know that flannel will also allocate the podCIDR for each node, how to ensure that the nodepool podCIDR contains podCIDRs of all its member nodes. (It may be one of the CNI compatibility problems that we will encounter later）
3. If we select some nodes to be a new nodepool and want to join it into the network, do original pods need to restart to get their new podIP according to the nodepool podCIDR? Or, on the contary, is it the original podIPs that determine the nodepool podCIDR? (In the later case, we can also solve the problem2 if there's no conflict)
3. Host network subnet confilt is really a diffcult problem to solve, because we cannot determine which nodepool to send the packet. I think, in such scenario, the network situation can not be absolutely transparent to the application. Maybe we can support it in the application layer and let the application itself make the determination.

DrmagicE:
@Congrool  Hi, thanks for your feedback.
> 1. How do nodepools get their podCIDR? Is it the responsibility of this network solution or dependent on other components, such as yurt-app-manager?
> 2. We know that flannel will also allocate the podCIDR for each node, how to ensure that the nodepool podCIDR contains podCIDRs of all its member nodes. (It may be one of the CNI compatibility problems that we will encounter later

`podCIDR` is a field of the node resource. We can track all `podCIDR` belonging to a nodepool by list/watch mechanism.
```
[root@master-1 /]# kubectl get nodes master-1 -oyaml | grep podCIDR
        f:podCIDR: {}
        f:podCIDRs:
  podCIDR: 10.244.0.0/24
  podCIDRs:
```
But we should notice that not all CNI respects `podCIDR`.  If we use flannel, we will be fine because flannel respect `podCIDR`.
However, some CNI like calico does not respect `podCIDR`. We have to figure out another way to get `podCIDR` of a node for such CNI. In other words, how to get `podCIDR` of a node may vary from CNI to CNI. 

I suggest we start with flannel CNI, which is simple and widely use in OpenYurt.


> 3. If we select some nodes to be a new nodepool and want to join it into the network, do original pods need to restart to get their new podIP according to the nodepool podCIDR? Or, on the contary, is it the original podIPs that determine the nodepool podCIDR? (In the later case, we can also solve the problem2 if there's no conflict)

No they don't need a restart. The solution introduced in this proposal will not change the `podCIDR` on the node, and will not affect IP allocation of pods, IP allocation is still managed by the CNI.

> 4. Host network subnet conflict is really a diffcult problem to solve, because we cannot determine which nodepool to send the packet. I think, in such scenario, the network situation can not be absolutely transparent to the application. Maybe we can support it in the application layer and let the application itself make the determination.

Yes, still trying to figure out a way to solve the problem. It is an inevitable issue If we want this solution to replace the YurtTunnel.

rambohe-ch:
@DrmagicE @Congrool podCIDR is allocated to every node by rangeAllocator in `kube-controller-manager`, and NodePool contains name info for nodes that resides in the NodePool, so YurtGateway can calculate all podCIDRs through list/watch nodePool and nodes, and not need to consider the cni solution is flannel or calico.

rambohe-ch:
@DrmagicE Thanks for your feedback. if calico IPAM does not belong to `node.Spec.PodCIDR`, maybe we need to consider the other solution instead of `PodCIDR`.  At first, we can start by flannel.

docs/proposals/20211123-enhancement-of-cluster-networking.md
```diff
@@ -0,0 +1,306 @@
...
+
+#### 1. YurtGateway
+A new component that is responsible for the VPN setup and teardown. The nodes YurtGateway deployed on is called gateway node.
+YurtGateway runs on specifically designated nodes in a nodepool. There will be only one active YurtGateway instance at a time in a nodepool.
+
+#### 2. YurtRouter
+A new component that is responsible for configuring route table, route policy and other network-related configurations on nodes.
+YurtRouter is a daemonset and is deployed on all nodepools that are participating in the VPN tunnel. 
+
+### Network Reachability Requirement
```
yixingjia, line 112: This requirement is pretty challenge on edge side, Do you consider running a service on cloud to help establish the tunnel between different node-pools?

rambohe-ch: @yixingjia edge yurt-gateway can connect with each other through cloud yurt-gateway. and the detail info will be added after next community meeting discussion.

yixingjia: Ok, then the yurt-gateway on cloud can consider to change name like yurt-cloud-gateway and the gateway on the edge called yurt-edge-gateway. SDN have similar implementation for those kinds of requirements.

adamzhoul:
hi @DrmagicE 

can you add some detail about how to config vxlan to redirect traffic to the gateway node
more importantly how vxlan device route packet to IPsec on the gateway node

I'm a little confused.

thanks

docs/proposals/20211123-enhancement-of-cluster-networking.md
rambohe-ch: @yixingjia we will discuss component name in the other proposal.

docs/proposals/20211123-enhancement-of-cluster-networking.md
```diff
@@ -0,0 +1,306 @@
...
+ready to take over should the active instance fail.
+
+### Node Autonomy Consideration
+Yurthub should cache apiserver response for YurtGateway and YurtRouter.
+When the edge-cloud connection gets blocked, YurtGateway can not get the real-time status of each other via 
+list/watch mechanism.
+Thus, YurtGateway is not aware of failover of the other side, and when failover occurs, the VPN tunnel is broken.
+
+To fix that:
+1. YurtGateway should be able to detect the VPN status. Once it detects failover on the other side, it will try to connect the other backup.
```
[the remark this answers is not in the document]
DrmagicE: @vincent-pli Hi, sorry for the late reply. Base on discussions on our latest community meeting, we may not support H/A  in our first release.We need to think more carefully about how to achieve  H/A, especially in node autonomy circumstances. 

Welcome to get involved and share your ideas.

DrmagicE:
> hi @DrmagicE
> 
> can you add some detail about how to config vxlan to redirect traffic to the gateway node more importantly how vxlan device route packet to IPsec on the gateway node
> 
> I'm a little confused.
> 
> thanks

Hi, "redirect vxlan traffic to gateway node" is based on IP packet forwarding. We can configure the IP route table of non-gateway nodes via the `ip r add` command.  I think from the perspective of the Linux routing table, there is not much difference between routing vxlan packets and normal IP packets.

For vxlan mode, the routing rules are as same as host-gw mode. Here is an example shown in the proposal:
```
$ ip rule
0:	from all lookup local
# Set up route policy for pod CIDR of nodepoolB to use cross_edge_table.
32764:	from all to 10.244.2.0/24 lookup cross_edge_table
32765:	from all to 10.244.3.0/24 lookup cross_edge_table
32766:	from all lookup main
32767:	from all lookup default
$ ip r list table cross_edge_table
# 10.0.20.13 is the private IP of the local gateway node.
# We need to set a smaller MTU for IPsec traffic, the concrete number is yet to be determined.
default via 10.0.20.13 dev eth0 mtu 1400
```

DrmagicE:
@rambohe-ch Ok.

adamzhoul:
> > hi @DrmagicE
> > can you add some detail about how to config vxlan to redirect traffic to the gateway node more importantly how vxlan device route packet to IPsec on the gateway node
> > I'm a little confused.
> > thanks
> 
> Hi, "redirect vxlan traffic to gateway node" is based on IP packet forwarding. We can configure the IP route table of non-gateway nodes via the `ip r add` command. I think from the perspective of the Linux routing table, there is not much difference between routing vxlan packets and normal IP packets.
> 
> For vxlan mode, the routing rules are as same as host-gw mode. Here is an example shown in the proposal:
> 
> ```
> $ ip rule
> 0:	from all lookup local
> # Set up route policy for pod CIDR of nodepoolB to use cross_edge_table.
> 32764:	from all to 10.244.2.0/24 lookup cross_edge_table
> 32765:	from all to 10.244.3.0/24 lookup cross_edge_table
> 32766:	from all lookup main
> 32767:	from all lookup default
> $ ip r list table cross_edge_table
> # 10.0.20.13 is the private IP of the local gateway node.
> # We need to set a smaller MTU for IPsec traffic, the concrete number is yet to be determined.
> default via 10.0.20.13 dev eth0 mtu 1400
> ```

thanks for the reply @DrmagicE .  and sorry for the late.

personal tried to lead traffic from nodeA -> nodeB using vxlan device.
and encountered problems.

1. on both nodes, add vxlan device 

```
ip link add vxlan type vxlan \
    id 1 \
    dstport 4789 \
    local ${localMachineIp} \
    nolearning \
    dev eth0
```

2.  add a testIp and config to using vxlan  on nodeA

```
# manual add a route to test traffic
 ip route add ${githubIp} dev vxlan

bridge fdb append ${vxlanMacNodeB} dev vxlan dst ${nodeBIP}
ip neigh add ${githubIp}  lladdr ${vxlanMacNodeB} dev vxlan
```

3. ping $githubIp  on nodeA

 `tcpdump -i vxlan`  on nodeB
 capture packet with ICMP request, no ICMP reply.
 capture no packet on eth0

in all, simply adding vxlan、route table is not enough to redirect traffic.

what I did to make it work:
1. give both  vxlan device IP, then we can capture packet on eth0
2. add `iptables -t nat  -A POSTROUTING -s {vxlanIP}/16 -j MASQUERADE` to do SNAT when going through eth0

maybe I make things complicated.
if I misunderstood your point or you have a better solution
please let me know,  thanks. 


Oh, by the way. I can't simply use the `route table` to redirect traffic from nodeA -> nodeB.
because in cloud env, there arp is answered by the gateway device, packet out of nodeA arrived at gateway will never find nodeB(dst IP is podIP, not nodeB IP). this is why we have to rely on vxlan.

rambohe-ch:
@adamzhoul we can discuss the details of raven at https://github.com/openyurtio/raven, and i will merge this pull request at first.

PR #662, opened 2021-12-08
ci(master): add golint makefile

cuisongliu:
/kind feature

add makefile ci tools and fix error golint in macos

Fixes #

```release-note

```

### DIFF
--- a/pkg/util/iptables/iptables_unsupported_test.go
+++ b/pkg/util/iptables/iptables_unsupported_test.go
@@ -23,13 +23,13 @@ import (
 func Test_grabIptablesFileLock(t *testing.T) {
 	err := grabIptablesFileLock(nil)
 	if err != nil {
-		t.Errorf(err.Error())
+		t.Log(err.Error())
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/util/iptables/iptables_unsupported_test.go
```diff
@@ -0,0 +1,35 @@
...
+limitations under the License.
+*/
+
+package iptables
+
+import (
+	"testing"
+)
+
+func Test_grabIptablesFileLock(t *testing.T) {
```
rambohe-ch, line 23: why add `iptables_unsupported_test.go` file?

cuisongliu: in linux is ok, in Mac  make lint is error ,because the func not used. we can ignore it.

rambohe-ch: ok, it sounds reasonable to add this file.

rambohe-ch (approved):

PR #690, opened 2021-12-20
user can specifies the certificate directory

### LINKED ISSUE #409: [yurt-tunnel]user can specifies the certificate directory
At present, the certificate directory for yurt-tunnel components is fixed as following:
- yurt-tunnel-server: /var/lib/yurt-tunnel-server/pki
- yurt-tunnel-agent: /var/lib/yurt-tunnel-agent/pki

I'd like to add a parameter for yurt-tunnel to specify certificate directory, so certificates can be managed according to the actual situation of the user. 

**Why is this needed**:
In order to make the certificates of yurt-tunnel components can be saved in any directory that user wants to save.

**others**
/kind feature

YRXING:
#### What type of PR is this?
 /kind enhancement
 /kind good-first-issue
 /kind feature

In order to make the certificates of yurt-tunnel components can be saved in any directory that user wants to save.

Fixes #409 

user can use `--cert-dir` parameter to specify the directory of certificate about both yurt tunnel server and agent.

### DIFF
--- a/cmd/yurt-tunnel-agent/app/config/config.go
+++ b/cmd/yurt-tunnel-agent/app/config/config.go
@@ -16,7 +16,14 @@ limitations under the License.
 
 package config
 
-import "k8s.io/client-go/kubernetes"
+import (
+	"fmt"
+
+	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	"github.com/openyurtio/openyurt/pkg/yurttunnel/constants"
+
+	"k8s.io/client-go/kubernetes"
+)
 
 // Config is the main context object for yurttunel-agent
 type Config struct {
@@ -26,6 +33,7 @@ type Config struct {
 	Client           kubernetes.Interface
 	AgentIdentifiers string
 	AgentMetaAddr    string
+	CertDir          string
 }
 
 type completedConfig struct {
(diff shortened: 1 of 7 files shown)


cmd/yurt-tunnel-agent/app/config/config.go
```diff
@@ -42,5 +43,8 @@ type CompletedConfig struct {
 func (c *Config) Complete() *CompletedConfig {
 	cc := completedConfig{c}
 
+	if cc.CertDir == "" {
+		cc.CertDir = "/var/lib/yurttunnel-agent/pki"
```
rambohe-ch, line 47: how about use `fmt.Sprintf(constants.YurttunnelAgentCertDir, projectinfo.GetAgentName())`?

cmd/yurt-tunnel-server/app/config/config.go
```diff
@@ -62,5 +63,8 @@ func (c *Config) Complete() *CompletedConfig {
 	if cc.InterceptorServerUDSFile == "" {
 		cc.InterceptorServerUDSFile = "/tmp/interceptor-proxier.sock"
 	}
+	if cc.CertDir == "" {
+		cc.CertDir = "/var/lib/yurttunnel-server/pki"
```
rambohe-ch, line 67: how about use `fmt.Sprintf(constants.YurttunnelServerCertDir, projectinfo.GetServerName())`?

rambohe-ch (approved):

PR #694, opened 2021-12-23
[CI Enhancement] Add gci linter check

Congrool:
#### What type of PR is this?
/kind enhancement

After intergating gci linter into CI workflow, we can check the import package format automatically, including
1. repeated import
2. wrong 3-rd part package import order
3. no empty line before import with comment

More import format to be checked by gci linter can be seen in the [README](https://github.com/daixiang0/gci) of gci repo.

Fixes #

#### other Note

It's recommanded that developer should always run `make lint` at local host before submitting a pr.
Developers can run `make lint` at local host to get more info about lint error if any. There're more details than what github action outputs, including the file name and line number which cause this error.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: ${{ env.GOLANGCI_VERSION }}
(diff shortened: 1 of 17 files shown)


rambohe-ch:
@Congrool Very appreciate for your job. It is very helpful to add gci linter checker for OpenYurt.

PR #696, opened 2021-12-26
ingress: add nodepool endpoints filtering for nginx ingress controller

zzguang:
When users access service through nodepool ingress, ingress controller
will balance the users rquests to the endpoints binding to the service,
the ingress controller in a nodepool should only balance the requests
to endpoints belonging to this nodepool, so the service endpoints need
to be filtered to exclude endpoints from other endpoints.

Note:
Currently we leverage yurthub data filtering framework to achieve the
purpose, which is actually a node level sidecar solution, if a nodepool
level data filtering sidecar is implemented in future, nodepool ingress
data filtering is suggested to switch to that solution.

Signed-off-by: zhenggu1 <zhengguang.zhang@intel.com>

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/ingresscontroller/filter.go
+++ b/pkg/yurthub/filter/ingresscontroller/filter.go
@@ -1,0 +1,168 @@
+/*
+Copyright 2021 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ingresscontroller
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
+	filterutil "github.com/openyurtio/openyurt/pkg/yurthub/filter/util"
+	"github.com/openyurtio/openyurt/pkg/yurthub/kubernetes/serializer"
+	"github.com/openyurtio/openyurt/pkg/yurthub/util"
+	yurtinformers "github.com/openyurtio/yurt-app-manager-api/pkg/yurtappmanager/client/informers/externalversions"
+	appslisters "github.com/openyurtio/yurt-app-manager-api/pkg/yurtappmanager/client/listers/apps/v1alpha1"
+)
+
+// Register registers a filter
+func Register(filters *filter.Filters) {
+	filters.Register(filter.IngressControllerFilterName, func() (filter.Interface, error) {
+		return NewFilter(), nil
+	})
+}
+
+func NewFilter() *ingressControllerFilter {
+	return &ingressControllerFilter{
+		Approver:    filter.NewApprover("nginx-ingress-controller", "endpoints", []string{"list", "watch"}...),
+		workingMode: util.WorkingModeEdge,
+		stopCh:      make(chan struct{}),
+	}
+}
+
+type ingressControllerFilter struct {
+	*filter.Approver
+	serviceLister     listers.ServiceLister
+	serviceSynced     cache.InformerSynced
+	nodepoolLister    appslisters.NodePoolLister
+	nodePoolSynced    cache.InformerSynced
+	nodeGetter        filter.NodeGetter
+	nodeSynced        cache.InformerSynced
+	nodeName          string
+	serializerManager *serializer.SerializerManager
+	workingMode       util.WorkingMode
+	stopCh            chan struct{}
+}
+
+func (ssf *ingressControllerFilter) SetWorkingMode(mode util.WorkingMode) error {
+	ssf.workingMode = mode
+	return nil
+}
+
+func (ssf *ingressControllerFilter) SetSharedInformerFactory(factory informers.SharedInformerFactory) error {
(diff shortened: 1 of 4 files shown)


pkg/yurthub/filter/ingresscontroller/handler.go
```diff
@@ -0,0 +1,179 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	AnnotationIngressControllerKey           = "openyurt.io/topologyKeys"
+	AnnotationIngressControllerValueNode     = "kubernetes.io/hostname"
+	AnnotationIngressControllerValueZone     = "kubernetes.io/zone"
+	AnnotationIngressControllerValueNodePool = "openyurt.io/nodepool"
```
rambohe-ch, line 37: L34~L37 vars are unused.

pkg/yurthub/filter/ingresscontroller/handler.go
```diff
@@ -0,0 +1,179 @@
...
+	svc, err := fh.serviceLister.Services(endpoints.Namespace).Get(svcName)
+	if err != nil {
+		klog.Infof("skip reassemble endpoints, failed to get service %s/%s, err: %v", endpoints.Namespace, svcName, err)
+		return endpoints
+	}
+	// discard endpoints if service type is LoadBalancer
+	if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
+		klog.Infof("endpoints(%s/%s) of load balancer service is discarded", endpoints.Namespace, svcName)
+		return nil
+	}
```
rambohe-ch, line 133: LB service maybe left on cloud nodePool, so L130~L133 should be removed.

pkg/yurthub/filter/ingresscontroller/handler.go
```diff
@@ -0,0 +1,179 @@
...
+					isNodePoolValidEps = true
+					newEpsA = append(newEpsA, endpoints.Subsets[i].Addresses[j])
+					klog.Infof("endpoints/%s address IP/%s is valid to nodepool %s", svcName, endpoints.Subsets[i].Addresses[j].IP, nodePoolName)
+				}
+			}
+			endpoints.Subsets[i].Addresses = newEpsA
+			newEpsA = nil
+		}
+		if !isNodePoolValidEps {
+			return nil
```
rambohe-ch, line 165: after reconstruct endpoints, i think we need to just return `endpoints`, why need to verify `isNodePoolValidEps ` and return nil if `isNodePoolValidEps ` is false?

pkg/yurthub/filter/ingresscontroller/filter.go
```diff
@@ -0,0 +1,148 @@
...
+	ssf.nodeName = nodeName
+
+	return nil
+}
+
+func (ssf *ingressControllerFilter) SetStorageWrapper(s cachemanager.StorageWrapper) error {
+	if len(ssf.nodeName) == 0 {
+		return fmt.Errorf("node name for ingressControllerFilter is not ready")
+	}
+
```
rambohe-ch, line 89: please reference the ServiceTopology filter, we need to consider node that works as cloud mode,  then we can not use `StorageWrapper` to construct `nodeSynced ` and `nodeGetter `.

rambohe-ch:
@zzguang please fix golangci lint error.

pkg/yurthub/filter/ingresscontroller/handler.go
zzguang: If the endpoints is not valid to the nodepool the node belonging to, it will not be reconstructed and keep unchanged. For these kinds of endpoints I think we should ignore them, so I return nil in my implementation. Not sure whether I have some misunderstanding about it, if yes, please help me to clarify it, thanks!

zzguang: If the endpoints is not valid to the nodepool the node belonging to, it will not be reconstructed and keep unchanged. For these kinds of endpoints I think we should ignore them, so I return nil in my implementation. Not sure whether I have some misunderstanding about it, if yes, please help me to clarify it, thanks!

pkg/yurthub/filter/ingresscontroller/handler.go
zzguang: Yes, forgot to polish it, thanks!

pkg/yurthub/filter/ingresscontroller/handler.go
zzguang: Okay, I will clean it.

pkg/yurthub/filter/ingresscontroller/filter.go
zzguang: Okay, I will update it to the latest design.

pkg/yurthub/filter/ingresscontroller/handler.go
rambohe-ch: @zzguang Thanks for your explanation, i got it.

rambohe-ch: another question is: endpoints.Subset.NotReadyAddresses should also be reassembled?

pkg/yurthub/filter/ingresscontroller/handler.go
```diff
@@ -0,0 +1,174 @@
...
+		nodePool, err := fh.nodePoolLister.Get(nodePoolName)
+		if err != nil {
+			klog.Infof("skip reassemble endpoints, failed to get nodepool %s, err: %v", nodePoolName, err)
+			return endpoints
+		}
+		var newEpSubsets []v1.EndpointSubset
+		for i := range endpoints.Subsets {
+			endpoints.Subsets[i].Addresses = filterValidEndpointsAddr(endpoints.Subsets[i].Addresses, nodePool)
+			endpoints.Subsets[i].NotReadyAddresses = filterValidEndpointsAddr(endpoints.Subsets[i].NotReadyAddresses, nodePool)
+			if endpoints.Subsets[i].Addresses != nil || endpoints.Subsets[i].NotReadyAddresses != nil {
```
rambohe-ch, line 138: `if endpoints.Subsets[i].Addresses != nil || endpoints.Subsets[i].NotReadyAddresses != nil ` --> 
`if len(endpoints.Subsets[i].Addresses) != 0 || len(endpoints.Subsets[i].NotReadyAddresses) != 0 `

rambohe-ch: empty slice is equal nil, so no problem.

PR #698, opened 2021-12-28
[bug] fix make lint bug

Congrool:
#### What type of PR is this?
/kind bug

Fix some bugs when `make lint`:
1. `GOBIN` is defined nowhere, so when we `make lint` with no `golangci-lint` installed, the shell will run `/golangci-lint` and result in error.
2. `golangci-lint` install command bug
3.  gofmt some files to enable `make lint` to support go 1.17 or later version

Since go 1.17, it will transition from `// +build:` to `//go:build` to do conditional compilation. Thus, when `make lint` with go 1.17 or later version, it will end in gofmt error.
More Info: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md

Fixes #

```release-note

```

### DIFF
--- a/pkg/util/iptables/iptables_linux.go
+++ b/pkg/util/iptables/iptables_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 /*
--- a/Makefile
+++ b/Makefile
@@ -82,9 +82,9 @@ install-golint: ## check golint if not exist install golint tools
 ifeq (, $(shell which golangci-lint))
 	@{ \
 	set -e ;\
-	o install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.31.0 ;\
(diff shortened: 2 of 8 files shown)


pkg/util/iptables/iptables_linux.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux
```
rambohe-ch, line 1: why add this comment line?

Congrool: Since go 1.17, it will transition from // +build: to //go:build to do conditional compilation. Thus, when make lint with go 1.17 or later version, it will end in gofmt error.
More Info: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md
I think we'd better to allow developers to use go 1.17.

rambohe-ch (approved):

PR #704, opened 2022-01-03
refactor yurtctl init

### LINKED ISSUE #672: [feature request]Refactor yurtctl init to call sealer to initialize OpenYurt cluster
As mentioned in the issue #656 , after we can use sealer to initialize an OpenYurt cluster, how about refactor `yurtctl init` to call sealer(https://github.com/alibaba/sealer) to initialize the cluster?

**Why is this needed**:
1.We can remove the dependency of `k8s.io/kubernetes` in `yurtctl init` .(issue #550 )
2.Using `yurtctl init` to initialize the cluster instead of using sealer can unify the interface of cluster installation and is more user-friendly.

**others**
/kind feature
/kind good-first-issue

Peeknut:
/kind feature

Use sealer to init the cluster which is more user-friendly.
In addition, we can remove the dependency of k8s.io/kubernetes in yurtctl init.

Fixes #672

```release-note

```


Peeknut:
@rambohe-ch PTAL
The local test passed. And the cluster image needs to be uploaded to the registry.

pkg/yurtctl/cmd/yurtinit/init.go
```diff
@@ -19,620 +18,286 @@ package yurtinit
...
+	klog.Infof("generate Clusterfile for openyurt")
+	err := os.MkdirAll(TmpDownloadDir, dirMode)
 	if err != nil {
 		return err
 	}
-	joinCloudNodeCommand, err := getJoinCommand(adminKubeConfigPath, token, "cloud-node")
+
+	clusterfile, err := tmplutil.SubsituteTemplate(OpenYurtClusterfile, map[string]string{
+		"apiserver_address":         ci.AdvertiseAddress,
+		"cluster_image":             fmt.Sprintf(LocalInitClusterImage, ci.OpenYurtVersion),
```
rambohe-ch, line 274: maybe we need to use `PullInitClusterImage ` instead `LocalInitClusterImage `

pkg/yurtctl/cmd/yurtinit/options.go
```diff
@@ -0,0 +1,89 @@
...
+func (o *InitOptions) Validate() error {
+	if err := validateServerAddress(o.AdvertiseAddress); err != nil {
+		return err
+	}
+	if o.YurttunnelServerAddress != "" {
+		if err := validateServerAddress(o.YurttunnelServerAddress); err != nil {
+			return err
+		}
+	}
+	if o.Password == "" {
```
rambohe-ch, line 55: if run `yurtctl init` on master node,  i think `password` should be not needed.

rambohe-ch:
@Peeknut Maybe the readme of `yurtctl init` should be updated.

pkg/yurtctl/cmd/yurtinit/options.go
Peeknut: In the Linux system, the user login password is stored in /etc/shadow, the password is encrypted, and the encryption method may be different, such as sha512, md5, and password decryption requires tools. So i keep password here.
Is there a better way?

rambohe-ch: @Peeknut I think that `sealer` should support install `OpenYurt controlplane` without password.

Peeknut: The current version(0.6.0) of sealer does not support it. How about keeping this parameter for now? After they support, I will remove `password`.

pkg/yurtctl/cmd/yurtinit/init.go
Peeknut: down

Peeknut:
> @Peeknut Maybe the readme of `yurtctl init` should be updated.

done

pkg/yurtctl/cmd/yurtinit/options.go
rambohe-ch: ok, i got it.

rambohe-ch (approved):

PR #724, opened 2022-01-25
bugfix: panic happened when x-tunnel-server-svc service type is lb

rambohe-ch:
/kind bug

when type of service `x-tunnel-server-svc` is LoadBalancer, a panic will happen when yurt-tunnel-server component startups.

Fixes #

```release-note

```


SataQiu (commented):
/lgtm

PR #725, opened 2022-01-25
enhancement: tunnel-server supports to proxy requests that access tunnel-server directly with specified destination

rambohe-ch:
/kind enhancement

if clients on cloud side want to access edge site through yurt-tunnel and these clients can not use dns mode or iptables mode, so yurt-tunnel need to support to proxy requests that access tunnel-server directly with specified edge node in request header.

Fixes #

```release-note
yurt-tunnel will support to proxy requests that access yurt-tunnel directly with specified header. take `curl` command as an example: 
- tunnel-server address: http://10.100.2.2:10264
- edge node endpoints: http://192.168.1.2:8088/metrics
- so user can exec command to access edge node endpoints: 
     curl -H "X-Tunnel-Proxy-Dest: 192.168.1.2:8088" http://10.100.2.2:10264/metrics
```

### DIFF
--- a/pkg/util/certmanager/certmanager.go
+++ b/pkg/util/certmanager/certmanager.go
@@ -196,7 +196,7 @@ func newCertManager(
 		if getIPs != nil {
 			tmpIPs, err := getIPs()
 			if err == nil && len(tmpIPs) != 0 {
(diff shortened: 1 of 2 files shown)


Congrool (commented):
It looks good.

PR #748, opened 2022-02-24
Remove convert and revert command

### LINKED ISSUE #736: [feature request]deprecate Yurtctl convert/revert command
deprecate `Yurtctl convert/revert command`

**Why is this needed**:
As discussed in [YurtCluster Operator Propsal](https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20210722-yurtcluster-operator.md#non-goalsfuture-work), we will deprecate `yurtctl convert/revert` command and use `YurtCluster Operator` to install OpenYurt on K8s instead.

**others**
/kind feature

MIBc:
#### What type of PR is this?
/kind feature

Remove convert and revert command

Fix #736


pkg/yurtctl/kubernetes/kubeadm/app/phases/kubelet/flags_windows.go
```diff
@@ -18,8 +18,10 @@ limitations under the License.
 
 package kubelet
 
+import "github.com/openyurtio/openyurt/pkg/yurtctl/cmd/join/joindata"
+
 // buildKubeletArgMap takes a kubeletFlagsOpts object and builds based on that a string-string map with flags
 // that should be given to the local Windows kubelet daemon.
-func buildKubeletArgMap(opts kubeletFlagsOpts) map[string]string {
```
Congrool, line 23: why do we need modify this?

Congrool (commented):
The e2e-tests in Github Actions uses `yurtctl convert` to convert the k8s cluster to openyurt cluster. So i think it should be revised as well.  See file https://github.com/openyurtio/openyurt/blob/master/hack/local_up_openyurt.sh#L250-L274

pkg/yurtctl/kubernetes/kubeadm/app/phases/kubelet/flags_windows.go
MIBc: The yurtctl can not build on windows

rambohe-ch:
@Congrool as we discussed before in proposal [YurtCluster Operator Propsal](https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20210722-yurtcluster-operator.md#non-goalsfuture-work), in order to reduce user's choice difficulty, so deprecate `yurtctl convert/revert` and recommend user to use `YurtCluster Operator` to convert Kubernetes cluster to OpenYurt cluster. 

but `local_up_openyurt.sh` have used `yurtctl convert/revert`, do you have any suggestions for improving `local_up_openyurt.sh`?

rambohe-ch:
@MIBc Very appreciate for raising pull request. 

`local_up_openyurt.sh` is used by OpenYurt github actions, and `yurtctl convert/revert` is used by `local_up_openyurt.sh`, so we need to discuss how to tune `local_up_openyurt.sh` before deprecate `yurtctl convert/revert`. and @Congrool is the owner of `local_up_openyurt.sh`.

Congrool:
@rambohe-ch 

In my mind, there're two approaches to solve this problem:
1. use YurtCluster Operator to do the conversion
2. enable `yurtctl init` to setup openyurt cluster with `kind`

Personally, I don't recommand 1, because it will cause coupling between OpenYurt Repo and YurtCluster Operator Repo. Then, we may need to change the local_up_openyurt.sh when the YurtCluster Repo modifies its API. It's not convenient for us.

For 2, we can add another option for `yurtctl init`, called `--provider`. If we set it as `--provider=kind`, then an openyurt cluster will be created through `kind`. Otherwise, it will be created through `sealer` as what it does currently.

By the way, code of `yurtctl convert` can be reused when initing OpenYurt cluster with `kind`.

Of course, it will be better if sealer provides a way to init a cluster with `kind`, but I do not find any doc about it.

How do you think? @MIBc @rambohe-ch

rambohe-ch:
> @rambohe-ch
> 
> In my mind, there're two approaches to solve this problem:
> 
> 1. use YurtCluster Operator to do the conversion
> 2. enable `yurtctl init` to setup openyurt cluster with `kind`
> 
> Personally, I don't recommand 1, because it will cause coupling between OpenYurt Repo and YurtCluster Operator Repo. Then, we may need to change the local_up_openyurt.sh when the YurtCluster Repo modifies its API. It's not convenient for us.
> 
> For 2, we can add another option for `yurtctl init`, called `--provider`. If we set it as `--provider=kind`, then an openyurt cluster will be created through `kind`. Otherwise, it will be created through `sealer` as what it does currently.
> 
> By the way, code of `yurtctl convert` can be reused when initing OpenYurt cluster with `kind`.
> 
> Of course, it will be better if sealer provides a way to init a cluster with `kind`, but I do not find any doc about it.
> 
> How do you think? @MIBc @rambohe-ch

@Congrool Thanks for your feedback. i also agree with the above approach 2. so we need to refactor `yurtctl init` command before deprecate `yurtctl convert/revert`. @MIBc Do you have any comments?

Congrool:
I will open a new issue about the refactor of `yurtctl init`. After then, we can revise `local_up_openyurt.sh` using `yurtctl init`.

rambohe-ch:
> I will open a new issue about the refactor of `yurtctl init`. After then, we can revise `local_up_openyurt.sh` using `yurtctl init`.

@MIBc how about wait for new `local_up_openyurt.sh` ready? and then we can merge this pull request.

/hold

rambohe-ch:
@MIBc please fix the conflicting files.

rambohe-ch:
@MIBc please `git rebase` the newest master branch to fix e2e error.

pkg/yurtctl/kubernetes/kubeadm/app/phases/kubelet/flags_windows.go
Congrool: It seems that opts used in the following statement is defined no where. Should it be `data` instead?
```go
return buildKubeletArgMapCommon(opts)
```

rambohe-ch:
@MIBc In order to promote the development of yurtctl, #826 for deprecating `yurtctl convert/revert` had been merged. so we will close this pull request.

PR #749, opened 2022-02-26
Use configmap to configure the data source of filter framework

### LINKED ISSUE #699: [feature request] Yurthub data filtering framework optimization
Current yurthub data filtering framework has some limitations, such as:
1. It hardcoded the components and the resources/verbs, which is not convenient to reuse the same resource filtering for different components.
2. One component can only filter one resource, which can not meet the requirements for some components.
So it seems the current implementation needs to be optimized to be a more generic way:
1. Implement the different resources filtering in the core framework
2. Define the components and resources/verbs map in a configurable way, e.g. configmap
3. Implementing it through webhook may be another option

**Why is this needed**:
1. One resource filtering can be reused by multiple components
2. One component can filter multiple resources conveniently, especially for the components which needs to cache data locally by yurthub, filtering to cache the resources only it requires can reduce the storage burden of the Edge device.

**others**
/kind feature

yingjianjian:
Use configmap to configure the data source of filter framework

Fixes #699


yingjianjian:
Merged

pkg/yurthub/filter/register/register.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/discardcloudservice"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/endpointfilter"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/masterservice"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/servicetopology"
+)
```
zzguang, line 27: Here I just feel confusing that discardcloudservice/masterservice/servicetopology are classified by feature name, while endpointfilter is classified by resource name, which seems not to be consistent. How about we implement to filter all the necessary resources for nodepools together in one place, then make configuration to bind the corresponding resources filtering to the related features? Not sure whether it is a better implementation, just a reference for you, thanks!

rambohe-ch:
@yingjianjian Very appreciate for you to make pull request. I think we need to discuss the format of new field for configuring filters in `kube-system/yurthub-cfg` configmap at first. and then refactor the pull request on the base of new field.

yingjianjian:
> @yingjianjian Very appreciate for you to make pull request. I think we need to discuss the format of new field for configuring filters in `kube-system/yurthub-cfg` configmap at first. and then refactor the pull request on the base of new field.

Optimize the filter field configuration, please review！

rambohe-ch:
@yingjianjian we can reuse shared informer to get yurt-hub-cfg configmap instead of getting configmap from local storage.
and you can reference the code in the link: https://github.com/openyurtio/openyurt/blob/master/cmd/yurthub/app/config/config.go#L275-L283

config/yaml-template/yurthub-cfg.yaml
```diff
@@ -46,3 +46,7 @@ metadata:
   namespace: kube-system
 data:
   cache_agents: ""
+  filter_endpoints: coredns/endoints#list;watch
```
rambohe-ch, line 49: endoints --> endpoints

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+// Parse the special format of filter config: filter_{name}: user-agent1/resource#verb1;verb2, user-agent2/resource#verb1;verb2.
+func (a *approver) parseFilterConfig(cfg string, filterType string) []*requestInfo {
+	var reqs []*requestInfo
+	a.Lock()
+	defer a.Unlock()
+	cfg = a.mergeFilterConfig(cfg, filterType)
+	for _, configArr := range strings.Split(cfg, ",") {
+		cfg := strings.Split(configArr, "#")
+		if len(cfg) != 2 {
+			panic("Error in filter config format in yurt-hub-cfg")
```
rambohe-ch, line 141: panic("Error in filter config format in yurt-hub-cfg") --> continue

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+		}
+
+		v := strings.Split(cfg[0], "/")
+		if len(v) != 2 {
+			panic("Error in filter config format in yurt-hub-cfg")
```
rambohe-ch, line 146: panic("Error in filter config format in yurt-hub-cfg") --> continue

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+		}
+	}
+	return false
+}
+
+// Parse the special format of filter config: filter_{name}: user-agent1/resource#verb1;verb2, user-agent2/resource#verb1;verb2.
+func (a *approver) parseFilterConfig(cfg string, filterType string) []*requestInfo {
+	var reqs []*requestInfo
+	a.Lock()
+	defer a.Unlock()
```
rambohe-ch, line 136: ``
a.Lock()
defer a.Unlock()
```
is not needed, because `approver` has not been modified in this func.

File: pkg/yurthub/filter/approver.go [hunk 6]
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+	return reqs
+}
+
+// merge default filter to custom filter
+func (a *approver) mergeFilterConfig(cfg, filterType string) string {
+	if config, ok := defaultFilterCfg[filterType]; ok {
+		if cfg == "" {
+			return config
+		}
+		return fmt.Sprintf("%v,%v", config, cfg)
```
rambohe-ch, line 165:
```
		if cfg == "" {
			return config
		}
		return fmt.Sprintf("%v,%v", config, cfg)
```
can be optimized as following:

```
		if len(cfg) != 0 {
			return fmt.Sprintf("%v,%v", config, cfg)
		}
```

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+}
+
+func (a *approver) addConfigMap(obj interface{}) {
+	cfg, ok := obj.(*corev1.ConfigMap)
+	if !ok {
+		return
+	}
+	for key, value := range cfg.Data {
+		a.updateYurtHubFilterCfg(key, value, "add")
+	}
```
rambohe-ch, line 178: we can add a if condition before update as following:
```
	for key, value := range cfg.Data {
                if strings.HasPrefix(key, "filter_") {
		     a.updateYurtHubFilterCfg(key, value, "add")
                }
	}
```

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
-func (a *Approver) Approve(comp, resource, verb string) bool {
-	if a.comp != comp || a.resource != resource {
+func (a *approver) Approve(comp, resource, verb string) bool {
+	if a.isWhitelistReq(comp, resource, verb) {
 		return false
 	}
+	if ok := cache.WaitForCacheSync(a.stopCh, a.configMapSynced); !ok {
+		panic("wait for configMap cache sync timeout")
+	}
+
```
rambohe-ch, line 89: add `a.Lock()` and `defer a.Unlock()` because `a.nameToRequest` will be used.

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+	comp, ok := util.ClientComponentFrom(ctx)
+	if !ok {
+		return ""
+	}
+
+	info, ok := apirequest.RequestInfoFrom(ctx)
+	if !ok {
+		return ""
+	}
+
```
rambohe-ch, line 111: add a.Lock() and defer a.Unlock() because a.nameToRequest will be used.

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+// update filter cfg
+func (a *approver) updateYurtHubFilterCfg(filterCfgKey, filterCfgValue, action string) {
+	s := strings.Split(filterCfgKey, "_")
+	if len(s) != 2 {
+		return
+	}
+	if s[0] != "filter" {
+		return
+	}
```
rambohe-ch, line 212: add a.Lock() and defer a.Unlock() because a.nameToRequest will be modifed.

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+
+type requestInfo struct {
+	comp     string
+	resource string
+	verbs    sets.String
+}
+
+type approver struct {
+	sync.RWMutex
+	nameToRequest     map[string][]*requestInfo
```
rambohe-ch, line 44: nameToRequest --> nameToRequests

pkg/yurthub/filter/constant.go
```diff
@@ -34,9 +34,9 @@ const (
 	// End users should add annotation["openyurt.io/skip-discard"]="true" for LB service.
 	SkipDiscardServiceAnnotation = "openyurt.io/skip-discard"
 
-	// ingresscontroller filter is used to reassemble endpoints in order to make the data traffic be
+	// endpoint filter is used to reassemble endpoints in order to make the data traffic be
```
rambohe-ch, line 37: endpoint --> endpoints

pkg/yurthub/filter/constant.go
```diff
@@ -34,9 +34,9 @@ const (
...
 	// load balanced only to the nodepool valid endpoints.
-	IngressControllerFilterName = "ingresscontroller"
+	EndpointFilterName = "endpoints"
```
rambohe-ch, line 39: EndpointFilterName --> EndpointsFilterName

pkg/yurthub/filter/filter.go
```diff
@@ -49,8 +49,10 @@ func NewFilters(disabledFilters []string) *Filters {
 	}
 }
 
-func (fs *Filters) NewFromFilters(initializer FilterInitializer) (Interface, error) {
-	var filters []Interface
+func (fs *Filters) NewFromFilters(initializer FilterInitializer) (map[string]Runner, error) {
+	var filterMapping = make(map[string]Runner)
+	fs.Lock()
+	defer fs.Unlock()
```
rambohe-ch, line 55:
```
	fs.Lock()
	defer fs.Unlock()
```
is not needed, please remove it

pkg/yurthub/filter/manager.go
```diff
@@ -0,0 +1,63 @@
...
+
+	apirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/informers"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/util"
+)
+
+type Manager struct {
+	Approver
+	NameToFilter *map[string]Runner
```
rambohe-ch, line 31: NameToFilter *map[string]Runner --> NameToFilter map[string]Runner

pkg/yurthub/filter/manager.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func NewFilterManager(sharedFactory informers.SharedInformerFactory, filters map[string]Runner) *Manager {
+	return &Manager{
+		Approver:     newApprover(sharedFactory),
+		NameToFilter: &filters,
```
rambohe-ch, line 37: NameToFilter: &filters --> NameToFilter: filters

pkg/yurthub/filter/manager.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+}
```
rambohe-ch, line 39:
```
	return &Manager{
		Approver:     newApprover(sharedFactory),
		NameToFilter: &filters,
	}
```
can be optimized as following:

```
	m := &Manager{
		Approver:     newApprover(sharedFactory),
		NameToFilter: make(map[string]filter.Runner),
	}
        
        for name, runner := range filters {
               m.NameToFilter[name] = runner
        }

        return m
```

pkg/yurthub/filter/manager.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (m *Manager) Filter(req *http.Request, rc io.ReadCloser, stopCh <-chan struct{}) (int, io.ReadCloser, error) {
+	reqName := m.Approver.GetFilterName(req)
+	for name, ins := range *m.NameToFilter {
+		if reqName == name {
+			return ins.Filter(req, rc, stopCh)
+		}
+	}
```
rambohe-ch, line 47:
```
	for name, ins := range *m.NameToFilter {
		if reqName == name {
			return ins.Filter(req, rc, stopCh)
		}
	}
```
can be optimized as following:

```
if ok, runner := m.NameToFilter[reqName]; ok {
         return runner.Filter(req, rc, stopCh)
}
```

yingjianjian:
fixed

pkg/yurthub/filter/approver.go
```diff
@@ -16,26 +16,209 @@ limitations under the License.
...
+	newCfg, ok := newObj.(*corev1.ConfigMap)
+	if !ok {
+		return
+	}
+
+	for key, value := range newCfg.Data {
+		if _, ok := oldCfg.Data[key]; !ok {
+			if strings.HasPrefix(key, "filter_") {
+				a.updateYurtHubFilterCfg(key, value, "update")
+				return
```
rambohe-ch, line 196: return --> continue

rambohe-ch:
@zzguang PTAL

rambohe-ch:
> Generally, it looks good to me, but I have a suggestion I am not sure whether it makes sense: defaultFilterCfg = map[string]string{ MasterServiceFilterName: "kubelet/services#list;watch", EndpointsFilterName: "nginx-ingress-controller/endpoints#list;watch", DiscardCloudServiceFilterName: "kube-proxy/services#list;watch", ServiceTopologyFilterName: "kube-proxy/endpointslices#list;watch", } For the current FilterNames, MasterServiceFilterName and DiscardCloudServiceFilterName are both services filtering for different components, how about combining them together into a single "ServicesFilterName", then in the filter manager, it checks all the components that registers the service resource to decide whether to do the services filtering operation. This aims to de-couple the resources and components completely, so that if a new component wants to add filtering for a current resource in future, it won't need to add new filter name any more. If the suggestion is reasonable, the filter names will be named clearly by the resources names, and the current filter names can be refined like below: ServicesFilterName: "kubelet/services#list;watch", ServicesFilterName: "kube-proxy/services#list;watch", EndpointsFilterName: "nginx-ingress-controller/endpoints#list;watch", EndpointslicesFilterName: "kube-proxy/endpointslices#list;watch",
> 
> This suggestion is just a reference for you, just ignore it if I am wrong, thanks!

@zzguang It's a good idea to keep `MasterServiceFilterName` and `DiscardCloudServiceFilterName` unchanged, because only from the name meaning of filters end users can decide to use this filter or not. if we combine these two meaningful filters into `service filter`, i think that end user will feel confused.

zzguang:
> 

Okay, I just feel a little confusing about the filter naming rule, if we name it by the feature name, then the "EndpointsFilterName" doesn't follow this rule, and if another feature/component wants to add endpoints filtering in future, it may lead to some un-consistent concepts.
As you say, I think we can merge it at current stage, and optimize it in another PR in future.

rambohe-ch:
Fixes #773

PR #751, opened 2022-03-01
support config k8s resource server

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurtctl/cmd/join/join.go
+++ b/pkg/yurtctl/cmd/join/join.go
@@ -66,6 +66,7 @@ type joinOptions struct {
 	unsafeSkipCAVerification bool
 	ignorePreflightErrors    []string
 	nodeLabels               string
+	kubernetesResourceServer string
 }
 
 // newJoinOptions returns a struct ready for being used for creating cmd join flags.
@@ -78,6 +79,7 @@ func newJoinOptions() *joinOptions {
 		caCertHashes:             make([]string, 0),
 		unsafeSkipCAVerification: false,
 		ignorePreflightErrors:    make([]string, 0),
+		kubernetesResourceServer: yurtconstants.DefaultKubernetesResourceServer,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Congrool:
@huiwq1990 Thanks for your contribution. And could you please add some details about why we need this option?

huiwq1990:
> @huiwq1990 Thanks for your contribution. And could you please add some details about why we need this option?

In private delivery scenes, network maybe not allow access public domain , it's better allow custom define resource address.

rambohe-ch:
> > @huiwq1990 Thanks for your contribution. And could you please add some details about why we need this option?
> 
> In private delivery scenes, network maybe not allow access public domain , it's better allow custom define resource address.

@huiwq1990 Thanks for your feedback. it looks reasonable to add this parameter, @Congrool how about your advices?

pkg/yurtctl/cmd/join/join.go
```diff
@@ -159,21 +161,26 @@ func addJoinConfigFlags(flagSet *flag.FlagSet, joinOptions *joinOptions) {
 		&joinOptions.nodeLabels, options.NodeLabels, joinOptions.nodeLabels,
 		"Sets the labels for joining node",
 	)
+	flagSet.StringVar(
+		&joinOptions.kubernetesResourceServer, options.KubernetesResourceServer, joinOptions.kubernetesResourceServer,
+		"Sets the address for download k8s node resources",
```
Congrool, line 166: Maybe we should modify it as `Sets the address for downloading k8s node resources`

Congrool (commented):
It looks good :-), please correct the garmar.

pkg/yurtctl/cmd/join/join.go
huiwq1990: Just correct it.

rambohe-ch:
@huiwq1990 Thanks for fixing comments, please merge two commits into one commit.

huiwq1990:
> @huiwq1990 Thanks for fixing comments, please merge two commits into one commit.

Already done.

Congrool (commented):
/lgtm

rambohe-ch (approved):

Issue #522, opened 2021-10-14
[Question] How to proxy requests from cloud to pod (not hostNetwork) on edge?


tiezhuoyu:
**What happened**:
Hi, i want to know if it is possible to proxy requests from cloud to pods (not hostNetwork mod) on edge? I found issue [issue#138](https://github.com/openyurtio/openyurt/issues/138), but it works on hostNetwork mod. 

Thanks!

**What you expected to happen**:
For example, pod1(192.168.0.5) on cloud node, and pod2(192.168.4.5) on edge node.
If it is possible pod1 send reqeust like 'GET http://192.168.4.5:80' ?

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- OpenYurt version: v0.5.0
- Kubernetes version (use `kubectl version`): v1.18.2
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools: yurtctl init & yurtctl join
- Others:

**others**
/kind question

rambohe-ch:
@tiezhuoyu Thank you for raising this issue.
Proxy request from cloud to edge pods(not hostnetwork) is not supported at present as @DrmagicE has replied. and Would you be able to describe the detail scenario about this requirement?

DrmagicE:
@rambohe-ch Yes, I'll try it.

lujinda:
We are designing a solution that connects the edge and the edge, and the edge and the cloud network. The general idea is to use IPSec (or other) to connect to each data center, and then use IPSec to send the traffic of the Pod in one data center to other data centers. Plan to integrate with openyurt. The plan hopes that the intrusion to the Pod is small enough.

rambohe-ch:
@tiezhuoyu Pods(not hostNetwork) connection across public network will be solved by [raven](https://github.com/openyurtio/raven),  also you can dive into the proposal: https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20211123-enhancement-of-cluster-networking.md

and we will close this issue, and you reopen or create a new issue if you have any other questions.

tiezhuoyu:
@rambohe-ch that's a greate work 👍

PR #763, opened 2022-03-02
Agent tolerate the edge taint

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

[yurtcluster-operator](https://github.com/openyurtio/yurtcluster-operator/blob/19ba3a03f319b01c21ca63ccf1a556a4536bdd1b/pkg/util/kubernetes.go#L355) auto add taints, we need tolerate it.

Fixes problem as issue <https://github.com/openyurtio/openyurt.io/pull/83>

```release-note

```

### DIFF
--- a/config/setup/yurt-tunnel-agent.yaml
+++ b/config/setup/yurt-tunnel-agent.yaml
@@ -51,6 +51,9 @@ spec:
       restartPolicy: Always
       tolerations:
         - operator: Exists
+        - operator: Equal
+          key: node-role.openyurt.io/edge
+          effect: NoSchedule
       volumes:
       - name: k8s-dir
         hostPath:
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@huiwq1990 Very appreciate for raising pull request. and i think the taints that YurtCluster added for edge nodes should be discussed. because the taints are not added in other ways. @SataQiu

rambohe-ch:
@huiwq1990 as we discussed in issue #522 , YurtCluster will be deprecated for OpenYurt installation. so the problem that solved by this pull request will disappear.  maybe we can close this pull request manually.

PR #787, opened 2022-03-29
finish ut of sa token replacement

MrGirl:
add tenant sa token subsitute

> /kind enhancement

> /sig iot

Subsitute bearer token with tenant's default serviceaccount token when requests are from namespace kube-system. This PR make the master understand which tenant the request is from, when the request is from kube-system and flannel 

If we want to active tenant sa toke subsitution, we should add param "hub-cert-organizations"

```bash
--hub-cert-organizations=system:bootstrappers:kubeadm:default-node-token,openyurt:tenant:iot-test
```
```release-note

```

#### other Note

Maybe we should update clusterrolebinding first.
```bash
kubectl create clusterrolebinding system-node-role-bound --clusterrole=system:node --group=system:nodes  
```


rambohe-ch:
@MrGirl Very appreciate for your pull request. please fix the errors of github actions and merge two commits into one commit.

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,184 @@
...
+
+		klog.Errorf("failed to create client set, %v", err)
+
+	}
+
+	tenantMgr.Storage = cfg.StorageWrapper
+
+	if clientSet != nil {
+
+		factory := informers.NewSharedInformerFactoryWithOptions(clientSet, time.Minute, informers.WithNamespace(tenantNs))
```
rambohe-ch, line 68: please reuse `sharedInformer` in `config.YurtHubConfiguration` instead of creating a new `sharedInformer`.

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,184 @@
...
+
+	kubeConf := configMgr.GetRestConfig(true)
+	clientSet, err := kubernetes.NewForConfig(kubeConf)
+	if err != nil {
+
+		klog.Errorf("failed to create client set, %v", err)
+
+	}
+
+	tenantMgr.Storage = cfg.StorageWrapper
```
rambohe-ch, line 64: It's not need to store `secret` in tenant manager. the secret will be stored by `remoteProxy` automatically.

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +297,22 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
...
 	})
 }
+
+func WithSaTokenSubsitute(handler http.Handler, token string) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+
+		if info, ok := apirequest.RequestInfoFrom(req.Context()); ok {
+
+			if info.Namespace == "kube-system" && (info.Verb == "get" || info.Verb == "list" || info.Verb == "watch") {
```
rambohe-ch, line 307: the condition of replacing bearer token is the bearer token in request header is a bearer token of  kube-system namespace. not the request namespace is kube-system.

pkg/yurthub/proxy/proxy.go
```diff
@@ -105,7 +108,11 @@ func (p *yurtReverseProxy) buildHandlerChain(handler http.Handler) http.Handler
 	}
 	handler = util.WithMaxInFlightLimit(handler, p.maxRequestsInFlight)
 	handler = util.WithRequestClientComponent(handler)
+
+	handler = util.WithSaTokenSubsitute(handler, p.tenantToken)
```
rambohe-ch, line 112: yurthub should not enable `WithSaTokenSubsitute` handler in a default mode, maybe we can add a startup parameter for yurthub that used to decide to enable this feature or not.

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +297,22 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
...
+
+				req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", token))
```
rambohe-ch, line 309: the token maybe empty, so we need to wait for the token is ready before replace the bearer token in request header.

cmd/yurthub/app/config/config.go
```diff
@@ -88,6 +88,7 @@ type YurtHubConfiguration struct {
 	WorkingMode                       util.WorkingMode
 	KubeletHealthGracePeriod          time.Duration
 	FilterManager                     *filter.Manager
+	DiskCachePath                     string
```
rambohe-ch, line 91: DiskCachePath is not need to add

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +300,38 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
...
+func WithSaTokenSubsitute(handler http.Handler, tenantNs string) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+
+		if common.TenantToken == "" {
+			return
+		}
+		if info, ok := apirequest.RequestInfoFrom(req.Context()); ok {
+
+			if info.Verb == "get" || info.Verb == "list" || info.Verb == "watch" {
```
rambohe-ch, line 313: all kind of request should be handled

pkg/yurthub/storage/factory/factory.go
```diff
@@ -17,11 +17,25 @@ limitations under the License.
...
 // CreateStorage create a storage.Store for backend storage
+// Singleton mode
 func CreateStorage(cachePath string) (storage.Store, error) {
-	return disk.NewDiskStorage(cachePath)
+
+	once.Do(func() {
+		diskStorage, err = disk.NewDiskStorage(cachePath)
+	})
+
+	return diskStorage, err
```
rambohe-ch, line 40: undo

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,171 @@
...
+	tenantNs := ParseTenantNs(cfg.YurtHubCertOrganizations)
+	klog.Infof("parse tenant ns: %s", tenantNs)
+	if tenantNs == "" {
+		return nil
+	}
+
+	tenantMgr := &tenantManager{TenantNs: tenantNs}
+
+	kubeConf := configMgr.GetRestConfig(true)
+	clientSet, err := kubernetes.NewForConfig(kubeConf)
```
rambohe-ch, line 52: remove clientSet

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+
+		klog.Errorf("failed to create client set, %v", err)
+
+	}
+
+	tenantMgr.Storage = cacheMgr
```
rambohe-ch, line 59: remove Storage

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,171 @@
...
+		//add handlers
+		tenantMgr.Informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+			AddFunc:    tenantMgr.addSecret,
+			UpdateFunc: tenantMgr.updateSecret,
+			DeleteFunc: tenantMgr.deleteSecret})
+
+		//get token
+		//req, _ := http.NewRequest(http.MethodGet, "/api/v1/namespaces/iot-test/secrets?limit=500", nil)
+		//secretList, err := cacheMgr.QueryCache(req)
+		secretList := tenantMgr.Informer.GetStore().List() //Duplicated list. If before reflector, get from apiserver, or from local cache
```
rambohe-ch, line 74: It's not a good idea to list at

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +300,38 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
 		handler.ServeHTTP(w, req)
 	})
 }
+
+func WithSaTokenSubsitute(handler http.Handler, tenantNs string) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+
+		if common.TenantToken == "" {
```
rambohe-ch, line 308: please use `hasSynced` of sharedInformer.

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	tokenStr := string(mgr.TenantSecret.Data["token"])
+
+	return tokenStr
+}
+
+func (mgr *tenantManager) addSecret(sec interface{}) {
+
+	secret := sec.(v1.Secret)
```
rambohe-ch, line 133: secret, ok := sec.(*v1.Secret)
if !ok {
      // log warning
      return
}

pkg/yurthub/tenant/common/token.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package common
+
+var TenantToken string
```
rambohe-ch, line 18: remove global var

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,148 @@
...
+	tenantNs := util.ParseTenantNsFromOrgs(orgs)
+	klog.Infof("parse tenant ns: %s", tenantNs)
+	if tenantNs == "" {
+		return nil
+	}
+
+	tenantMgr := &tenantManager{TenantNs: tenantNs}
+
+	if factory != nil {
+		tenantMgr.Informer = factory.InformerFor(&v1.Secret{}, nil) //get registered secret informer
```
rambohe-ch, line 58: factory.InformerFor(&v1.Secret{}, nil) --> sharedFactory.Core().V1().Secrets().Informer()

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,148 @@
...
+
+	GetTenantToken() string
+
+	InformerSynced() cache.InformerSynced
+
+	SetSecret(sec *v1.Secret)
+}
+
+type tenantManager struct {
+	Informer cache.SharedIndexInformer
```
rambohe-ch, line 38: Informer field is unused

pkg/yurthub/util/util.go
```diff
@@ -473,3 +473,47 @@ func NewGZipReaderCloser(header http.Header, body io.ReadCloser, req *http.Reque
...
 	}, true
 }
+
+func ParseTenantNs(certOrg string) string {
+
+	if !strings.Contains(certOrg, "openyurt:tenant:") {
+		return ""
+	}
+
+	idx := strings.LastIndex(certOrg, "openyurt:tenant:") + len("openyurt:tenant:")
```
rambohe-ch, line 483: return strings.TrimPrefix(certOrg, "openyurt:tenant:")

MrGirl: `
strings.TrimPrefix("a", "openyurt:tenant:") will return "a" as tenant ns
`

pkg/yurthub/tenant/tenant.go
```diff
@@ -0,0 +1,148 @@
...
+func (mgr *tenantManager) GetTenantToken() string {
+
+	if mgr.TenantSecret == nil {
+
+		return ""
+	}
+
+	tokenStr := string(mgr.TenantSecret.Data["token"])
+
+	return tokenStr
```
rambohe-ch, line 81: return mgr.TenantSecret.Data["token"]

pkg/yurthub/storage/factory/factory.go
```diff
@@ -23,5 +23,7 @@ import (
 
 // CreateStorage create a storage.Store for backend storage
 func CreateStorage(cachePath string) (storage.Store, error) {
+
 	return disk.NewDiskStorage(cachePath)
+
```
rambohe-ch, line 28: please undo the change.

pkg/yurthub/tenant/tenant.go
MrGirl: you mean change  GetTenantToken() to return []byte

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +300,31 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
...
+
+		if tenantMgr.GetTenantToken() == "" {
+			klog.V(2).Info("tenant token is empty.")
+		} else if oldToken := util.ParseBearerToken(req.Header.Get("Authorization")); oldToken != "" { //bearer token is not empty&valid
+
+			if jsonWebToken, err := jwt.ParseSigned(oldToken); err != nil {
+
+				klog.Errorf("invaled bearer token %s, err: %v", oldToken, err)
+			} else {
+				oldClaim := jwt2.BearerClaims{}
```
rambohe-ch, line 316: in order to get `namespace` from token, we can only use subject in `jwt.Claims` instead of defining `BearerClaims`. we can reference code of the subject format here: 
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/serviceaccount/tokens_controller.go#L404
https://github.com/kubernetes/kubernetes/blob/master/pkg/serviceaccount/legacy.go#L32-L41

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +300,31 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
 		handler.ServeHTTP(w, req)
 	})
 }
+
+func WithSaTokenSubstitute(handler http.Handler, tenantMgr tenant.Interface) http.Handler {
```
rambohe-ch, line 304: i think that the logic of this handler should be as following:
```
if request has bearer token {
     // parse bearer token to resolve namespace

     // verify namespace in bearer token is tenant namespace or not.
     // if equal to tenant namespace, do not replace bearer token, goto handler.ServeHTTP(w, req)

     // wait for tenant token is synced

    // after tenant token synced, replace bearer token with tenant token in request header.
}
```

pkg/yurthub/proxy/util/util.go
MrGirl:
```
type legacyPrivateClaims struct {
	ServiceAccountName string `json:"kubernetes.io/serviceaccount/service-account.name"`
	ServiceAccountUID  string `json:"kubernetes.io/serviceaccount/service-account.uid"`
	SecretName         string `json:"kubernetes.io/serviceaccount/secret.name"`
	Namespace          string `json:"kubernetes.io/serviceaccount/namespace"`
}
```

legacyPrivateClaims  is private, so I copy the type to a new type.

pkg/yurthub/tenant/tenant.go
rambohe-ch: ok, but i think it's more readable to use `sharedFactory.Core().V1().Secrets().Informer()` to get informer.

pkg/yurthub/tenant/tenant.go
rambohe-ch: @MrGirl I think that it's not need to define local var `tokenStr`, just return `mgr.TenantSecret.Data["token"]` is ok.

pkg/yurthub/proxy/util/util.go
rambohe-ch: @MrGirl we only need namespace, so `jwt.Claims` can be used. we need try our best to reduce the code complexity.

pkg/yurthub/proxy/util/util.go
MrGirl: I think there's no need to invoke  WaitForCacheSync func in handler, the reason is as follows:
1. waitforcachesync is just used to identify that all items in fifo have been pop to callback funcs,  it check no need to do it in every request handling.
2. yurthub will get the tenant token at its first start, if the net connection is down, informer will  start its list-watch from  
cacheManager's cache, it can still get the tennat token, and I have test disconnection scenario

If WaitForCacheSync must been invoked, the time should be at yurthub initialization, the informerFactory do the waitforcachesync for one time.



waitforcachesync will call 
`// HasSynced returns true if an Add/Update/Delete/AddIfNotPresent are called first,
// or the first batch of items inserted by Replace() has been popped.
func (f *DeltaFIFO) HasSynced() bool {
	f.lock.Lock()
	defer f.lock.Unlock()
	return f.populated && f.initialPopulationCount == 0
}`

pkg/yurthub/proxy/util/util.go
MrGirl: there's no namespace field in jwt.Claims.  I this this is the reason why kubernetes had to define the type legacyPrivateClaims with a Namespace filed.

rambohe-ch: we can get namespace from `jwt.Claims.Subject` field. the contents of `Subject` is `apiserverserviceaccount.MakeUsername(serviceAccount.Namespace, serviceAccount.Name)`.

pkg/yurthub/tenant/tenant.go
rambohe-ch: ok, so `return string(mgr.TenantSecret.Data["token"])`, remove local var `tokenStr`.

pkg/yurthub/util/util.go
rambohe-ch: @MrGirl we need to simplify our design, customized org should be start with `openyurt:tenant:` prefix.

pkg/yurthub/proxy/util/util.go
rambohe-ch: @MrGirl if `WaitForCacheSync` is not used, we can not sure that tenant token is ready. requests of flanneld/kube-proxy/coredns maybe come into handler before edge-hub's requests.

pkg/yurthub/util/util.go
MrGirl: nodeclient: kubelet ask crt with system:bootstrappers 
selfnodeclient: kubelet client update crt with system:nodes 

I think it is a convention.

```
A kubelet authenticating using bootstrap tokens is authenticated as a user in the group system:bootstrappers, which is the standard method to use.
```
from [A kubelet authenticating using bootstrap tokens is authenticated as a user in the group system:bootstrappers, which is the standard method to use.](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet-tls-bootstrapping/#initial-bootstrap-authentication)

rambohe-ch: @MrGirl certOrg is used for tenant and comes from a parameter of yurthub, and the format is defined by `openyurt:tenant:xxx` in yurt-controller-manager. none business of bootstrap by kubelet.

pkg/yurthub/util/util.go
```diff
@@ -473,3 +473,46 @@ func NewGZipReaderCloser(header http.Header, body io.ReadCloser, req *http.Reque
 		body: body,
 	}, true
 }
+
+func ParseTenantNs(certOrg string) string {
+
+	if !strings.Contains(certOrg, "openyurt:tenant:") {
```
rambohe-ch, line 479: strings.Contains() --> strings.HasPrefix()

pkg/yurthub/proxy/util/util.go
```diff
@@ -295,3 +300,36 @@ func WithRequestTimeout(handler http.Handler) http.Handler {
...
+			} else {
+				oldClaim := jwt.Claims{}
+
+				if err := jsonWebToken.UnsafeClaimsWithoutVerification(&oldClaim); err == nil {
+
+					if tenantNs, _, err := serviceaccount.SplitUsername(oldClaim.Subject); err == nil {
+
+						if tenantMgr.GetTenantNs() != tenantNs && tenantNs == "kube-system" && tenantMgr.WaitForCacheSync() { // token is not from tenant's namespace
+							req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", tenantMgr.GetTenantToken()))
+							klog.V(2).Infof("replace token, old: %s, new: %s", oldToken, tenantMgr.GetTenantToken())
```
rambohe-ch, line 322: how about adjust the log level from  2 to 5 of printing token? we should not dump token in the logs.

rambohe-ch (approved):

rambohe-ch:
Fixes #788

rambohe-ch:
Fixes #788

PR #790, opened 2022-04-06
[enhancement]optimize approver of data filtering framework in yurthub

rambohe-ch:
/kind enhancement

1. response should be filtered or not  is decided in `approve() func` by iterating `nameToRequests`, and the efficiency is `N`(the length of nameToRequests slice), in this pull request, optimize the efficiency to `Log(N)` by using a `reqKeyToName` map to decide need filter or not.
2. end user can configure the `yurt-hub-cfg` configmap more freely. for example: no filter setting in `yurt-hub-cfg` configmap will not effect the default filter setting.

Fixes #

```release-note

```


rambohe-ch:
@yingjianjian PTAL

.github/workflows/ci.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Lint golang code
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
```
Congrool, line 45: It seems that #792 also works on it. How about letting #792 solve it, and making this pr focus on its own work?

rambohe-ch: ok, i will rebase this pull request after #792 merged.

pkg/yurthub/filter/approver.go
```diff
@@ -33,194 +33,247 @@ import (
...
-	for _, req := range a.whiteListRequests {
-		if req.Equal(comp, resource, verb) {
-			return true
+// merge is used to add specified setting into reqKeyToName map.
+func (a *approver) merge(action string, keyToNameSetting map[string]string) {
+	a.Lock()
+	defer a.Unlock()
+	// remove current user setting from reqKeyToName and left default setting only
+	for key := range a.reqKeyToName {
+		if _, ok := defaultReqKeyToName[key]; !ok {
```
huiwq1990, line 179: This code may be som confuse.  Delete not default keys mean only use default keys.
```go
	newReqKey := map[string]string{}
	for key, name := range defaultReqKeyToName {
		newReqKey[key] = name
	}
	for key, name := range keyToNameSetting {
		// if filter setting is duplicated, only recognize the first setting.
		if currentName, ok := newReqKey[key]; !ok {
			newReqKey[key] = name
		} else {
			klog.Warningf("request %s has already used filter %s, so skip filter %s setting", key, currentName, name)
		}
	}
	a.Lock()
	defer a.Unlock()
	a.reqKeyToName = newReqKey
```

huiwq1990, line 179: This code may be som confuse.  Delete not default keys mean only use default keys.
```go
	newReqKey := map[string]string{}
	for key, name := range defaultReqKeyToName {
		newReqKey[key] = name
	}
	for key, name := range keyToNameSetting {
		// if filter setting is duplicated, only recognize the first setting.
		if currentName, ok := newReqKey[key]; !ok {
			newReqKey[key] = name
		} else {
			klog.Warningf("request %s has already used filter %s, so skip filter %s setting", key, currentName, name)
		}
	}
	a.Lock()
	defer a.Unlock()
	a.reqKeyToName = newReqKey
```

rambohe-ch: @huiwq1990  In order to increase efficiency of `golang runtime GC`, so i don't want to create a new map every time when `yurt-hub-cfg` filter setting changed. and the new reqKeyToName is based defaultReqKeyToName plus current filter setting in `yurt-hub-cfg` configmap.

PR #791, opened 2022-04-06
support configmap delete

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/cachemanager/cache_agent.go
+++ b/pkg/yurthub/cachemanager/cache_agent.go
@@ -39,6 +39,7 @@ func (cm *cacheManager) initCacheAgents() error {
 	configmapInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    cm.addConfigmap,
 		UpdateFunc: cm.updateConfigmap,
+		DeleteFunc: cm.deleteConfigmap,
 	})
 
 	klog.Infof("init cache agents to %v", cm.cacheAgents)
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@huiwq1990 Very appreciate for making pull request. end user should not delete `yurt-hub-cfg` configmap because yurthub component on edge nodes may not work.  if `yurt-hub-cfg` configmap is deleted accidentally, the feature of yurthub can not be guaranteed. so `configmap delete event` is not handled.

rambohe-ch:
> @huiwq1990 Very appreciate for making pull request. end user should not delete `yurt-hub-cfg` configmap because yurthub component on edge nodes may not work. if `yurt-hub-cfg` configmap is deleted accidentally, the feature of yurthub can not be guaranteed. so `configmap delete event` is not handled.

@huiwq1990 I'd like to add a new comment to reply. I think that maybe handle `configmap delete event` maybe is reasonable, because end user will find feature can not work after delete `yurt-hub-cfg` configmap. In the current process, even if `yurt-hub-cfg` configmap is deleted, the original setting feature still can work, it's not reasonable.

rambohe-ch:
@huiwq1990 how about add `configmap delete event` handler for `approver`, you can dive into the code here: 
https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/filter/approver.go#L73-L76

rambohe-ch:
@huiwq1990 welcome you to apply OpenYurt community member. and the apply example is here: https://github.com/openyurtio/community/issues/33

rambohe-ch (approved):

PR #792, opened 2022-04-06
upgrade the golint ci step 

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Process [golint error](https://github.com/golangci/golangci-lint-action/issues/442).

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,7 +42,7 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Lint golang code
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
           version: ${{ env.GOLANGCI_VERSION }}
           args: -v


.github/workflows/ci.yaml
```diff
@@ -42,10 +42,9 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
       - name: Lint golang code
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v2.5.2
```
Congrool, line 45: Could you please set it as
>golangci/golangci-lint-action@v3

and find if it works?

.github/workflows/ci.yaml
```diff
@@ -42,10 +42,9 @@ jobs:
...
         with:
-          version: ${{ env.GOLANGCI_VERSION }}
-          args: -v
```
Congrool, line 48: How about keep the `-v` flag. It will be passed to golangci-lint cmd and make it output more info about errors if any. 
https://github.com/golangci/golangci-lint-action/blob/v3.1.0/action.yml#L8-L11

Congrool:
It seems that currently the error is not caused by golang 1.18 any more. There're some files cannot pass the lint check. Maybe we can revise these files according to the error output.

Congrool:
Good Job.
/lgtm

huiwq1990:
> It seems that currently the error is not caused by golang 1.18 any more. There're some files cannot pass the lint check. Maybe we can revise these files according to the error output.

Yes, all step are success.

rambohe-ch (approved):

PR #794, opened 2022-04-08
[bugfix] ungzip response by yurthub when response header contains content-encoding=gzip

rambohe-ch:
/kind bug

1. From kubernetes v1.16, kube-apiserver will compress(use gzip) response body when body data size is more than 128KB, the code is here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/writers.go#L217-L224

2. filter error or cache error will happen in yurthub component because yurthub can not uncompress response body correctly.

3. so we add a feature to enable gzip read in terms of content-encoding header in response.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/proxy/remote/remote.go
+++ b/pkg/yurthub/proxy/remote/remote.go
@@ -159,7 +159,8 @@ func (rp *RemoteProxy) modifyResponse(resp *http.Response) error {
 		// filter response data
 		if rp.filterManager != nil {
 			if rp.filterManager.Approve(req) {
-				size, filterRc, err := rp.filterManager.Filter(req, resp.Body, rp.stopCh)
+				wrapBody, needUncompressed := util.NewGZipReaderCloser(resp.Header, resp.Body, req, "filter")
+				size, filterRc, err := rp.filterManager.Filter(req, wrapBody, rp.stopCh)
 				if err != nil {
 					klog.Errorf("failed to filter response for %s, %v", util.ReqString(req), err)
 					return err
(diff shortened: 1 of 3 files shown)


pkg/yurthub/util/util.go
```diff
@@ -433,3 +434,42 @@ func CreateKubeConfigFile(kubeClientConfig *rest.Config, kubeconfigPath string)
...
+			grc.zerr = err
+			return 0, err
+		}
+	}
+
+	return grc.zr.Read(b)
+}
+
+func (grc *gzipReaderCloser) Close() error {
+	return grc.body.Close()
```
Congrool, line 463: I think we'd better call `grc.zr.Close` before calling `grc.body.Close`

rambohe-ch: `grc.zr.Close` is used to verify the reader is fully consumed or not, we need to close the underlying `io.Reader` here, so i think we only need to call `grc.body.Close()` here.

BTW: if we add `grc.zr.Close`, i don't how to handle the return error of `grc.zr.Close` now.

Congrool: OK, i'm not familiar with the gzip package. lgtm for this case.

Just in terms of how to handle the error, we can do it as follows
```go
func (grc *gzipReaderCloser) Close() error {
	if err := grc.zr.Close(); err != nil {
		klog.Errorf("info here")
	}
	return grc.body.Close()
```

DrmagicE: The `Close` method is used to verify the GZIP checksum:
```
// Close closes the Reader. It does not close the underlying io.Reader.
// In order for the GZIP checksum to be verified, the reader must be
// fully consumed until the io.EOF.
func (z *Reader) Close() error { return z.decompressor.Close() }
```
But I believe we don't need to call `grc.zr.Close` anymore, because this method now actually do nothing:
https://github.com/golang/go/blob/3e387528e54971d6009fe8833dcab6fc08737e04/src/compress/flate/inflate.go#L354-L359
```
func (f *decompressor) Close() error {
    if f.err == io.EOF {
        return nil
    }
    return f.err
}
```
 If the checksum is invalid, the `Read` method will set `f.err` and return it. I think we don't need a redundant check in `gzipReaderCloser.Close`.  

So I think `grc.body.Close()` is ok.

Reference: 
https://stackoverflow.com/questions/68074107/what-is-the-point-of-gzip-reader-close

pkg/yurthub/proxy/remote/remote.go
```diff
@@ -169,18 +170,25 @@ func (rp *RemoteProxy) modifyResponse(resp *http.Response) error {
 					resp.ContentLength = int64(size)
 					resp.Header.Set("Content-Length", fmt.Sprint(size))
 				}
+
+				// after un-gzip in filter, the header content encoding should be removed.
```
DrmagicE, line 174: I think "gunzip" is more accurate than "un-gzip".

DrmagicE:
It would be great if we add some test cases for `modifyResponse`.  (not a must for this PR)

pkg/yurthub/util/util.go
Congrool: Thanks, it's reasonable.

pkg/yurthub/proxy/remote/remote.go
rambohe-ch: done

rambohe-ch:
> It would be great if we add some test cases for `modifyResponse`. (not a must for this PR)

@DrmagicE It's a good idea to add test case for `modifyResponse`, i will add a issue about it.

PR #805, opened 2022-04-15
【Optimize】unified image pull policy

JameKeal:
unified image pull policy, It all becomes the same as `imagePullPolicy: IfNotPresent`

Fixes #

```release-note

```


pkg/yurtctl/constants/yurt-app-manager-tmpl.go
```diff
@@ -668,7 +668,7 @@ spec:
         command:
         - /usr/local/bin/yurt-app-manager
         image: {{.image}}
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
rambohe-ch, line 671: yurt-app-manager is recommended to deploy on cloud nodes, so it's better to set imagePullPolicy as `Always` instead of `IfNotPresent`

JameKeal: I know you means, but yurt-tunnel-server is also deploy on cloud nodes, it's set imagePullPolicy as IfNotPresent, I just think we need to unify, but that's just a suggestion

rambohe-ch: @JameKeal ok. please `git rebase` the newest master branch to fix e2e test error.

JameKeal: OK，finished.

rambohe-ch (approved):

PR #810, opened 2022-04-19
[enhancement]: filter http: TLS handshake error for https server at 10263 port

rambohe-ch:
/kind enhancement

skip to print log of tls handshake error for yurt-tunnel-server https server at 10263.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurttunnel/server/anpserver.go
+++ b/pkg/yurttunnel/server/anpserver.go
@@ -20,8 +20,10 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"log"
 	"net"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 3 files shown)


pkg/yurttunnel/server/anpserver.go
```diff
@@ -146,6 +148,7 @@ func runMasterServer(handler http.Handler,
 			ReadTimeout:  constants.YurttunnelANPInterceptorReadTimeoutSec * time.Second,
 			TLSConfig:    tlsCfg,
 			TLSNextProto: make(map[string]func(*http.Server, *tls.Conn, http.Handler)),
+			ErrorLog:     log.New(NewHandshakeFilterWriter(os.Stderr), "", log.LstdFlags),
```
Congrool, line 151: Why do we need to filter out the handshake error log? 
It will make debug a bit more difficult, won't it?

rambohe-ch: @Congrool some healthy checkers use tcp protocol to  check tunnel-server 10263 port status but do not start tls handshake, so a large amount `tls handshake error` logs are dumped in tunnel-server, this logs will confused end user to debug `tunnel-server` problems. at the same time, if tunnel-agent fails to connect `tunnel-server` for `tls handshake error`, tunnel-agent side also will get enough log message.

Congrool: OK, but I am still confused about this scenario. Is it possible to solve this problem through opening a new port for health check instead of using 10263?
By the way, in my mind, 10263 is the port to handle dnat requests from apiserver. The `tunnel-agent` uses port 10262 to connect to the `tunnel-server`. https://github.com/openyurtio/openyurt/blob/master/cmd/yurt-tunnel-server/app/options/options.go#L69

rambohe-ch:
I think we should remove `10263` port setting from `x-tunnel-server-svc` service that only used by yurt-tunnel-agents. and on clouds nodes, we should use `x-tunnel-server-internal-svc` service to access yurt-tunnel-server.

so the feature in this pull request is not suitable for yurt-tunnel-server, and i will close this pull reqeust.

PR #819, opened 2022-05-04
Rename yurtctl init/join/reset to yurtadm init/join/reset

### LINKED ISSUE #737: [feature request]rename yurtctl init/join to yurtadm init/join
rename yurtctl init/join to yurtadm init/join

**Why is this needed**:
`yurtctl init/join` are not suitable as name for creating OpenYurt cluster and join nodes.
so rename to `yurtadm init/join` for conforming to `kubeadm init/join` so reducing confusion for end users.

**others**
/kind feature

lonelyCZ:
Signed-off-by: lonelyCZ <531187475@qq.com>

/kind feature

yurtctl init/join/reset are not suitable as name for creating OpenYurt cluster and join nodes.
so rename to yurtadm init/join/reset for conforming to kubeadm init/join/reset so reducing confusion for end users.

```
[root@master68 bin]# ./yurtadm -h
yurtadm administers the yurt cluster

Usage:
  yurtadm [command]

Available Commands:
  help        Help about any command
  init        Run this command in order to set up the OpenYurt control plane
  join        Run this on any machine you wish to join an existing cluster
  reset       Performs a best effort revert of changes made to this host by 'yurtadm join'
```

```
[root@master68 bin]# ./yurtctl -h
yurtctl controls the yurt cluster

Usage:
  yurtctl [command]

Available Commands:
  cluster-info   list cloud nodes and edge nodes in cluster
  convert        Converts the kubernetes cluster to a yurt cluster
  help           Help about any command
  markautonomous mark the nodes as autonomous
  revert         Reverts the yurt cluster back to a Kubernetes cluster
  test           Tools for developers to test the OpenYurt Cluster
```

Fixes #737 

Since it is uncertain whether yurtctl will be retained, in order to reduce the workload, I don't move the dependency for now.

/assign @rambohe-ch @Congrool 

```release-note
rename yurtctl init/join/reset to yurtadm init/join/reset
```


pkg/yurtadm/cmd/cmd.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	"github.com/openyurtio/openyurt/pkg/yurtctl/cmd/join"
+	"github.com/openyurtio/openyurt/pkg/yurtctl/cmd/reset"
+	"github.com/openyurtio/openyurt/pkg/yurtctl/cmd/yurtinit"
```
rambohe-ch, line 31: please do not import `yurtctl/cmd/xxx`, and rename to `yurtadm/cmd/xxx`

lonelyCZ:
Since some of the functionality in `yurtctl` will be removed, it is possible to temporarily make `yurtctl` dependent on `yurtadm` to reduce the amount of renamed work and subsequent removal efforts.

pkg/yurtctl/cmd/clusterinfo/clusterinfo.go
```diff
@@ -30,7 +30,7 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
-	kubeutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/kubernetes"
+	kubeutil "github.com/openyurtio/openyurt/pkg/yurtadm/util/kubernetes"
```
Congrool, line 33: It looks strange that yurtctl import package of yurtadm.
Is it possible to move `pkg/yurtadm/util/kubernetes` to `pkg/util/kubernetes`

pkg/yurtctl/cmd/markautonomous/markautonomous.go
```diff
@@ -29,9 +29,9 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
-	"github.com/openyurtio/openyurt/pkg/yurtctl/constants"
+	"github.com/openyurtio/openyurt/pkg/yurtadm/constants"
```
Congrool, line 32: For `constants`, how about retaining the `constants` pkg both in `yurtctl` and `yurtadm`?
We can extract constants what yurtadm and yurtctl need respectively in `pkg/yurtadm/constants` and `pkg/yurtctl/constants`.
And each cmd only needs to import the constants pkg under its folder.

pkg/yurtctl/cmd/clusterinfo/clusterinfo.go
lonelyCZ: It can be, but `pkg/util/kubernetes` also depend on `pkg/yurtadm/kubernetes`.

pkg/yurtctl/cmd/markautonomous/markautonomous.go
lonelyCZ: Many constants is common, may also need to put the part in `pkg/util`.

Congrool: You can have a try, I don't find so many common constants. If there're only several common constants used by both yurtctl and yurtadm, the duplication is bearable.

pkg/yurtctl/cmd/clusterinfo/clusterinfo.go
Congrool: What do you mean? If we move `pkg/yurtadm/util/kubernetes` entirely to `pkg/util/kubernetes`, there's no dependence between them.

lonelyCZ: There's a lot of subtle dependencies between them. For example, `pkg/yurtadm/kubernetes/kubeadm/app/discovery/token` depend on `pkg/yurtadm/cmd/join/joindata`. And `pkg/yurtadm/kubernetes` and `pkg/yurtadm/util/kubernetes` are different files. :(

Congrool: Well, that's a problem. I find that yurtctl depends on the `yurtadm/kubernetes` mostly beacuse of convert/revert and test cmd. And test relies on `yurtadm/kubernetes` beacuse it reuses the code of convert. It's a solution that move test into yurtadm and remove convert/revert subcommand(which is under work in pr #748).

For dependence between `pkg/yurtadm/kubernetes/kubeadm/app/discovery/token` and `pkg/yurtadm/cmd/join/joindata`, can we move the joindata into `yurtadm/kubernetes`? It seems that `yurtadm/kubernetes` only depends on joindata pkg.

And what do you think @rambohe-ch ?

rambohe-ch: @Congrool @lonelyCZ The original idea of `yurtadm/kubernetes` or `yurtctl/kubernetes` is that codes comes from kubernetes project in order to avoid to import `k8s.io/kubernetes`. so move `joindata` into `yurtadm/kubernetes` maybe change the original idea.

how about deprecate `yurtctl convert/revert` at first?  and now we have no response form writer of #748, so maybe we can post a new pull request to solve this case.  @lonelyCZ would you like to take over this case?

lonelyCZ: Ok.

rambohe-ch:
@Congrool PTAL

pkg/yurtctl/cmd/markautonomous/markautonomous.go
Congrool: We only use the `constants.AnnotationAutonomy` in yurtctl, please move it from `pkg/yurtadm/constants` to `pkg/yurtctl/constants` and import `pkg/yurtctl/constants` instead.

pkg/yurtctl/cmd/yurttest/kindinit/init.go
```diff
@@ -30,11 +30,11 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 
-	"github.com/openyurtio/openyurt/pkg/yurtctl/constants"
-	enutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/edgenode"
-	kubeutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/kubernetes"
-	strutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/strings"
-	tmplutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/templates"
+	"github.com/openyurtio/openyurt/pkg/yurtadm/constants"
```
Congrool, line 33: We should import `pkg/yurtctl/constants` instead of `pkg/yurtadm/constants`. Any constants used in this file should be in the `pkg/yurtctl/constants`, even there's duplication. In this way, we can decouple yurtctl with yurtadm.

lonelyCZ:
/cc @Congrool  
PTAL

Congrool:
It's not an easy job to entirely decouple yurtadm with yurtctl, more work need to be put on it. In terms of this pr,  it looks good.
/lgtm

rambohe-ch (approved):

PR #822, opened 2022-05-07
add enable-node-pool parameter for yurthub in order to disable nodepools list/watch in filters when testing.

rambohe-ch:
/kind enhancement

add a `--enable-node-pool` parameter for yurthub component. by setting  up `--enable-node-pool=false` for disabling nodepools list/watch in filters so `local-up-openyurt.sh` can work when `nodepools` resource doesn't exist in cluster.

Fixes #

/assign @Congrool 

```release-note

```

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -27,10 +27,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/informers"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


Congrool (commented):
It looks strange to have an API-related option at the edgenode-side.
It's more suitable to diable the module that uses the nodepool resource. It looks like `ingressControllerFilter` and `serviceTopologyFilter` use nodepool resource. How about adding option `--disable-filters` instead of the `--enable-node-pool`?

rambohe-ch:
> It looks strange to have an API-related option at the edgenode-side. It's more suitable to diable the module that uses the nodepool resource. It looks like `ingressControllerFilter` and `serviceTopologyFilter` use nodepool resource. How about adding option `--disable-filters` instead of the `--enable-node-pool`?

@Congrool Thanks for your feedback. we already have `--disable-filters` parameter. In this case(start local-up-openyurt.sh, yurt-app-manager will not be installed) we don't want to disable `serviceTopology` filter, we only want to disable `service traffic inclose of pool topology` only, so we can continue use `service traffic inclose in node topology` ability. so a new parameter `--enable-node-pool` is added.

PR #821, opened 2022-05-07
adapt k8s v1.22 version

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

#### What this PR does / why we need it:

Adapt openyurt support k8s 1.22 version.

#### Which issue(s) this PR fixes:

Fixes #

##### API removals
We need upgrade api versions. And should care the pr https://github.com/openyurtio/yurt-app-manager/pull/49 too. 
https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/

##### Kubernetes signers
As the default signers have restriction, we need implement myself. 
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers

##### StreamingProxyRedirects Deprecated
As the deprecation, openyurt's tunnel-server proxy not work as before.
https://github.com/kubernetes/enhancements/issues/1558
https://bugs.launchpad.net/charm-kubernetes-master/+bug/1940527

#### Does this PR introduce a user-facing change?

```release-note

```

PR #823, opened 2022-05-09
support multi k8s ci

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Use github's matrix feature, we could test openyurt adapt multi k8s versions.

Fixes #

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # openyurtio/openyurt
-
+xxx
 <div align="center">
 
 <img src="docs/img/OpenYurt.png" width="400" height="94"><br/>


rambohe-ch:
@Congrool PTAL

.github/workflows/ci.yaml
```diff
@@ -112,8 +116,43 @@ jobs:
           restore-keys: ${{ runner.os }}-go-
       - name: Build All Binaries
         run: make build
+
+  detect-noop:
```
Congrool, line 120: Could you please give a little more information of the `detect-noop` job about why we need it? It seems to have nothing to do with the multi-k8s ci.

Congrool (commented):
If I don't misunderstand the purpose of this pr, it is for #821. Because k8s v1.22 change the CSR API, we have to support versions before v1.22 and also support the versions after v1.22.

PR #829, opened 2022-05-13
【ADD】update node servant revert step

JameKeal:
Sometimes the node has residual certificate files related to tunnels, if we still use this nodes to join the cluster, the invalid certificate cannot create the grpc tunnel, it can make the error for connection closed.

If we don’t open the grpc process debug parameter, it’s hard to find it. So we need to clean the tunnel directory when node-servant do the revert option.

Fixes #

```release-note

```

### DIFF
--- a/pkg/node-servant/components/tunnel.go
+++ b/pkg/node-servant/components/tunnel.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2021 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


pkg/node-servant/components/tunnel.go
```diff
@@ -0,0 +1,60 @@
...
+	"os"
+
+	"k8s.io/klog/v2"
+
+	enutil "github.com/openyurtio/openyurt/pkg/yurtctl/util/edgenode"
+)
+
+const (
+	yurttunnelAgentWorkdir  = "/var/lib/yurttunnel-agent"
+	yurttunnelServerWorkdir = "/var/lib/yurttunnel-server"
```
rambohe-ch, line 29: please use `projectinfo.GetServerName() and projectinfo.GetAgentName()` to define these two vars.

JameKeal: OK, It's done.

rambohe-ch (approved):

PR #832, opened 2022-05-15
Cleanup: use verb %w with fmt.Errorf() when generate new wrapped error from error

zhaodiaoer:
After go1.13, it is best to use %w in fmt.Errorf() when generate a new wrapped error from error, 
and use errors.Is() or errors.As() when compare between error instance

> None

```release-note
None
```

#### other Note

> None


pkg/preflight/checks.go
```diff
@@ -273,7 +273,7 @@ func (dac DirExistingCheck) Check() (warnings, errorList []error) {
 	defer f.Close()
 
 	_, err = f.Readdirnames(1)
-	if err == io.EOF {
+	if errors.Is(err, io.EOF) {
```
rambohe-ch, line 276: There are some introduction about `io.EOF` In the link: https://github.com/golang/go/wiki/ErrorValueFAQ
> Comparisons to io.EOF need not be changed, because io.EOF should never be wrapped.

so i think maybe we should keep the `if err == io.EOF` comparison code. how about recover all of modifications about `io.EOF` comparisons?

rambohe-ch:
@zhaodiaoer Thanks for raising pull request. please fix comment and unit test errors.

pkg/preflight/checks.go
zhaodiaoer: ok, i will fix all

zhaodiaoer:
@rambohe-ch Thanks, i have fixed all the buggy usage related to io.EOF and have successfully run the unit tests.

rambohe-ch (approved):

PR #838, opened 2022-05-18
use new version `BootstrapToken`

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

1) Kubeadm alreay move `BootstrapToken` to new package.
2) Prepare for support `yurtadm token` command which just like `kubeadm token`, for join node case.

Fixes #

```release-note

```


lonelyCZ:
Thanks, @huiwq1990 .

Why just update this apis? I am worried that if we only update part of dependence of `kubeadm`, we will have difficulty maintaining the entire dependency, because we depend on `kubeadm v1.20.11` presently.

PR #842, opened 2022-05-20
support ipv6 network

tydra-wang:
#### What type of PR is this?
/kind enhancement

We need to run openyurt on ipv6 network.

Current works makes yurt-tunnel and yurthub work well on ipv6 network. 

- yurt-tunnel-server

  - determine protocol stack according to bind ip format
  - set default insecure bind address to ::1 when ipv6
  - use ip6tables instead of iptables when ipv6

- yurt-tunnel-agent

  - determine protocol stack according to node ip format
  - set default meta host to ::1 when ipv6
 
- yurthub

  - use `fd00::2:1` as the default dummy if ip when binding an ipv6 address(yurthub still use 127.0.0.1 as default, set --bind-address to ::1 in ipv6 )
  - use ip6tables instead iptables when dummy ip is an ipv6 address
 
  use fd00::2:1 in ipv6, as the counterpart of 169.254.2.1 in ipv4, referring to [node local dns cache ipv6 configuration](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/#configuration). 

- other changes:

  - use `net.JoinHostPort` instead of `fmt.Sprintf("%s:%d", host, port)`
  - use ::1 as loopback address when ipv6
  - add ::1 to cert SANs

### DIFF
--- a/cmd/yurt-tunnel-agent/app/options/options.go
+++ b/cmd/yurt-tunnel-agent/app/options/options.go
@@ -25,10 +25,12 @@ import (
 
 	"github.com/spf13/pflag"
 	"k8s.io/klog/v2"
+	utilnet "k8s.io/utils/net"
 	"sigs.k8s.io/apiserver-network-proxy/pkg/agent"
 
 	"github.com/openyurtio/openyurt/cmd/yurt-tunnel-agent/app/config"
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	utilip "github.com/openyurtio/openyurt/pkg/util/ip"
 	"github.com/openyurtio/openyurt/pkg/yurttunnel/constants"
 	kubeutil "github.com/openyurtio/openyurt/pkg/yurttunnel/kubernetes"
 )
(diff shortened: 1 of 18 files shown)


rambohe-ch:
@tydra-wang Very appreciate you for raising pull request to support IPV6 for OpenYurt. please fix golangci lint error. and you can run `make lint` locally to check the errors.

tydra-wang:
`make lint`  succeeded in my local env, while workflow still failed. I didn't see any specified files in the workflow log.

cmd/yurt-controller-manager/controller-manager.go
```diff
@@ -26,10 +26,8 @@ import (
 	"time"
 
 	"k8s.io/component-base/logs"
-
 	// load all the prometheus client-go plugin
```
Congrool, line 29: The golangci-lint error can be find here.
It seems that we should retain the empty line above the annotated import package.

cmd/yurt-controller-manager/controller-manager.go
```diff
@@ -26,10 +26,8 @@ import (
...
 	_ "k8s.io/component-base/metrics/prometheus/clientgo"
-
 	// for version metric registration
```
Congrool, line 31: ditto

cmd/yurt-controller-manager/controller-manager.go
tydra-wang: thank you.

i removed the empty lines because my local golangci-lint (the latest version) throw errors for them.

rambohe-ch:
@tydra-wang please rebase the newest master branch, and fix the conflicts.

rambohe-ch (approved):

PR #843, opened 2022-05-20
[enhancement] add release workfolow for daily release and tag event.

rambohe-ch:
/kind documentation

Add a release workflow for updating images daily and updating images when a new tag is created.

Fixes #

/assign @Congrool 

```release-note

```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,57 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v*"
+  schedule:
+    # run at UTC 1:30 every day
+    - cron:  '30 1 * * *'
+  workflow_dispatch: {}
+
+env:
+  ALI_REGISTRY: registry.cn-hangzhou.aliyuncs.com/openyurt
+
+jobs:
+  docker-push:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: true
+        id: get_version
+        run: |
+          VERSION=${GITHUB_REF#refs/tags/}
+          if [[ ${GITHUB_REF} == "refs/heads/master" ]]; then
+            VERSION=latest
+          fi
+          echo ::set-output name=VERSION::${VERSION}
+      - name: Install Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Release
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Install Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Release
+        run: make docker-push TARGET_PLATFORMS=linux/amd64,linux/arm64,linux/arm/v7 IMAGE_TAG=${{ steps.get_version.outputs.VERSION }}
```
Congrool, line 39: It seems to should be `make docker-build` .

Congrool (commented):
Seems some problems here:
If we are not at the tag, the image-tag should be `latest`, otherwise it should be `latest`. However, make rule will tag the image as `GIT_VERSION` which is based on the `IMAGE_TAG` and not be exposed to outside. Thus, what we set in `IMAGE_TAG` will not work.
Maybe we have to revise the makefile to make it work.

rambohe-ch:
> Seems some problems here: If we are not at the tag, the image-tag should be the tag, otherwise it should be `latest`. However, make rule will tag the image as `GIT_VERSION` which is based on the `IMAGE_TAG` and not be exposed to outside. Thus, what we set in `IMAGE_TAG` will not work. Maybe we have to revise the makefile to make it work.

@rambohe-ch Thanks for your feedback. and I have exposed `IMAGE_TAG` as global env, so end user can set this env in the github action.

.github/workflows/release.yaml
rambohe-ch: @Congrool because i want to use `docker buildx build --push`, so `make docker-push` command has been added.

Congrool:
>  Thanks for your feedback. and I have exposed IMAGE_TAG as global env, so end user can set this env in the github action.

Well, I mean `GIT_VERSION` is not exposed to the outside, it is the tag what images will actually use. You can find it here:
https://github.com/openyurtio/openyurt/blob/master/Makefile#L99

The arguments of `-t` option is `${IMAGE_REPO}/yurthub:${GIT_VERSION}`, not the ${IMAGE_TAG}. So, I think it may need to be revised.

rambohe-ch:
> > Thanks for your feedback. and I have exposed IMAGE_TAG as global env, so end user can set this env in the github action.
> 
> Well, I mean `GIT_VERSION` is not exposed to the outside, it is the tag what images will actually use. You can find it here: https://github.com/openyurtio/openyurt/blob/master/Makefile#L99
> 
> The argument of `-t` option is `${IMAGE_REPO}/yurthub:${GIT_VERSION}`, not the ${IMAGE_TAG}. So, I think it may need to be revised.

@Congrool fixed it.

Congrool:
/lgtm

It looks good to release image daily. But these logic in makefile will make it a little bit complicated and difficult to use. It's better the move them into a script. Maybe we can do it in further work.

PR #844, opened 2022-05-20
Add orgs to yurthub server crt

MrGirl:
we use tenant ns as org  when we do client csr, but when we do server csr, the org is not used. And the csr will be rejected by beehive 'cause beehive think the server's csr tenant ns is not equal to it should be.

This pr use command parameter "YurtHubCertOrganizations" as org to ask for its server crt.

As k8s 1.22 does not allow custom orgs except "system:nodes", the orgs will be added into ExtraExtension 

@rambohe-ch 

NONE
```release-note

```

### DIFF
--- a/pkg/util/certmanager/factory/factory.go
+++ b/pkg/util/certmanager/factory/factory.go
@@ -20,8 +20,10 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"fmt"
 	"net"
+	"strings"
 
 	certificatesv1 "k8s.io/api/certificates/v1"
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@MrGirl In order to adapt Yurthub for K8s v1.22, Settings for Yurthub server cert manager needed to be changed, and Org for cert can only be `system:nodes`, so we need to add tenant info into dnsNames of CSR.

rambohe-ch:
@MrGirl please rebase the newest master branch, and add tenant info into CSR dnsNames.

MrGirl:
As below, tenant ns is added into extension
image

MrGirl:
As the newest code has add orgs into csr when yurthub ask for the client certificate, there's no need to patch this pull request, close it

Issue #859, opened 2022-05-27
[feature request] yurtadm token subcommand


huiwq1990:
**What would you like to be added**:

Add a `token` subcommand for admin when need to join an edge node.

**Why is this needed**:

When join a edge node by execute `yurtadm join`, we should set token parameter. Openyurt should provide a way to generate the token, for examle imitate `kubeadm join`'s implementation.

**others**
/kind feature

Congrool:
sounds good, agree +1

rambohe-ch:
@huiwq1990 Very appreciate for your suggestion. would you be able to take over this work?

huiwq1990:
I wolud like to implement this. There ary many subcommends, like `yurtadm token create`, `yurtadm token list` and so on. How many features do `yurtadm token` should cover. I suggest we just implement `yurtadm token create`, and  close the loop on `yurtadm`. If openyurt users want to use more management functions, we cloud recommend them use `kubeadm`. 
@Congrool @rambohe-ch What's your suggestions?

rambohe-ch:
> I wolud like to implement this. There ary many subcommends, like `yurtadm token create`, `yurtadm token list` and so on. How many features do `yurtadm token` should cover. I suggest we just implement `yurtadm token create`, and close the loop on `yurtadm`. If openyurt users want to use more management functions, we cloud recommend them use `kubeadm`. @Congrool @rambohe-ch What's your suggestions?

@huiwq1990 I think feature of command `yurtadm token create` is the same as `kubeadm token create`. and how about wrap `kubeadm token` in `yurtadm token` command, so maybe we will not confused how many features does `yurtadm token` should cover.

PR #874, opened 2022-06-06
support  yurtadm token subcommand

huiwq1990:
/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Fixes https://github.com/openyurtio/openyurt/issues/859

```release-note

```


rambohe-ch:
@huiwq1990 Thanks for posting pull request. and i want to introduce my opinions about `yurtadm token` command
1. The functions of `yurtadm token` command are the same as `kubeadm token` command
2. so we can wrap `kubeadm token` command in `yurtadm token`, and download kubeadm binary if it is not exist.
3. Maybe it's a good idea to migrate code of `kubeadm token` into yurtadm, because it will carry a lot of work for maintaining these codes.

PR #875, opened 2022-06-08
support yurtadm token subcommand

huiwq1990:
Signed-off-by: huiwq1990 <huiwq1990@163.com>

/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage
> /sig storage

Fixes https://github.com/openyurtio/openyurt/issues/859

```release-note

```

### DIFF
--- a/pkg/yurtadm/cmd/token/token.go
+++ b/pkg/yurtadm/cmd/token/token.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2022 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


pkg/yurtadm/cmd/token/token.go
```diff
@@ -0,0 +1,129 @@
...
+		FParseErrWhitelist: cobra.FParseErrWhitelist{UnknownFlags: true},
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			if err := CheckAndInstallKubeadm(); err != nil {
+				return err
+			}
+
+			klog.V(2).InfoS("kubeadm command exec", "args", os.Args[2:])
+
+			kubeadmCmd := exec.Command("kubeadm", os.Args[2:]...)
```
rambohe-ch, line 59: I am not sure that `os.Args[2:]` should be `os.Args[1:]` or not?

huiwq1990: It should be `os.Args[1:] `, otherwise the command will become something like `yurtadm token token list`.

Issue #778, opened 2022-03-10
[feature request]`NodePool Governance Capability`: YurtHub supports writing data to pool-spirit


Peeknut:
**What would you like to be added**:
Yurthub needs to add the following features:

-  When the pool-spirit starts or migrates and rebuilds, actively establish a connection with the pool-spirit and send local data to the pool-spirit.
- When the local data is updated, the changed data needs to be synchronized to the pool-spirit.

Since multiple nodes are likely to update a resource at the same time, conflicts need to be considered.

 **Notes**: For details, please refer to the proposal: https://github.com/openyurtio/openyurt/pull/772

**Why is this needed**:
As mentioned in the proposal(https://github.com/openyurtio/openyurt/pull/772), users can obtain resource data or kubectl exec/logs in the node pool dimension through pool-spirit. Considering that there is no guarantee that pool-spirit is always connected to the cloud, in order to ensure real-time data, the data in pool-spirit is updated by each yurthub.

**others**
/kind feature

Congrool:
Hi，I want to take over this work.
/assign

Congrool:
This work is almost finished, including:
1. add a client to send objs in response to etcd in pool-coordinator.
2. add a reverse proxy to re-direct requests to apiserver in pool-coordinator when the node cannot reach the cloud.

Some problems still unsolved:
1. cannot cache secret, which is used when using InClusterConfig to setup pod.
2. unit tests have not been revised
3. resource gc in pool-coordinator. Considering that when an edge component uses GET request to get a resource, this resource will be cached into the pool-coordinator, but when to delete it?
4. support node-scope and pool-scope cache at the same time

Congrool:
For **ReverseProxy**, the processing logic of request from edge side is something like:
```go
switch{
case cloud.IsHealthy:
   if poolCoordinator.IsHealthy && ! LeaderYurthub && IsPoolScopeResource(req.Resource) {
    fallthrough
  }
  if err := cloud.Handle(req); err != nil {
    fallthrough
  }
case poolCoordinator.IsHealthy:
  if err := poolCoordinator.Handle(req); err != nil {
    fallthrough
  }
default:
  localCache.Handle(req)
}
```

So I need a way to determine whether this yurthub is leader yurthub. 
What do you think? @rambohe-ch

rambohe-ch:
> For **ReverseProxy**, the processing logic of request from edge side is something like:
> 
> ```go
> switch{
> case cloud.IsHealthy:
[...]

Issue #857, opened 2022-05-27
ASOC 2022: Make container network solution(calico) work on the OpenYurt even if cloud-edge network off


rambohe-ch:
# Describe your problem
  Deploy container network solution(calico) in OpenYurt cluster, and make calico network can work even if edge nodes restart when cloud-edge network off.
  Meanwhile, when the edge network can not work properly for some unknown reasons(like iptables settings fault), provide a way or a tool to detect the network problem.

  在OpenYurt集群中部署calico容器网络方案，并且保证在云边网络断连状态下即使边缘节点重启calico网络也能正常工作。同时，当边缘网络由于某些未知原因(如iptables设置错误)无法正常工作时，提供一种检测网络问题的方法或工具。

# Additional context
This issue is part of our [Alibaba Summer of Code (2022) Program](https://github.com/openyurtio/openyurt/issues/851)

Difficulty: Hard
Mentor: wangchenglong (@luckymrwang )

rambohe-ch:
@Congrool please comment and close this issue.

Congrool:
Work of this issue has been completed in #882 . #882 introduces great changes to yurthub which not only containing supporting calico and should be merged according to the community road map.
.
So, I close this issue. If any further problem found, we can discuss in #882.

PR #882, opened 2022-06-14
refactor yurthub cache to adapt different storages

Congrool:

#### What type of PR is this?
/kind enhancement
/sig storage

Currently, the cacheManager deeply couple with the diskStorage. In #778 , we'll add a new storage backend for cacheManager, called pool-coordiantor which actually an etcd storage at the view of cacheManager. There're some problems need to be solved before implement this feature:
1. the format of key for disk storage is different from that of pool-coordinator, the former is `/component/resource/namespace/name` and the later is `/registry/resource/namespace/name`. Also considering that there're some special case for pool-coordinator, such as resource `nodes` should be `minions` instead.
2. For update operation, there's also difference. For disk storage, we have to decode the stored json data and retrieve the rv field, but for pool-coordinator we can directly compare the rv in etcd. Also considering that, the comparision and update should be atomic, so we should not compare the rv in cache manager and update in storage.
3. The [storage.Store](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/storage/store.go#L39-L48) interface is not explict enough. We should define the operation of each function before implementing other storages.
4. We should make the in-memory cache at the level of cachemanager, which is used to optimize the cache query effiency, because there're two many kinds that can read/write storage, it's not easy to manage the in-memory cache. But at the cachemanager layer, only CacheResponse will write to the storage and QueryCache will read from the storage, the management of in-memory cache will be much more easier.


Congrool:
TODO List:
1. revise unit tests 
2. check the cache ability in actual environment

Congrool:
Proposal： #897

pkg/yurthub/storage/disk/key_test.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	disk, err := NewDiskStorage(keyFuncTestDir)
+	if err != nil {
+		t.Errorf("failed to create disk store, %v", err)
+		return
+	}
+	keyFunc := disk.KeyFunc
+	for c, s := range cases {
+		key, err := keyFunc(s.info)
```
rambohe-ch, line 93: you can add all the test contents in the following snippet:
```
t.Run(k, func(t *testing.T){
     xxxx
})
```

Congrool: @rambohe-ch Thanks, I'll fix it.

I'm now working on the autonomy test in the actual environment, and stucked at some problems. I'll create a issue talking about it later. So I think, this pr is not ready for review.

I'll remove the WIP, if I finish all the job.

Congrool:
I reconsider the timing when merge this pr. It seems more work need to be done that original thought. Is it better to merge this pr after v1.0? 

Meanwhile, I'll continue my work about autonomy tests, #778 and #857 based on this pr.

pkg/yurthub/util/fs/store.go
```diff
@@ -0,0 +1,288 @@
...
+		if !ok {
+			return nil, ErrIsNotDir
+		}
+		return nil, err
+	}
+
+	dirs := []string{}
+	files := []string{}
+	if isRecursive {
+		filepath.Walk(rootDir, func(path string, info os.FileInfo, err error) error {
```
rambohe-ch, line 107: please use `filepath.WalkDir()` because it's more efficient.

pkg/yurthub/util/fs/store.go
```diff
@@ -0,0 +1,288 @@
...
+		if ok, err := IsDir(path); !ok && err == nil {
+			return ErrIsNotDir
+		}
+		return ErrExists
+	}
+
+	if err := os.MkdirAll(path, 0755); err != nil {
+		return err
+	}
+	return nil
```
rambohe-ch, line 205: optimize codes from 202~205: return os.MkdirAll(path, 0755)

pkg/yurthub/util/fs/store.go
```diff
@@ -0,0 +1,288 @@
...
+	if IfExists(path) {
+		if ok, err := IsRegularFile(path); !ok && err == nil {
+			return ErrIsNotFile
+		}
+		return ErrExists
+	}
+
+	// ensure the base dir
+	dir := filepath.Dir(path)
+	if _, err := os.Stat(dir); err != nil {
```
rambohe-ch, line 223: line 214: `IfExists` has already called `os.Stat` to check existence of path, so we should not call `os.Stat()` twice here.

Congrool: The first IfExists checks if the file exists, and the second check if the dir of this file exists. We need to create the dir if it does not exist before create the file.

pkg/yurthub/util/fs/store.go
Congrool: fixed

pkg/yurthub/util/fs/store.go
Congrool: fixed

pkg/yurthub/util/fs/store.go
```diff
@@ -0,0 +1,288 @@
...
+func (fs *FileSystemOperator) Read(path string) ([]byte, error) {
+	if !IfExists(path) {
+		return nil, ErrNotExists
+	}
+
+	if ok, err := IsRegularFile(path); !ok || err != nil {
+		if !ok {
+			return nil, ErrIsNotFile
+		}
+		return nil, err
```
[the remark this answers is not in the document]
Congrool: fixed

pkg/yurthub/storage/disk/key_test.go
Congrool: fixed

Congrool:
I've rebased the pr of autonomy e2e tests, and seems that all tests can pass. I think it's ready for review, PTAL @rambohe-ch

pkg/yurthub/storage/store.go
```diff
@@ -16,33 +16,115 @@ limitations under the License.
...
 
-// ErrStorageAccessConflict is an error for accessing key conflict
-var ErrStorageAccessConflict = errors.New("specified key is under accessing")
+// Store is an interface for caching data into store
+type Store interface {
+	// Name will return the name of this store.
+	Name() string
+	clusterInfoRelatedInterface
+	objectRelatedInterface
+	componentRelatedInterface
```
rambohe-ch, line 41: The definition of interfaces should be match with proposal. 
for example: clusterInfoRelatedInterface --> clusterInfoHandler

Congrool: fixed

pkg/yurthub/kubernetes/meta/restmapper.go
```diff
@@ -135,16 +151,17 @@ func (rm *RESTMapperManager) deleteKind(gvk schema.GroupVersionKind) error {
...
 	rm.RUnlock()
 	if err != nil {
 		return err
 	}
-	return rm.storage.Update(CacheDynamicRESTMapperKey, d)
+	err = rm.storage.Write(rm.cachedFilePath, d)
+	if err != nil {
+		return fmt.Errorf("failed to update cached dynamic RESTMapper, %v", err)
+	}
+	return nil
```
rambohe-ch, line 164: how about modify Line160~Line164 to `return rm.storage.Write(rm.cachedFilePath, d)`?

pkg/yurthub/kubernetes/meta/restmapper.go
```diff
@@ -193,7 +210,7 @@ func (rm *RESTMapperManager) UpdateKind(gvk schema.GroupVersionKind) error {
 // and delete the corresponding file in the disk (cache-crd-restmapper.conf), it should be used carefully.
 func (rm *RESTMapperManager) ResetRESTMapper() error {
 	rm.dynamicRESTMapper = make(map[schema.GroupVersionResource]schema.GroupVersionKind)
-	err := rm.storage.DeleteCollection(CacheDynamicRESTMapperKey)
+	err := rm.storage.DeleteFile(rm.cachedFilePath)
 	if err != nil {
 		return err
 	}
```
rambohe-ch, line 216: how about modify Line 213~217 to `return rm.storage.DeleteFile(rm.cachedFilePath)`?

pkg/yurthub/cachemanager/storage_wrapper.go
```diff
@@ -196,74 +171,77 @@ func (sw *storageWrapper) List(key string) ([]runtime.Object, error) {
...
+	return obj, nil
 }
 
-// Replace will delete the old objects, and use the given objs instead.
-func (sw *storageWrapper) Replace(rootKey string, objs map[string]runtime.Object) error {
+func (sw *storageWrapper) ReplaceComponentList(component string, gvr schema.GroupVersionResource, namespace string, objs map[storage.Key]runtime.Object) error {
 	var buf bytes.Buffer
-	contents := make(map[string][]byte, len(objs))
+	contents := make(map[storage.Key][]byte, len(objs))
+	rvs := make(map[storage.Key]int64, len(objs))
```
rambohe-ch, line 198: rvs is unused?

pkg/yurthub/cachemanager/storage_wrapper.go
```diff
@@ -131,41 +118,29 @@ func (sw *storageWrapper) Get(key string) (runtime.Object, error) {
 	} else {
 		UnstructuredObj = new(unstructured.Unstructured)
 	}
-	obj, gvk, err := sw.backendSerializer.Decode(b, nil, UnstructuredObj)
+	obj, gvk, err := sw.universalDecoder.Decode(b, nil, UnstructuredObj)
```
rambohe-ch, line 121: why not use `jsonEncoder` here?

pkg/yurthub/cachemanager/storage_wrapper.go
```diff
@@ -184,9 +159,9 @@ func (sw *storageWrapper) List(key string) ([]runtime.Object, error) {
 			UnstructuredObj = new(unstructured.Unstructured)
 		}
 
-		obj, gvk, err := sw.backendSerializer.Decode(bb[i], nil, UnstructuredObj)
+		obj, gvk, err := sw.universalDecoder.Decode(bb[i], nil, UnstructuredObj)
```
rambohe-ch, line 162: ditto

pkg/yurthub/cachemanager/storage_wrapper.go
```diff
@@ -196,74 +171,77 @@ func (sw *storageWrapper) List(key string) ([]runtime.Object, error) {
...
+		return nil, err
 	}
 
-	if isCacheKey(key) {
-		sw.Lock()
-		sw.cache[key] = obj
-		sw.Unlock()
+	if buf, err := sw.store.Update(key, buf.Bytes(), rv); err != nil {
+		if err == storage.ErrUpdateConflict {
+			obj, _, dErr := sw.universalDecoder.Decode(buf, nil, nil)
```
rambohe-ch, line 183: ditto

pkg/yurthub/cachemanager/cache_manager.go
```diff
@@ -124,90 +126,183 @@ func (cm *cacheManager) QueryCache(req *http.Request) (runtime.Object, error) {
...
 			// if no pods in cache, maybe all of pods have been deleted by accident,
 			// if empty object is returned, pods on node will be deleted by kubelet.
 			// in order to prevent the influence to business, return error here so pods
 			// will be kept on node.
+			klog.Warningf("get 0 pods for kubelet pod request, there should be at least one, hack the response with ErrStorageNotFound error")
+			return nil, storage.ErrStorageNotFound
+		}
+		// There's no obj to fill in the list, so just return.
+		return listObj, nil
+	}
```
rambohe-ch, line 205: optimize Line184~205 as following will be more readable:
```
if err == storage.ErrStorageNotFound && proxyutil.IsListRequestWithNameFieldSelector(req) {
    // xxx
} else if err != nil {
    // xxx
} else if len(objs) == 0 {
    // xxx
}
```

pkg/yurthub/cachemanager/storage_wrapper.go
Congrool: I encountered the problem that the response from APIServer is not JSON format, but protobuf format. So I replace the decoder as universalDecoder to solve it. Since then, a lot of changes have be placed on code. Maybe we do not need to use universalDecoder here anymore.

pkg/yurthub/kubernetes/meta/restmapper.go
Congrool: Here, we may need error info with more details to specify which component failed, otherwise we can only get the error log like "path does not exist", unknown about which path does not exsit which may make us confused.

pkg/yurthub/kubernetes/meta/restmapper.go
Congrool: fixed

pkg/yurthub/cachemanager/storage_wrapper.go
Congrool: fixed

pkg/yurthub/cachemanager/storage_wrapper.go
Congrool: fixed

pkg/yurthub/cachemanager/storage_wrapper.go
Congrool: fixed

pkg/yurthub/cachemanager/storage_wrapper.go
Congrool: fixed

pkg/yurthub/cachemanager/cache_manager.go
Congrool: fixed

pkg/yurthub/storage/disk/key.go
```diff
@@ -0,0 +1,73 @@
...
+	if info.Name == "" {
+		isRoot = true
+	}
+
+	group := info.Group
+	if info.Group == "" {
+		group = "core"
+	}
+
+	gvrName := strings.Join([]string{info.Resources, info.Version, group}, ".")
```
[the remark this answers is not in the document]
Congrool: fixed

pkg/yurthub/storage/disk/key.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (k storageKey) isRootKey() bool {
+	return k.rootKey
+}
+
+// Key for disk storage is
+// /<Component>/<Resource>/<Namespace>/<Name>, or
+// /<Component>/<Resource>/<Name>, if there's no namespace provided in info.
+// /<Component>/<Resource>/<Namespace>, if there's no name provided in info.
+// /<Component>/<Resource>, if there's no namespace and name provided in info.
```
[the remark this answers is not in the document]
Congrool: fixed

Congrool:
Ensure the compability with old key format. 

The disk storage will use old key format if it detects there's old key cache under the cache dir.
Users needs to manually delete all old cache before updating the yurthub, if they want to use the enhancement mode of yurthub which uses `/<component>/<resource.version.group>/<namespace>/<name>` as key format.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -457,51 +503,172 @@ func (ds *diskStorage) Recover(key string) error {
...
+
+func (ds *diskStorage) unLockKey(key storageKey) {
 	ds.Lock()
 	defer ds.Unlock()
-	delete(ds.keyPendingStatus, key)
+	delete(ds.keyPendingStatus, key.Key())
+}
+
+func ifEnhancement(baseDir string, fsOperator fs.FileSystemOperator) (bool, error) {
+	resDirs, err := fsOperator.List(baseDir, fs.ListModeDirs, false)
```
rambohe-ch, line 600: The format of dir's name under `baseDir` is `<component>`, and doesn't include `<Resource.Version.Group>` info.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -46,408 +60,440 @@ func NewDiskStorage(dir string) (storage.Store, error) {
...
+
 	ds := &diskStorage{
 		keyPendingStatus: make(map[string]struct{}),
 		baseDir:          dir,
+		serializer:       json.NewSerializerWithOptions(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, json.SerializerOptions{}),
+		fsOperator:       fsOperator,
 	}
 
-	err := ds.Recover("")
+	ifEnhancement, err := ifEnhancement(ds.baseDir, *ds.fsOperator)
```
rambohe-ch, line 80: how about rename var `ifEnhancement` to `enhancementMode`?

pkg/yurthub/storage/disk/storage.go
Congrool: fixed

pkg/yurthub/storage/disk/storage.go
Congrool: fixed

rambohe-ch (approved):

PR #897, opened 2022-06-27
proposal of yurthub cache refactoring

Congrool:
#### What type of PR is this?
/kind documentation
/kind design
/sig storage

When implementing the feature of pool-coordinator, I encountered serval problems to be solved first. So, it is this pr. Details are in the proposal.

Fixes #

```release-note

```


rambohe-ch:
@Congrool Maybe we should also consider the interface for reading objects from yurthub cache?

rambohe-ch:
> @rambohe-ch Currently we have Get/List/ListKeys only for reading objects in diskStorage. In the case of Pool-Coordiantor, the read request will be send to the APIServer in pool-coordinator, so it does not need to use the read interface to read object from cache directly.

@Congrool  ok, i think pool-coordinator will not implement Get/List/ListKeys funcs for Storage interface because yurthub read from kube-apiserver of pool-coordinator directly.

Congrool:
need revision
/hold

rambohe-ch (approved):

Congrool:
revised according to the newest implementation

rambohe-ch:
@Congrool how about the definition of `KeyBuildInfo` and `Key` struct?

Congrool:
The definition of Key is like:
```go
type Key interface {
	Key() string
}

type KeyBuildInfo struct {
	Component string
	Namespace string
	Name      string
	Resources string
	Group     string
	Version   string
}
```

I'll update it in this proposal.

Congrool:
proposal is updated with definition of Key and new interface names

rambohe-ch (approved):

PR #902, opened 2022-07-05
[Proposal] add yurtadm proposal

windydayc:
#### What type of PR is this?
/kind feature
/kind design

add yurtadm proposal.

/assign @rambohe-ch 
/assign @Peeknut


rambohe-ch:
@windydayc because `sealer` can not support render `openyurt version info` into kubefile, so it's not convenient to build OpenYurt cluster image in a github action pipeline. 

and we need to follow the sealer progress, then we can add a new pipeline when sealer supports it. so i will close this proposal now.

PR #921, opened 2022-07-19
[Proposal] add proposal of daemosnet update model

hxcGit:
Signed-off-by: hxcGit <houxc_mail@163.com>

/kind design

#### What this PR does / why we need it:
Optimize workload update model in cloud-edge collaboration scenarios

#### Which issue(s) this PR fixes:

#914 

#### Special notes for your reviewer:

/assign @rambohe-ch 

#### Does this PR introduce a user-facing change?

```release-note

```

PR #922, opened 2022-07-19
CI fix: codecov failure 

Congrool:
#### What type of PR is this?
/kind bug

non-code files are included when codecov computes the coverage. Thus, it always fails when we add docs because the test coverage rate will decrease.

For example: [pr921](https://github.com/openyurtio/openyurt/pull/921/checks?check_run_id=7401365549)

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/masterservice/handler_test.go
+++ b/pkg/yurthub/filter/masterservice/handler_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"io"
 	"net/http"
+	"sync"
 	"testing"
 	"time"
 
@@ -421,6 +422,7 @@ func TestStreamResponseFilter(t *testing.T) {
 
 	for k, tt := range testcases {
 		t.Run(k, func(t *testing.T) {
+			wg := sync.WaitGroup{}
 			req, _ := http.NewRequest(tt.verb, tt.path, nil)
 			req.Header.Set("User-Agent", tt.userAgent)
 			req.Header.Set("Accept", tt.accept)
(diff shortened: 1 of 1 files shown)


Congrool:
It seems that codecov should ignore some paths which do not contain code.

Congrool:
The codecov.yml refers to https://docs.codecov.com/docs/ignoring-paths.

Congrool:
@rambohe-ch  PTAL

Congrool:
Seems `ignore` cannot work well. Need to dig a little more deeper.

Congrool:
The conclusion is that the failure is not caused by the codecov configuration, but the test itself.

An analysis why we have the coverage jitter at `pkg/yurthub/filter/masterservice/handler_test.go`.

In [TestStreamResponseFilter](https://github.com/openyurtio/openyurt/blob/39a28c16ea6d89673306865733c998b5e1102577/pkg/yurthub/filter/masterservice/handler_test.go#L250-L464), we run serval goroutines to test the [StreamResponseFilter](https://github.com/openyurtio/openyurt/blob/39a28c16ea6d89673306865733c998b5e1102577/pkg/yurthub/filter/masterservice/handler.go#L93-L136), but we exit the test without waiting for all goroutine return. Thus the [defer block in StreamResponseFilter](https://github.com/openyurtio/openyurt/blob/39a28c16ea6d89673306865733c998b5e1102577/pkg/yurthub/filter/masterservice/handler.go#L94-L96) may run or not run depending on the runtime case. 

Then we find that sometiimes the coverage rate is a bit higher and sometimes not.

Congrool:
Not sure if we have similiar problems in our existing unit tests, maybe it's better to set the threshold for codecov like [this](https://docs.codecov.com/docs/common-recipe-list#:~:text=threshold%3A%201%25%20%20%23%20the%20leniency%20in%20hitting%20the%20target)

PR #927, opened 2022-07-22
clean redundant lint in makefile

Congrool:
#### What type of PR is this?
/kind clean

We have already had `vet` and `fmt` in [golangci-lint](https://github.com/openyurtio/openyurt/blob/master/.golangci.yaml#L42-L48). So we don't need to run them before running unit tests.

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ ifneq (${https_proxy},)
 DOCKER_BUILD_ARGS += --build-arg https_proxy='${https_proxy}'
 endif
 
-.PHONY: clean all build
+.PHONY: clean all build test
 
 all: test build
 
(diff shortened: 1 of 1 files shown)


rambohe-ch (approved):

rambohe-ch:
> It seems that we can still merge pr even codecov failed. Is it what we expect? @rambohe-ch

@Congrool yes, I think we can decide to merge pull request manually because codecov data maybe influenced by doc modification on some occasions.

Congrool:
@rambohe-ch Sounds good, but this failure may be confusing for contributors. We'd better make sure that codecov can succeed in most cases.

As described in #922, we can add threshold to codecov to tolerate some small coverage rate jitter.

>codecov data maybe influenced by doc

It seems that doc does not affect the codecov. We can verify it in #921 after #922 being merged.

rambohe-ch:
> @rambohe-ch Sounds good, but this failure may be confusing for contributors. We'd better make sure that codecov can succeed in most cases.
> 
> As described in #922, we can add threshold to codecov to tolerate some small coverage rate jitter.
> 
> > codecov data maybe influenced by doc
> 
> It seems that doc does not affect the codecov. We can verify it in #921 after #922 being merged.

@Congrool yes, codecov is a great tool, it will give the detail test file that decrease the code test coverage.

PR #934, opened 2022-07-28
allow codecov to tolerate coverage rate jitter

Congrool:
#### What type of PR is this?
/kind enhancement

As we can see in #922 , there're some unforseen mistakes in our unit tests, which may result in coverage rate jitter and cause unexpected and unrelative codecov failure when submiting prs. Thus, we'd better tolerate small coverage rate drop to make contribution process more friendly.

Currently, we set the threshold as 2%. And when coverage rate drop exceeds 2%, this pr will received a failure notification from codecov.

Fixes #

```release-note

```

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 2%


rambohe-ch:
@Congrool ok, we can tolerate it now, but we should solve coverage rate jitter finally.

rambohe-ch (approved):

PR #952, opened 2022-08-15
fix: get proper nodename from hostname in yurtadm

luc99hen:
/kind bug
> /kind documentation
> /kind enhancement
> /kind good-first-issue
> /kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage

Currently when joining node with `yurtadm join`, it will get the nodename from `os.hostname()` directly which is problematic when hostname doesn't comply to the k8s object name specification. And this pr make this compatible with k8s.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurtadm/cmd/join/join.go
+++ b/pkg/yurtadm/cmd/join/join.go
@@ -20,7 +20,6 @@ package join
 import (
 	"fmt"
 	"io"
-	"os"
 	"strings"
 
 	"github.com/lithammer/dedent"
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@luc99hen Thank you for posting pull request. please add test case for lines #L159 - L163 in pkg/yurtadm/util/edgenode/util.go file.

rambohe-ch (approved):

PR #960, opened 2022-08-19
fix issue 957

### LINKED ISSUE #957: [BUG] YurtHub cache resource namespace hava error
image

if resource is namespace，in method saveListObject，info.Namespace is nil, so the cachekey is comp/namespace/kube-system, but in method saveOneObject, info.Namespace is kube-system, so the cachekey is comp/namespace/kube-system/kube-system.

if yurtHub unable to connect apiserver, There is an error in getting a resource from the local cache.

slm940208:
Fixes #957

image

/root/go/pkg/mod/k8s.io/apiserver@v0.22.3/pkg/endpoints/request/requestinfo.go:180

This side causes the get and list namespace to return different namespace fields in requestInfo.
But there is a problem with the resources of namespace alone. So the specific resource namespace is circumvented in the code.

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager_test.go
+++ b/pkg/yurthub/cachemanager/cache_manager_test.go
@@ -428,6 +428,37 @@ func TestCacheGetResponse(t *testing.T) {
 			resource:         "nodes",
 			cacheResponseErr: true,
 		},
+		"cache response for get namespace": {
+			group:   "",
+			version: "v1",
+			key:     "kubelet/namespaces/kube-system",
+			inputObj: runtime.Object(&v1.Namespace{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: "v1",
+					Kind:       "Namespace",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:            "kube-system",
+					ResourceVersion: "1",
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@slm940208 please add unit test for your added code.

slm940208:
@rambohe-ch Unit tests have been added

Congrool:
Could you please also add a test case in [TestCacheGetResponse](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_manager_test.go#L61) of cache_manager_test.go ?
It can help us better to understand this case in the context of cacheManager.

slm940208:
@Congrool OK

slm940208:
@Congrool @rambohe-ch Test case in [TestCacheGetResponse](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_manager_test.go#L61) and [TestCacheListResponse](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_manager_test.go#L861) have been added.

Congrool:
/lgtm
Thanks for your contribution

rambohe-ch (approved):

Issue #940, opened 2022-08-04
[BUG] waiting for the master to sign the yurthub certificate


beatyman:
![企业微信截图_16595949747885](https://user-images.githubusercontent.com/16757974/182779671-ac3b2121-a407-4d37-8ed5-276fdd3763f5.png)
![企业微信截图_16595950152438](https://user-images.githubusercontent.com/16757974/182779733-aa813b09-e34a-43d8-ba8d-9dc82ba1ead0.png)
![企业微信截图_16595950323600](https://user-images.githubusercontent.com/16757974/182779777-611226f6-ca3a-4948-a9bb-6df7620fd1fa.png)

Congrool:
Could you please post the version of yurthub and yurt-controller-manager and the log of yurt-controller-manager?

beatyman:
yurt-controller-manager: lastest 
yurthub: : lastest

beatyman:
yurt-controller-manager log:  
![企业微信截图_16596059525447](https://user-images.githubusercontent.com/16757974/182815896-b757d10b-a6eb-46f6-8ff8-fe9ffa7a1712.png)

beatyman:
![企业微信截图_16596059852888](https://user-images.githubusercontent.com/16757974/182816024-cad84d73-263c-43bc-a9af-acb09e097a9d.png)

Congrool:
It seems that images are not latest enough, since we post images to dockerhub every day. It's impossible that the latest images were created 8 months ago.

Also, the tag `latest` is not the official released version. Please use v0.7.0 and try again. 
```
openyurt/yurthub:v0.7.0
openyurt/yurt-controller-manager:v0.7.0
```

Congrool:
And I guess the error occurs because of using a newer `yurt-controller-manager.yaml` in which the authorization of approving `kubernetes.io/legacy-unknown` has been removed about 7 months ago in pr #809.

rambohe-ch:
@beatyman Thank you for raising issue. as @Congrool reply, if you can update your local yurt-app-manager/yurthub image, the problem will be solved.

PR #962, opened 2022-08-24
keep the images of the latest version of each openyurt component consistent

windydayc:
/kind bug

Ensure that the images of the latest version of each openyurt component pulled are always consistent.

Since we post images to dockerhub every day, in the process of deploying openyurt, we often encounter the problem that the 
 latest versions of these openyurt components are inconsistent, such as issue #940. The main reason may be that the imagePullPolicy of some components is `Always`, where others are `IfNotPresent`.

### DIFF
--- a/config/setup/yurt-controller-manager.yaml
+++ b/config/setup/yurt-controller-manager.yaml
@@ -130,7 +130,7 @@ spec:
       - operator: "Exists"
       affinity:
         nodeAffinity:
-          # we prefer allocating ecm on cloud node
+          # we prefer allocating ycm on cloud node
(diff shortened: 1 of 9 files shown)


rambohe-ch:
@windydayc Thank you for posting pull request. It's not a good idea to change the `imagePullPolicy` from `IfNotPresent` to `Always` because edge nodes can not pull images successfully if we reboot edge nodes when cloud-edge network is off. so `imagePullPolicy` is `Always`, the pods on edge nodes may not heal when restartup.

rambohe-ch:
@windydayc and i will close this pull request manually because we should keep `imagePullPolicy` as `IfNotPresent`.

PR #1002, opened 2022-09-14
gc error process

huiwq1990:
Signed-off-by: huiwq1990 <huiwq1990@163.com>

/kind enhancement
> /kind good-first-issue
> /kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/gc/gc.go
+++ b/pkg/yurthub/gc/gc.go
@@ -62,7 +62,7 @@ func NewGCManager(cfg *config.YurtHubConfiguration, restConfigManager *rest.Rest
 		eventsGCFrequency: time.Duration(gcFrequency) * time.Minute,
 		stopCh:            stopCh,
 	}
-	_ = mgr.gcPodsWhenRestart()
+	mgr.gcPodsWhenRestart()
 	return mgr, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/yurthub/gc/gc.go
```diff
@@ -99,7 +100,7 @@ func (m *GCManager) gcPodsWhenRestart() error {
 	cfg := m.restConfigManager.GetRestConfig(true)
 	if cfg == nil {
 		klog.Errorf("could not get rest config, so skip gc pods when restart")
-		return err
+		return errors.New("could not get rest config, so skip gc pods when restart")
```
Congrool, line 103: It seems that we do not care about error returned from `gcPodsWhenRestart`. Could we modify this function as
```go
func (m *GCManager) gcPodsWhenRestart()
```

rambohe-ch: +1

rambohe-ch: > It seems that we do not care about error returned from `gcPodsWhenRestart`. Could we modify this function as
> 
> ```go
> func (m *GCManager) gcPodsWhenRestart()
> ```

agree +1

pkg/yurthub/gc/gc.go
```diff
@@ -89,29 +89,29 @@ func (m *GCManager) Run() {
 	}, m.eventsGCFrequency, 2, true, m.stopCh)
 }
 
-func (m *GCManager) gcPodsWhenRestart() error {
+func (m *GCManager) gcPodsWhenRestart() {
 	localPodKeys, err := m.store.ListKeys("kubelet/pods")
 	if err != nil || len(localPodKeys) == 0 {
-		return nil
```
Congrool, line 95: Could we add a log here, since we do not gc in this path.

rambohe-ch (approved):

PR #1004, opened 2022-09-14
add ota update RESTful API

xavier-hou:
add ota pod update REST API
1. `GET /openyurt.io/v1/pods`
2. `POST /openyurt.io/v1/namespaces/{ns}/pods/{podname}/update`

Related proposal #921 

```release-note

```


pkg/yurthub/server/server.go
```diff
@@ -170,6 +171,10 @@ func registerHandlers(c *mux.Router, cfg *config.YurtHubConfiguration, certifica
 
 	// register handler for metrics
 	c.Handle("/metrics", promhttp.Handler())
+
+	// register handler for ota upgrade
+	c.Handle("/openyurt.io/v1/pods", ota.GetPods(clientset)).Methods("GET")
+	c.Handle("/openyurt.io/v1/namespaces/{ns}/pods/{podname}/update", ota.UpdatePod(clientset)).Methods("POST")
```
rambohe-ch, line 177: according to https://github.com/openyurtio/openyurt/pull/921 proposal, the path should be the following:
```
/openyurt.io/v1/pods --> /pods
/openyurt.io/v1/namespaces/{ns}/pods/{podname}/update --> /openyurt.io/v1/namespaces/{ns}/pods/{podname}/upgrade
```

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,177 @@
...
+	// TODO: whether need to display
+	// OldImgs    []string
+	// NewImgs    []string
+	Updatable bool
+}
+
+// GetPods return all daemonset pods' update information by PodStatus
+func GetPods(clientset kubernetes.Interface) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		podsStatus, err := getPodUpdateStatus(clientset)
```
rambohe-ch, line 48: we need to return `PodList` for `/pods` request, not to define a new struct `PodsStatus`.

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,177 @@
...
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+	})
+}
+
+// getDaemonsetPodUpdateStatus check pods annotation "apps.openyurt.io/pod-updatable"
+// to determine whether new version application is availabel
+func getPodUpdateStatus(clientset kubernetes.Interface) ([]*PodStatus, error) {
+	pods, err := clientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{})
```
rambohe-ch, line 97: only pods on this node should be returned. not all of pods in the cluster.

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	podsStatus := make([]*PodStatus, 0)
+	for _, pod := range pods.Items {
+		var updatable bool
+		v, ok := pod.Annotations[PodUpdatableAnnotation]
```
rambohe-ch, line 105: we should not use `pod.Annotations[PodUpdatableAnnotation]` to verify that pods need to upgrade or not. i think we can add a new status condition  in `Pod.Status.Conditions`. and the new status condition as following:
```
condition type: PodNeedUpdate
condition status: True/False
```

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// applyUpdate execute pod update process by deleting pod under OnDelete update strategy
+func applyUpdate(clientset kubernetes.Interface, namespace, podName string) (error, bool) {
+	pod, err := clientset.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Get pod %v/%v failed, %v", namespace, podName, err)
+		return err, false
+	}
+
```
rambohe-ch, line 137: please add a if condition when pod is terminating, we should return directly.

pkg/yurthub/otaupdate/ota.go
xavier-hou: Should we do this by permission control? 
Maybe we need add a check in `/openyurt.io/v1/namespaces/{ns}/pods/{podname}/upgrade` as well?

rambohe-ch: > Should we do this by permission control?

yes, we can only upgrade pods that run on the same node.

zyjhtangtang:
Are pods allowed to be updated when the cloud edge is disconnected?

rambohe-ch:
@xavier-hou please fix unit tests error

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,148 @@
...
+			klog.Errorf("Marshal pod list failed: %v", err.Error())
+			returnErr(fmt.Errorf("Get pod list failed: data transfer to json format failed."), w, http.StatusInternalServerError)
+			return
+		}
+
+		w.WriteHeader(http.StatusOK)
+		n, err := w.Write(data)
+		if err != nil || n != len(data) {
+			klog.Errorf("Write resp for request, expect %d bytes but write %d bytes with error, %v", len(data), n, err)
+		}
```
rambohe-ch, line 63: please use `WriteObject()` to return response, you can dive the code here:
https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/util/util.go#L180-L210

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,148 @@
...
+
+	klog.V(5).Infof("Pod: %v/%v is updatable", namespace, podName)
+	err = clientset.CoreV1().Pods(namespace).Delete(context.TODO(), podName, metav1.DeleteOptions{})
+	if err != nil {
+		klog.Errorf("Update pod %v/%v failed when delete, %v", namespace, podName, err)
+		return err, false
+	}
+
+	klog.Infof("Update pod: %v/%v success", namespace, podName)
+	return nil, true
```
rambohe-ch, line 121: I think do we need to wait for the new version pod to be created and then return?

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,148 @@
...
+		err, ok := applyUpdate(clientset, namespace, podName, nodeName)
+		if err != nil {
+			returnErr(fmt.Errorf("Apply update failed"), w, http.StatusInternalServerError)
+			return
+		}
+		if !ok {
+			returnErr(fmt.Errorf("Pod is not-updatable"), w, http.StatusForbidden)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
```
rambohe-ch, line 83: maybe we also need to return the deleted pod beside statusOK.

rambohe-ch: we can reference the API definitions in K8s.

xavier-hou: Logically, the function of this interface is to upgrade applications. Therefore, i think return the updated pod (newly created) is better. 

This also corresponds to your last comment.
> I think do we need to wait for the new version pod to be created and then return?

rambohe-ch: It's not a good idea to wait for the new pod created, maybe we can add a new API for querying the status of upgrade.

pkg/yurthub/otaupdate/ota.go
```diff
@@ -0,0 +1,177 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Pre-check if edge yurthub node is connected to the cloud
+		if !isEdgeCloudConnected(checker, servers) {
+			klog.Errorf("Get pod list is not allowed when edge is disconnected to cloud")
+			WriteErr(w, "Get pod list is not allowed when edge is disconnected to cloud", http.StatusForbidden)
+			return
+		}
+
+		podList, err := clientset.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
+			FieldSelector: "spec.nodeName=" + nodeName,
```
rambohe-ch, line 50: "spec.nodeName=" + nodeName --> fmt.Sprintf("spec.nodeName=%s", nodeName)

pkg/yurthub/server/server.go
```diff
@@ -60,15 +62,16 @@ type yurtHubServer struct {
 func NewYurtHubServer(cfg *config.YurtHubConfiguration,
 	certificateMgr interfaces.YurtCertificateManager,
 	proxyHandler http.Handler,
-	rest *rest.RestConfigManager) (Server, error) {
+	rest *rest.RestConfigManager,
+	checker healthchecker.HealthChecker) (Server, error) {
 	hubMux := mux.NewRouter()
-	registerHandlers(hubMux, cfg, certificateMgr)
 	restCfg := rest.GetRestConfig(false)
 	clientSet, err := kubernetes.NewForConfig(restCfg)
```
rambohe-ch, line 69: clientSet is created with remoteServers[0], so healthy status of all remoteServers will be checked when ota upgrade is executed. so `isEdgeCloudConnected(checker, servers)` can return true, but clientSet maybe fail.

xavier-hou: @rambohe-ch  how about using `rest.GetRestConfig(true)` to generate a new rest config to solve this problem?

rambohe-ch: @xavier-hou we can not use `rest.GetRestConfig(true)` to fix this problem, because the remote server is healthy when creating clientSet, but the remote server maybe become unhealthy when upgrading pod on the edge.

rambohe-ch: It looks like we need to create a new ClientSet with healthy server when upgrading pod.

PR #1007, opened 2022-09-15
update valid openyurt version of yurtctl test

Congrool:
#### What type of PR is this?
/kind bug

Since we've released v0.7.1 and v1.0.0, we should update valid openyurt version of yurtctl test.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurtctl/cmd/yurttest/kindinit/init.go
+++ b/pkg/yurtctl/cmd/yurttest/kindinit/init.go
@@ -57,8 +57,9 @@ var (
 		"v0.5.0",
 		"v0.6.0",
 		"v0.6.1",
-		"v0.6.2",
 		"v0.7.0",
+		"v0.7.1",
+		"v1.0.0",
 		"latest",
 	}
 	validKindVersions = []string{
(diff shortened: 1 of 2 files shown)


rambohe-ch:
@Congrool It looks like that this pull request will not cause unit tests error. and i have re-run the unit tests job but failed again. please take a look for the unit test errors.

Congrool:
Well, I found the reason of why unit-test failed. In `TestKindOptions_Validate`, we have
```go
cases1 := []struct {
		nodeNum           int
		kubernetesVersion string
		openyurtVersion   string
		ignoreErr         bool
		want              string
	}{
		...
		{
			3,
			"v1.22",
			"v0.1.0",
			false,
			"v0.1.0 is not a valid openyurt version, all valid versions are v0.5.0,v0.6.0,v0.6.1,v0.6.2,v0.7.0,latest. If you know what you're doing, you can set --ignore-error",
		},
	}
```

Here, what we want couple with valid versions. We have to decouple with it.

rambohe-ch (approved):

PR #1009, opened 2022-09-16
recover cache error process

huiwq1990:
/kind bug
> /kind documentation
> /kind enhancement
> /kind good-first-issue
> /kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/storage/disk/storage.go
+++ b/pkg/yurthub/storage/disk/storage.go
@@ -57,9 +57,9 @@ func NewDiskStorage(dir string) (storage.Store, error) {
 		baseDir:          dir,
 	}
 
-	err := ds.Recover("")
-	if err != nil {
+	if err := ds.Recover(""); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/yurthub/storage/disk/storage.go
```diff
@@ -57,9 +57,9 @@ func NewDiskStorage(dir string) (storage.Store, error) {
 		baseDir:          dir,
 	}
 
-	err := ds.Recover("")
-	if err != nil {
+	if err := ds.Recover(""); err != nil {
 		klog.Errorf("could not recover local storage, %v, and skip the error", err)
+		return nil, err
```
rambohe-ch, line 62: as error info said: `and skip the error`, we should only dump the error info instead of exiting yurthub, because of some temp file can not recover should not prevent starting up yurthub.

PR #1025, opened 2022-10-09
improve data filter framework

rambohe-ch:
/kind enhancement

#### What this PR does:
improve data filter setting format in configmap `kube-system/yurthub-cfg` in order to leverage end users configure filter setting conveniently.

take servicetopology filter as an example:
- original setting
filter_servicetopology: "user-agent1/resource#verb1;verb2, user-agent2/resource#verb1;verb2"

- improved setting
servicetopology: "user-agent1, user-agent2"

by the way, the original setting format is also supported in order to keep consistency.

#### Why we need it:
because which resource and verbs of requests are supported is fixed by filter, for example, `servicetopology` filter can only support requests witch list/watch endpoints and endpointslices. so it's no need for end users to configure info of resource and verbs.

Fixes #

```release-note
Components on the edge nodes that want to use data filter framework, the way of configuring filter setting in configmap `kube-system/yurthub-cfg` is changed, the detail info is explained above.
```


pkg/yurthub/filter/manager/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	if options.AccessServerThroughHub {
+		mutatedMasterServicePort = options.YurtHubProxySecurePort
+		if options.EnableDummyIf {
+			mutatedMasterServiceHost = options.HubAgentDummyIfIP
+		} else {
+			mutatedMasterServiceHost = options.YurtHubProxyHost
+		}
+	}
+
+	runnerChain, err := createRunnerChain(filters, sharedFactory, yurtSharedFactory, serializerManager, storageWrapper, util.WorkingMode(options.WorkingMode), options.NodeName, mutatedMasterServiceHost, mutatedMasterServicePort)
```
Congrool, line 69: How about calling it `createFilterRunners` instead of `createRunnerChain`, as these returned runners will be called on demand rather than in a specific order indicated by this chain. It's a little confused when reading this code.

pkg/yurthub/filter/approver.go
```diff
@@ -36,38 +36,35 @@ import (
...
-		reqKey("nginx-ingress-controller", "endpoints", "watch"): ServiceTopologyFilterName,
-		reqKey("kube-proxy", "endpointslices", "list"):           ServiceTopologyFilterName,
-		reqKey("kube-proxy", "endpointslices", "watch"):          ServiceTopologyFilterName,
-		reqKey("coredns", "endpoints", "list"):                   ServiceTopologyFilterName,
-		reqKey("coredns", "endpoints", "watch"):                  ServiceTopologyFilterName,
-		reqKey("coredns", "endpointslices", "list"):              ServiceTopologyFilterName,
-		reqKey("coredns", "endpointslices", "watch"):             ServiceTopologyFilterName,
-	}
+	// defaultBlackListRequests is used for determining request to pass through filters.
+	defaultBlackListRequests = sets.NewString(reqKey(projectinfo.GetHubName(), "configmaps", "list"), reqKey(projectinfo.GetHubName(), "configmaps", "watch"))
```
Congrool, line 48: Does it mean that items in the `BlackList` will not be filtered? This comment seems to be ambiguous.

Congrool:
some little revisions are needed, otherwise LGTM

pkg/yurthub/filter/manager/manager.go
rambohe-ch: it's a good idea. fixed

pkg/yurthub/filter/approver.go
rambohe-ch: comments is improved.

PR #1026, opened 2022-10-10
support use multiple service topology setting at one time for service topology feature .

### LINKED ISSUE #1020: [feature request]service topology feature supports multiple topology configuration.
According to the tutorial: https://openyurt.io/docs/user-manuals/network/service-topology, service topology feature can only used for node or nodePool. in an other word, user can only select only one topology type.

But in the cases, users want to configure multiple topology types for service, for example: `openyurt.io/topologyKeys: kubernetes.io/hostname, kubernetes.io/zone`, so traffic will be forwarded to the pods on the same node, and traffic will be forwarded to the pods on the same nodePool when pods on the same nodes are not ready.

**Why is this needed**:
If service topology can be configured with multiple topology types, system stability can be improved because traffic will be handled in more bigger scope.

**others**
/kind feature

rambohe-ch:
/kind feature

As described in issue: https://github.com/openyurtio/openyurt/issues/1020, end user can configure `Service` with multiple topology setting(`openyurt.io/topologyKeys: kubernetes.io/hostname, kubernetes.io/zone, *`), so the traffic that access this service will be forwarded to endpoints on the same local node firstly(this means first setting `kubernetes.io/hostname` is considered). If there are no endpoints of service on the same node, the second setting `kubernetes.io/zone` will be considered(this means traffic will be forwarded to endpoints on the same node pool). and so forth for the last setting "*".

Fixes #1020 

```release-note
Multiple service topology setting can be configured for the service. Traffic is forwarded to the backend of the corresponding node according to the priority.(kubernetes.io/hostname > kubernetes.io/zone, openyurt.io/nodePool > *)
```

### DIFF
--- a/pkg/yurthub/filter/servicetopology/handler.go
+++ b/pkg/yurthub/filter/servicetopology/handler.go
@@ -18,6 +18,7 @@ package servicetopology
 
 import (
 	"io"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
@@ -42,7 +43,7 @@ const (
 )
 
 var (
-	topologyValueSets = sets.NewString(AnnotationServiceTopologyValueNode, AnnotationServiceTopologyValueZone, AnnotationServiceTopologyValueNodePool)
+	topologyValueSets = sets.NewString(AnnotationServiceTopologyValueNode, AnnotationServiceTopologyValueZone, AnnotationServiceTopologyValueNodePool, "*")
 )
 
 type serviceTopologyFilterHandler struct {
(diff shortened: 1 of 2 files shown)


rambohe-ch:
@LinFCai @zzguang PTAL

rambohe-ch:
@zzguang Thanks for noticing from @LinFCai, it's not easy for servicetopology filter to support multiple topologies, because there are multiple endpointslice objects for one service, but servicetopology filter in Yurthub can handle endpointslice object one by one, so if some endpointslice maybe satisfy `Hostname topology`, but the other endpointslice maybe satisfy `NodePool topology`, then traffic will be forwarded to endpoints in `NodePool topology` instead of `Hostname topology`.
so i will add `WIP` for this pull request temporarily until i can find solution.

PR #1027, opened 2022-10-11
add proposal to unify cloud edge comms solution

zzguang:
Current OpenYurt provides 2 independent solutions in cloud edge comms domain, which are Raven and YurtTunnel, this proposal aims to integrate YurtTunnel into Raven, and provide an unified cloud edge comms solution to users.

Signed-off-by: zzguang <zhengguang.zhang@intel.com>

Fixes #

```release-note

```


rambohe-ch:
@zzguang please fix misspell errors.

rambohe-ch:
@zzguang Thanks for posting pull request about integrating yurttunnel into raven. and i agree with you that yurttunnel and raven are independent solutions in cloud edge comms domain. but yurttunnel is a 7 layer solution for forwarding http/https requests from cloud to edge, and raven is a 3 layer solution for forwarding requests between cloud and edge, or edge to edge. 

DevOps traffic and business data traffic can use these two solutions simultaneously in terms of their requirements. for example, DevOps traffic can use both yurttunnel and raven.

I think that the solution 4 is so complicated because end user should take care 4 components. how about integrate raven-gateway-agent and raven-tunnel-agent into one component?

zzguang:
> @zzguang Thanks for posting pull request about integrating yurttunnel into raven. and i agree with you that yurttunnel and raven are independent solutions in cloud edge comms domain. but yurttunnel is a 7 layer solution for forwarding http/https requests from cloud to edge, and raven is a 3 layer solution for forwarding requests between cloud and edge, or edge to edge.
> 
> DevOps traffic and business data traffic can use these two solutions simultaneously in terms of their requirements. for example, DevOps traffic can use both yurttunnel and raven.
> 
> I think that the solution 4 is so complicated because end user should take care 4 components. how about integrate raven-gateway-agent and raven-tunnel-agent into one component?

Thanks for your comments, IMHO, since we can not merge the layer 7 yurttunnel completely into layer 3 raven, we should decouple them clearly for users. If we fuse them on edge side while keep them separated on cloud side, it may lead to confusions to users when they only want to adopt one of them.

rambohe-ch (approved):

PR #1028, opened 2022-10-11
bugfix: even no endpoints left after filter, an empty object should be returned to clients

rambohe-ch:
/kind bug

if no endpoints left after service topology filtering, yurthub should return a empty endpointslice to clients(like kube-proxy) instead of skipping the object. because kube-proxy need to use the empty endpointslice to update network setting on edge nodes.

Fixes #

```release-note

```


rambohe-ch:
@zzguang @yingjianjian PTAL

Congrool (commented):
/lgtm

PR #1029, opened 2022-10-13
do not set host service topology in test cluster

Congrool:
#### What type of PR is this?
/kind bug

When we use test cluster with more than 2 nodes, there may be no pods running on the control-plane-node, which will make the e2e tests of yurt-tunnel fail. And I think, there seems no reason to set host service topology in the test environment.

Fixes #

```release-note

```


Congrool:
/cc @lorrielau 
FYI, we can run autonomy test based on this pr.

rambohe-ch (approved):

PR #1040, opened 2022-10-29
add poolcoordinator controller and webhooks to replace nodelifecycle …

### LINKED ISSUE #776: [feature request]`NodePool Governance Capability`: remove nodelifecycle controller and add pool-coordinator controller in yurt-controller-manager component.

  We will remove the original NodeLifeCycle controller in Yurt-Controller-Manager component, so native Kube-Controller-Manager doesn't need to disable NodeLifeCycle controller when startup. at the same time, we will add a new controller and webhook for providing edge autonomy feature in Yurt-Controller-Manager component.

  - provide two edge autonomy mode for workload, node autonomy and nodepool autonomy.
    * node autonomy: the pod will not be evicted from node even if the node crashed. if end user want to recreate or migrate the pod,  end user should delete and recreate the pod manually.
    * nodepool autonomy: the pod will not be evicted from node when only cloud-edge network disconnected, and the pod will be evicted to other normal node when edge node is un-normal(like node crashed)

 - add a new annotation(apps.openyurt.io/autonomy) for workload to select the appropriate autonomy mode for its needs.

  This issue is used for tracking the newly added controller(named as pool-coordinator controller), and the controller will provide the following feature:

  - when the node lease is delegated to report continuously for more then 40s(4 time heartbeat interval), the node should be recognized that cloud-edge network disconnected, but the pods on node should be kept. so add a un-schedulable taint to the node, so newly created pods will not be scheduled to this node.

 **Notes**: For details, please refer to the proposal: https://github.com/openyurtio/openyurt/pull/772

**Why is this needed**:
As mentioned in the proposal(https://github.com/openyurtio/openyurt/pull/772) "NodePool Autonomy", in the node pool, all yurthubs connected to the cloud will elect a leader, and the leader yurthub will act as a heartbeat proxy to report the heartbeat of disconnected nodes to the cloud. The Yurt-Controller-Manager judges that the logic of the node has changed, and it needs to consider the two situations of node autonomy and NodePool autonomy.

**others**
/kind feature

LindaYu17:
…controller

/kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage

replace with nodelifecycle controller from k8s with pool-coordinator's pod management.

Fixes #776

```release-note

```


luc99hen:
It seems that this controller follows the code style of kubebuilder. If this is fine, I will build my cert management part based on this framework. @rambohe-ch

rambohe-ch:
> It seems that this controller follows the code style of kubebuilder. If this is fine, I will build my cert management part based on this framework. @rambohe-ch

@LindaYu17 @luc99hen I think that newly added controllers should follow the code style of `kube-controller-manager` instead of kubebuilder.

pkg/controller/poolcoordinator/client/client.go
```diff
@@ -0,0 +1,41 @@
+package client
```
rambohe-ch, line 1: please add license header for newly added file.

pkg/controller/poolcoordinator/webhook/pod_admission.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+type PodAdmission struct {
+	request *admissionv1.AdmissionRequest
+	pod     *corev1.Pod
+	node    *corev1.Node
+}
+
+// extracts pod from admission request
+func (pv *PodAdmission) getPod() error {
```
rambohe-ch, line 57: pv --> pa?

pkg/controller/poolcoordinator/utils/nodepool.go
```diff
@@ -0,0 +1,144 @@
...
+	return false
+}
+
+func NodeIsAlive(leaseLister leaselisterv1.LeaseNamespaceLister, nodeName string) bool {
+	lease, err := leaseLister.Get(nodeName)
+	if err != nil {
+		klog.Error(err)
+		return false
+	}
+	diff := time.Now().Sub(lease.GetCreationTimestamp().Time)
```
rambohe-ch, line 120: we should use `lease.Spec.RenewTime` instead of `GetCreationTimestamp()` to specify the last heartbeat time

pkg/controller/poolcoordinator/webhook/pod_admission.go
```diff
@@ -0,0 +1,409 @@
...
+		return reviewResponse(pv.request.UID, true, http.StatusAccepted, reason), nil
+	}
+
+	err := pv.getPod()
+	if err != nil {
+		e := fmt.Sprintf("could not parse pod in admission review request: %v", err)
+		return reviewResponse(pv.request.UID, false, http.StatusBadRequest, e), err
+	}
+
+	if pv.pod.Annotations == nil || pv.pod.Annotations[constant.PodAvailableAnnotation] != "true" {
```
rambohe-ch, line 201: toleration should also be added to pods for node with `autonomy` annotation.

pkg/controller/poolcoordinator/utils/nodepool.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// return count of active node
+func (ns *NodeSet) Nodes() []string {
+	return ns.nodes
+}
+
+func NewNodepoolMap() *NodepoolMap {
+	return &NodepoolMap{
+		nodepools: make(map[string]*NodeSet),
```
rambohe-ch, line 56: how about use `sets.String` instead of defining `NodeSet`?

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,258 @@
...
+package poolcoordinator
+
+import (
+	"context"
+	"sync"
+
+	"github.com/openyurtio/openyurt/pkg/controller/poolcoordinator/constant"
+	"github.com/openyurtio/openyurt/pkg/controller/poolcoordinator/lister"
+	"github.com/openyurtio/openyurt/pkg/controller/poolcoordinator/utils"
+	"github.com/openyurtio/openyurt/pkg/controller/poolcoordinator/webhook"
```
rambohe-ch, line 27: lib of openyurt should be put at the last and independent section.

pkg/controller/poolcoordinator/lister/lister.go
```diff
@@ -0,0 +1,64 @@
...
+		AddFunc:    afunc,
+		UpdateFunc: ufunc,
+		DeleteFunc: dfunc,
+	})
+}
+
+func CreateNodeInformer(client client.Interface, factory informers.SharedInformerFactory, stopper chan (struct{})) v1.NodeInformer {
+	nodeInformer := factory.Core().V1().Nodes()
+	factory.Start(stopper)
+	factory.WaitForCacheSync(stopper)
```
rambohe-ch, line 48: we can not start informer at here, because all of informers will be started at https://github.com/openyurtio/openyurt/blob/master/cmd/yurt-controller-manager/app/controllermanager.go#L208

rambohe-ch: and `factory.WaitForCacheSync(stopper)` should be executed in a independent goroutine, you can take YurtCSRApprover controller as a example: https://github.com/openyurtio/openyurt/blob/master/pkg/controller/certificates/csrapprover.go#L181

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,270 @@
...
+	nc.nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    onNodeCreate,
+		UpdateFunc: onNodeUpdate,
+		DeleteFunc: onNodeDelete,
+	})
+	nc.leaseInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    onLeaseCreate,
+		UpdateFunc: onLeaseUpdate,
+		DeleteFunc: nil,
+	})
```
luc99hen, line 265: How about moving these configurations to the `NewController` func

LindaYu17: updated

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,378 @@
...
+	add          bool
+}
+
+type LeaseDelegatedCounter struct {
+	v    map[string]int
+	lock sync.Mutex
+}
+
+var (
+	ctl *Controller
```
rambohe-ch, line 74: please do not define Controller var.

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,378 @@
...
+
+	ldc *LeaseDelegatedCounter
```
rambohe-ch, line 76: ditto

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,378 @@
...
+	defer pcc.nodePoolUpdateQueue.ShutDown()
+
+	klog.Info("start node taint worker")
+	go pcc.nodeWorker()
+
+	klog.Info("start nodepool maintenance worker")
+	go pcc.nodePoolWorker()
+
+	klog.Info("create webhook")
+	go webhook.Run(pcc.nodeLister, pcc.leaseLister, pcc.nodepoolMap)
```
rambohe-ch, line 375: maybe it's hard to understand that run webhook in the controller, how about create an independent webhook server for all webhooks in yurt-controller-manager? so it's easy to add other webhooks in the new server in the future.

pkg/controller/poolcoordinator/poolcoordinator_controller.go
```diff
@@ -0,0 +1,378 @@
...
+
+	item := nodeTaintItem{
+		nodeName: nl.Name,
+		taint:    false,
+	}
+	if nok && nval == "true" {
+		ldc.Inc(nl.Name)
+		if ldc.Counter(nl.Name) >= constant.LeaseDelegationThreshold {
+			item.taint = true
+			GetController().nodeUpdateQueue.Add(item)
```
rambohe-ch, line 152: in order to improve the readability of code, maybe we can add node name instead nodeTaintItem  into nodeUpdateQueue.

pkg/controller/poolcoordinator/webhook/pod_admission.go
```diff
@@ -0,0 +1,422 @@
...
+)
+
+var (
+	ValidatePath string = "/pool-coordinator-webhook-validate"
+	MutatePath   string = "/pool-coordinator-webhook-mutate"
+	HealthPath   string = "/pool-coordinator-webhook-health"
+
+	nodeLister  listerv1.NodeLister
+	leaseLister leaselisterv1.LeaseNamespaceLister
+	nodepoolMap *utils.NodepoolMap
```
rambohe-ch, line 58: we should try best not to define file-scope var.

PR #1044, opened 2022-11-01
bugfix: non resource handle miss for coredns

rambohe-ch:
/kind bug

Like coredns component will post non resource request to kube-apiserver through yurthub, but yurthub can not handle non resource request when cloud-edge line off, so coredns can not recover when cloud-edge disconnected.

this pull request is used for solving the above problem.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/certificate/hubself/fake_cert_mgr.go
+++ b/pkg/yurthub/certificate/hubself/fake_cert_mgr.go
@@ -58,7 +58,7 @@ e35EltCRCjoejRHTuN9TC0uCoVipAiAXaJIx/Q47vGwiw6Y8KXsNU6y54gTbOSxX
 -----END RSA PRIVATE KEY-----`
 )
 
-type fakeYurtHubCertManager struct {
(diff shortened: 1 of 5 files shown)


rambohe-ch:
@Congrool @Sodawyx PTAL

Congrool:
What's the reason that cause CoreDNS unable to restart when cloud-edge disconnect? 

I can find two main changes in this pr:
1. wrap https proxy
2. add `Content-Type: application/json` in header

Are these all we need to enable the CoreDNS to restart when offline?

rambohe-ch:
> What's the reason that cause CoreDNS unable to restart when cloud-edge disconnect?
> 
> I can find two main changes in this pr:
> 
> 1. wrap https proxy
> 2. add `Content-Type: application/json` in header
> 
> Are these all we need to enable the CoreDNS to restart when offline?

@Congrool Yes, the main reason is that https proxy is not wrapped by non resource handler.

Congrool (commented):
/lgtm

PR #1048, opened 2022-11-08
proposal of yurtadm join refactoring

### LINKED ISSUE #889: [feature request]optimize `yurtadm join` command
 Code of `kubeadm join` are imported into OpenYurt by `yurtadm join` command, how about import `kubeadm` binary file instead of importing code? 
 if `kubeadm` file does not exist, we can install it online. then we need to make sure that kubelet can access kube-apiserver through yurthub and we do not need to restart kubelet and other pods.

**Why is this needed**:
Compare to import code, using kubeadm binary directly has the following advantages:
1. It's very easy to follow K8s to update kubeadm version, do not need to maintain the code of `kubeadm join` in OpenYurt.
2. reduce the complexity of `yurtadm join`

**others**
/kind feature

YTGhost:
Signed-off-by: HIHIA <283304489@qq.com>

Fixes #889 

```release-note

```

### DIFF
--- a/docs/proposals/20221108-yurtadm-join-refactoring.md
+++ b/docs/proposals/20221108-yurtadm-join-refactoring.md
@@ -1,0 +1,236 @@
+---
+title: Yurtadm Join Refactoring
+authors:
+  - "@YTGhost"
+reviewers:
+  - "@rambohe-ch"
+creation-date: 2022-11-08
+last-updated: 2022-11-08
+status: provisional
+---
+
+# Yurtadm Join Refactoring
+
+<!-- END Remove before PR -->
+
+## Table of Contents
+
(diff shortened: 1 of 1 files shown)


docs/proposals/20221108-yurtadm-join-refactoring.md
```diff
@@ -0,0 +1,234 @@
...
+    server: https://101.34.232.45:6443
```
rambohe-ch, line 222: `101.34.232.45:6443` should be replaced with the public address of kube-apiserver.

YTGhost: Done

rambohe-ch (approved):

PR #1056, opened 2022-11-11
add timeout config in yurthub to handle those watch requests without …

AndyEWang:
#### What type of PR is this?
/kind enhancement

When yurthub takes charge of apiserver and if there is no timeoutSeconds parameter in watch requests, yurthub returns immediately. This causes the requester to retry watch request frantically. Also both yurthub and requester begin rolling logs.
So to add a new startup option for yurthub to define the default watch timeout and also give a default value(1800s like apiserver) to this option.

```release-note
A new optional arg "--min-request-timeout" is added in yurthub startup command. If not set, the default value is 1800s.
```

### DIFF
--- a/pkg/yurthub/proxy/local/local.go
+++ b/pkg/yurthub/proxy/local/local.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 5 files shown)


Congrool:
I also notice this problem, and my case is calico that results in watch request flood. It's reasonable to add timeout for watch request when offline. 

BTW, is there any reason for the value 1800s?

AndyEWang:
> is there any reason for the value 1800s?

just refer to https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/ --min-request-timeout int     Default: 1800

pkg/yurthub/proxy/local/local.go
```diff
@@ -173,8 +176,9 @@ func (lp *LocalProxy) localWatch(w http.ResponseWriter, req *http.Request) error
 	timeout := time.Duration(0)
 	if opts.TimeoutSeconds != nil {
 		timeout = time.Duration(*opts.TimeoutSeconds) * time.Second
-	} else {
-		return nil
+	}
+	if timeout == 0 && lp.minRequestTimeout > 0 {
+		timeout = time.Duration(float64(lp.minRequestTimeout) * (rand.Float64() + 1.0))
```
Congrool, line 181: Do we set seed before calling rand?

Congrool:
Another question here, what's the behaviour of APIServer when handling the watch request with timeoutSeconds parameter?

For example, if we set `--min-request-timeout=1800` for APIServer, and the coming watch request is `/api/v1/nodes?watch=true&timeoutSeconds=32`, how much time before APIServer timing it out?

pkg/yurthub/proxy/local/local.go
AndyEWang: From yurthub src code, cmd/yurthub/yurthub.go already set seed.
`func main() {
	rand.Seed(time.Now().UnixNano())`

AndyEWang:
apiserver respects the timeoutSeconds parameter and only use --min-request-timeout when timeoutSeconds is zero or not passed.

pkg/yurthub/proxy/local/local.go
Congrool: Thanks

rambohe-ch (approved):

PR #1066, opened 2022-11-18
bugfix: StreamResponseFilter of data filter framework can't work if size of one object is over 32KB

rambohe-ch:
/kind bug

1. bugfix: StreamResponseFilter of data filter framework can't work if size of one object is over 32KB
2. improve data filter framework
  - define ObjectHandler interface for filtering Runtime.Object, and other common routines are removed to FilterReadCloser
  - add SerializerManager as common parameter for all filters.

Fixes #

```release-note

```


pkg/yurthub/filter/filter.go
```diff
@@ -115,89 +116,151 @@ func (fis FilterInitializers) Initialize(ins Runner) error {
...
+	return frc.rc.Close()
+}
+
+func (frc *filterReadCloser) ObjectResponseFilter(rc io.ReadCloser) (*bytes.Buffer, error) {
+	var buf bytes.Buffer
+	_, err := buf.ReadFrom(rc)
+	if err != nil {
+		return &buf, err
+	}
+	listObj, err := frc.serializer.Decode(buf.Bytes())
```
Congrool, line 208: It seems that we do not only handle listobject but also standalone object. So, it maybe unsuitable to use the variable name as `listObj`.

pkg/yurthub/filter/filter.go
```diff
@@ -115,89 +116,151 @@ func (fis FilterInitializers) Initialize(ins Runner) error {
...
+
+	newData, err := frc.serializer.Encode(filteredObj)
+	return bytes.NewBuffer(newData), err
+}
 
-			return n, nil
+func (frc *filterReadCloser) StreamResponseFilter(rc io.ReadCloser, ch chan *bytes.Buffer) error {
+	defer func() {
+		close(ch)
+	}()
```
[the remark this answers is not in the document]
rambohe-ch: fixed

pkg/yurthub/filter/filter.go
rambohe-ch: listObj --> obj. 
fixed

rambohe-ch:
@Congrool PTAL

PR #1068, opened 2022-11-20
add ci CIFuzz Github action

DavidKorczynski:
#### What type of PR is this?
/kind feature

Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where OpenYurt already runs. CIFuzz can help detect regressions and catch fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

No

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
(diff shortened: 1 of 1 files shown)


huiwq1990:
@DavidKorczynski Hi, openyurt has some sub projects, like openyurt and yurt-app-manager. Currently, only yurt-app-manager uses oss-fuzz, as show in https://github.com/google/oss-fuzz/blob/master/projects/openyurt/Dockerfile. So I think it's better   make a pr for https://github.com/openyurtio/yurt-app-manager.

PR #1093, opened 2022-12-07
add_reset_test.go

Rudolf1998:

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurtadm/cmd/reset/reset_test.go
+++ b/pkg/yurtadm/cmd/reset/reset_test.go
@@ -1,0 +1,327 @@
+/*
+Copyright 2022 The OpenYurt Authors.
+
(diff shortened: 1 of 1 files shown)


pkg/yurtadm/cmd/reset/reset_test.go
```diff
@@ -0,0 +1,327 @@
+/*
+Copyright 2021 The OpenYurt Authors.
```
Congrool, line 2: Please replace the year as 2022.

Rudolf1998: Thanks for your suggestions，the bug is already fixed

Congrool:
Thanks!
/lgtm

rambohe-ch (approved):

PR #1103, opened 2022-12-20
generate coordinator client and pass through SubjectAccessReview according to secret

### LINKED ISSUE #778: [feature request]`NodePool Governance Capability`: YurtHub supports writing data to pool-spirit

Congrool:
#### What type of PR is this?

/kine bug
/kind feature

The following works has been done in this pr:
1. generate coordinator client according to ca, cert and key in the secret to communicate with coordinator
2. deliver SubjectAccessReview request to coresponding APIServer(Coordinator or Cloud APIServer) according to the Group value in it.
3. add `--enable-pool-coordinator` to enable the yurthub be aware of the pool-coordinator. If disabled, yurthub will work without pool-coordinator and no module about coordinator will start in yurthub.
4. fix bug that cannot cache response when proxying request, currently yurthub can work without `--enable-pool-coordinator`.

Fixes #778 

```release-note

```

### DIFF
--- a/pkg/yurthub/proxy/pool/pool.go
+++ b/pkg/yurthub/proxy/pool/pool.go
@@ -245,12 +245,12 @@ func (pp *PoolCoordinatorProxy) cacheResponse(req *http.Request, resp *http.Resp
 		req = req.WithContext(ctx)
 		wrapPrc, _ := hubutil.NewGZipReaderCloser(resp.Header, resp.Body, req, "cache-manager")
 
-		rc, prc := hubutil.NewDualReadCloser(req, wrapPrc, false)
+		rc, prc := hubutil.NewDualReadCloser(req, wrapPrc, true)
 		go func(req *http.Request, prc io.ReadCloser, stopCh <-chan struct{}) {
 			if err := pp.localCacheMgr.CacheResponse(req, prc, stopCh); err != nil {
 				klog.Errorf("failed to cache req %s in local cache when cluster is unhealthy, %v", hubutil.ReqString(req), err)
 			}
 		}(req, prc, ctx.Done())
-		req.Body = rc
+		resp.Body = rc
 	}
 }
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -96,6 +96,7 @@ type YurtHubConfiguration struct {
 	FilterManager                     *manager.Manager
 	CertIPs                           []net.IP
 	MinRequestTimeout                 time.Duration
+	EnableCoordinator                 bool
 	CoordinatorServerURL              *url.URL
 	CoordinatorStoragePrefix          string
 	CoordinatorStorageAddr            string // ip:port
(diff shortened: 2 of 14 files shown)


Congrool:
Expect e2e-tests to pass

Congrool:
Also do not disable nodelifecycle controller when setting up openyurt-e2e-test cluster with kind. It's the reason why the e2e test failed.

pkg/yurthub/proxy/pool/pool.go
```diff
@@ -243,14 +243,14 @@ func (pp *PoolCoordinatorProxy) cacheResponse(req *http.Request, resp *http.Resp
 	if pp.localCacheMgr.CanCacheFor(req) {
 		ctx := req.Context()
 		req = req.WithContext(ctx)
-		wrapPrc, _ := hubutil.NewGZipReaderCloser(resp.Header, resp.Body, req, "cache-manager")
+		rc, prc := hubutil.NewDualReadCloser(req, resp.Body, true)
+		wrapPrc, _ := hubutil.NewGZipReaderCloser(resp.Header, prc, req, "cache-manager")
```
rambohe-ch, line 247: would you like to explain the reason  of changing the order of gzip and dualreader?

Congrool: It seems no need to change. I'll fix it.

Congrool: Though it has no effect on execution, it seems that it's more readable if we change the order. The 3rd parameter of `NewDualReadCloser` is `isRespBody`. If we use the original code

```go
wrapPrc, _ := hubutil.NewGZipReaderCloser(resp.Header, resp.Body, req, "cache-manager")
rc, prc := hubutil.NewDualReadCloser(req, wrapPrc, false)
```

It does not obviously that if the wrapPrc is response body or not, thus it cause the mistaken 3rd argument as `false` which should be `true` in fact.

While if we use the modified order, we can avoid the mistake. It's obvious that we should set it as `true`.
```go
rc, prc := hubutil.NewDualReadCloser(req, resp.Body, true)
```

rambohe-ch: @Congrool we need to use GZIP reader before dualReader, so all readers in dualReader can get the response data that has been unpacked by GZIP reader.

Congrool: OK, I'll fix it

rambohe-ch (approved):

PR #1104, opened 2022-12-21
openyurt controller framework

kadisi:
#### What type of PR is this?

/kind enhancement

Fixes #

```release-note

```


cmd/yurt-controller/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
```
rambohe-ch, line 1: all Dockerfile are placed under `hack/dockerfiles` dir.

cmd/yurt-controller/OWNERS
```diff
@@ -0,0 +1,9 @@
+# See: https://go.k8s.io/owners
```
rambohe-ch, line 1: i think we need to add OWNERS for every controllers, not need to add OWNERS for yurt-controller-manager component.

rambohe-ch:
@kadisi please remove binary file `cmd/yurt-controller/yurt-controller`

config/yurt-controller/certmanager/certificate.yaml
```diff
@@ -0,0 +1,25 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
```
rambohe-ch, line 1: I think it's not a good idea to import cert manager into OpenYurt to manage the certificates for core components. we need to keep the simple and cleanness for OpenYurt.

config/yurt-controller/crd/bases/apps.openyurt.io_samples.yaml
```diff
@@ -0,0 +1,57 @@
+
```
rambohe-ch, line 1: we should not to maintain yaml files for OpenYurt. please remove yaml files into `charts/templates` dir.

config/yurt-controller/crd/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+# This kustomization.yaml is not intended to be run by itself,
```
rambohe-ch, line 1: at the same time, we should not maintain kustomization files in OpenYurt. only use helm charts to support OpenYurt.

config/yurt-controller/certmanager/certificate.yaml
rambohe-ch: by the way, we had a decision to import a more simpler tool: oamdev/kube-webhook-certgen:v2.4.1 to create or patch certificate for webhooks in OpenYurt community. 
maybe we can use this tool like yurt-app-manager: https://github.com/openyurtio/yurt-app-manager/tree/master/charts/yurt-app-manager/templates/admission-webhooks/job-patch

config/yurt-controller/manager/manager.yaml
```diff
@@ -0,0 +1,60 @@
...
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
```
rambohe-ch, line 12: how about deploy `yurt-controller-manager` under `kube-system` namespace instead system namespace?

config/yurt-controller/prometheus/monitor.yaml
```diff
@@ -0,0 +1,20 @@
+
```
rambohe-ch, line 1: maybe we don't need to consider to provide prometheus configuration for yurt-controller-manager. so how about remove this file to keep simplity?

config/yurt-controller/rbac/leader_election_role.yaml
```diff
@@ -0,0 +1,37 @@
...
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
```
rambohe-ch, line 18: please remove Line7~18, because we don't use configmap for leader election.

cmd/yurt-controller/Dockerfile
kadisi: This is only temporary and will be transferred when the processing makes build

cmd/yurt-controller/OWNERS
kadisi: This is just an owner restriction on the frame entry. Adding an owner to each component is not affected

config/yurt-controller/certmanager/certificate.yaml
kadisi: jet/kube-webhook-certgen is not recommended for production, as stated in this project. If a user deletes a certificate by mistake in a production environment, it does not automatically heal

config/yurt-controller/crd/bases/apps.openyurt.io_samples.yaml
kadisi: These yaml files are generated automatically. How to convert them into chart packages will be discussed later

config/yurt-controller/manager/manager.yaml
kadisi: ok

PR #1114, opened 2023-01-03
add tunnelAgentConnectPort to values

huweihuang:
Since users generally need to modify tunnelAgentConnectPort, it is recommended to add it to values.yaml for customization @rambohe-ch


rambohe-ch:
@huweihuang Thank you for posting pull request. I think it's a good idea to customize tunnel agent port and end users can get more higher flexibility.

rambohe-ch (approved):

PR #1123, opened 2023-01-04
Fix: add back kubeConfig init logic to fix npe

LaurenceLiZhixin:
Add back kubeConfig init logic  to fix npe.

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -136,12 +136,13 @@ func Complete(options *options.YurtHubOptions) (*YurtHubConfiguration, error) {
 	proxyServerDummyAddr := net.JoinHostPort(options.HubAgentDummyIfIP, options.YurtHubProxyPort)
 	proxySecureServerDummyAddr := net.JoinHostPort(options.HubAgentDummyIfIP, options.YurtHubProxySecurePort)
 	workingMode := util.WorkingMode(options.WorkingMode)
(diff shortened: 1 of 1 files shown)


cmd/yurthub/app/config/config.go
```diff
@@ -251,10 +252,14 @@ func buildProxiedClient(proxyAddr string) (kubernetes.Interface, error) {
 }
 
 // createSharedInformers create sharedInformers from the given proxyAddr.
-func createYurtSharedInformers(proxiedClient kubernetes.Interface, enableNodePool bool) (yurtinformers.SharedInformerFactory, error) {
+func createYurtSharedInformers(proxyAddr string, enableNodePool bool) (yurtinformers.SharedInformerFactory, error) {
 	var kubeConfig *rest.Config
```
LaurenceLiZhixin, line 256: kubeConfig would cause nil pointer  in line:
```go
if enableNodePool {
		yurtClient, err = yurtclientset.NewForConfig(kubeConfig)
		if err != nil {
			return nil, err
		}
	}

```

rambohe-ch (approved):

cmd/yurthub/app/config/config.go
Congrool: Could we just
```go
	if enableNodePool {
		yurtClient = proxiedClient
	}
```

Congrool: Ohh, it's merged, I did not fresh my page. I think it's fine to keep it as it is since it also works.

PR #1124, opened 2023-01-04
Feature: yurtadm reset/join modification. Do not remove k8s binaries, add flag for using local cni binaries.

### LINKED ISSUE #1095: [feature request] yurtadm reset caches executable files for user experience

- adding reset flag like `--cache-executable` to save executables such kubeadm, kubelet and kubectl so on.
- this option is `opt-in` option, so that it is effective only if this option is enabled. (keep the current behavior in default)

**Why is this needed**:

this will reduce the time window to complete `yurtadm join` in next time after `yurtadm reset` significantly.
according to the use case about the user devices are likely to join and leave the OpenYurt cluster system in edge environment, this will be really helpful to manage the devices.

**others**

actually i am not really sure why these executables need to be removed complete as following.
on the other hand, it sounds reasonable to reap the temporary and configuration data during reset to avoid unexpected behavior for next time to join the cluster.

https://github.com/openyurtio/openyurt/blob/c25be06cb4ae034038bc5b501e23356fa7941dc6/pkg/yurtadm/cmd/reset/phases/cleanyurtfile.go#L28-L49

**Implementation Consideration**

This could be really straight-forward to add the new flag option in,

https://github.com/openyurtio/openyurt/blob/c25be06cb4ae034038bc5b501e23356fa7941dc6/pkg/yurtadm/cmd/reset/reset.go#L66

take it effect during reset to see if it needs to keep the executable binaries as it is or not.

What needs to be cached needs to be discussed with mainline.
As far as i can see, all executable files as followings would be able to be cached.

- kubeadm, kubectl and kubelet
- all binaries under `/opt/cni/bin`

any thoughs?

/kind feature

Windrow14:
/kind feature

yurtadm reset/join modification. Do not remove k8s binaries, add flag for using local cni binaries.

Fixes #1095

- Do not remove `kubeadm/kubelet/kubelet` and cni binaries on `yurtadm reset`.
- Add new flag `--reuse-cni-bin` for `yurtadm join`. It determines whether to reuse local CNI binaries or not. Default value is `false`.

### DIFF
--- a/pkg/yurtadm/cmd/join/join.go
+++ b/pkg/yurtadm/cmd/join/join.go
@@ -271,7 +271,7 @@ func newJoinData(args []string, opt *joinOptions) (*joinData, error) {
 			Organizations: opt.organizations,
 		},
 		kubernetesResourceServer: opt.kubernetesResourceServer,
-		reuseCNIBin:           opt.reuseCNIBin,
+		reuseCNIBin:              opt.reuseCNIBin,
 	}
 
 	// parse node labels
--- a/pkg/yurtadm/cmd/join/join.go
+++ b/pkg/yurtadm/cmd/join/join.go
@@ -52,6 +52,7 @@ type joinOptions struct {
 	nodeLabels               string
 	kubernetesResourceServer string
 	yurthubServer            string
+	reuseCNIBin              bool
 }
 
 // newJoinOptions returns a struct ready for being used for creating cmd join flags.
@@ -66,6 +67,7 @@ func newJoinOptions() *joinOptions {
 		ignorePreflightErrors:    make([]string, 0),
 		kubernetesResourceServer: yurtconstants.DefaultKubernetesResourceServer,
 		yurthubServer:            yurtconstants.DefaultYurtHubServerAddr,
+		reuseCNIBin:              false,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


rambohe-ch:
@YTGhost PTAL

rambohe-ch:
@Windrow14 Thanks for posting pull request. please fix the golangci lint errors.

YTGhost:
@Windrow14 Thanks for posting pr, please also fix the unit test error, there seems to be a syntax error in the code?

rambohe-ch (approved):

fujitatomoya (commented):
sorry i was not aware of this PR, lgtm. @Windrow14 thanks for taking care on this.

Windrow14:
@YTGhost @Congrool Hi you guys, just to ask a question, why we `os.MkdirAll(constants.KubeCniDir, 0600)` but not `os.MkdirAll(constants.KubeCniDir, 0755)` when creating directory for cni binaries?

https://github.com/cilium/cilium/issues/22933

YTGhost:
> @YTGhost @Congrool Hi you guys, just to ask a question, why we `os.MkdirAll(constants.KubeCniDir, 0600)` but not `os.MkdirAll(constants.KubeCniDir, 0755)` when creating directory for cni binaries?
> 
> [cilium/cilium#22933](https://github.com/cilium/cilium/issues/22933)

@Windrow14  I think the reason for using `os.MkdirAll(constants.KubeCniDir, 0600)` at the time was to minimize permissions.

PR #1132, opened 2023-01-10
improve yurtadm join command

rambohe-ch:
/kind enhancement

1. add a util func: IsNil for verifying the nil interface.
2. change caCertHashes type from sets.String to []string
3. move some util functions from yurtadm/util to pkg/util because yurthub will use these utils later.
4. add unit tests for token util.

Fixes #

```release-note

```


rambohe-ch:
@YTGhost PTAL

YTGhost:
@rambohe-ch Please fix the ci error

PR #1138, opened 2023-01-11
bugfix: return fake getter if poolcoordinator is not enabled

Congrool:
#### What type of PR is this?
/kind bug

Fix getter function nil pointer, use fake getter instead of nil if poolcoordinator is not enabled.

Fixes #

```release-note

```


Congrool:
e2e-tests is expected to pass

Congrool:
I'll fix golangci-lint in another pr. PTAL @rambohe-ch

rambohe-ch (approved):

PR #1141, opened 2023-01-12
Fix: pool-coordinator chart format,  default service name, max body size .etc

LaurenceLiZhixin:
1. Validate pool-coordinator chart to ensure it is installable.
2. Add default pool coordinator related service name "pool-coordinator-etcd", "pool-coordinator-apiserver", and ports to yurthub default options.
3. Add '--max-txn-ops=1000' to pool-coordinator etcd container command.
4. Add default transfer body size between yurthub and etcd to '100M'

### DIFF
--- a/charts/openyurt/templates/pool-coordinator.yaml
+++ b/charts/openyurt/templates/pool-coordinator.yaml
@@ -36,57 +36,6 @@ spec:
   selector:
     k8s-app: pool-coordinator
 ---
-#apiVersion: v1
-#kind: ConfigMap
-#metadata:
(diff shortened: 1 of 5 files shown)


LaurenceLiZhixin:
After this pr is merged, I'll test both 'leader' and 'follower' mode at the same time, in my dev environment, with many edge nodes.
And I will add all Resources (not only endpoints and endpointslices) , as pool-scope resources, like pods, some CRDs, to test terformance optimization.

rambohe-ch (approved):

PR #1143, opened 2023-01-12
bugfix: yurthub can not exit when SIGINT/SIGTERM happened

rambohe-ch:
/kind bug

if you send SIGINT signal(kill -SIGINT pid) to yurthub component, yurthub can not exit and restart.
so i improved the yurthub http/https server by adding graceful stop feature.

Fixes #

```release-note

```

### DIFF
--- a/cmd/yurthub/app/start.go
+++ b/cmd/yurthub/app/start.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
@@ -32,11 +31,9 @@ import (
 	"github.com/openyurtio/openyurt/cmd/yurthub/app/options"
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
 	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
-	"github.com/openyurtio/openyurt/pkg/yurthub/certificate/token"
 	"github.com/openyurtio/openyurt/pkg/yurthub/gc"
 	"github.com/openyurtio/openyurt/pkg/yurthub/healthchecker"
 	hubrest "github.com/openyurtio/openyurt/pkg/yurthub/kubernetes/rest"
-	"github.com/openyurtio/openyurt/pkg/yurthub/network"
 	"github.com/openyurtio/openyurt/pkg/yurthub/proxy"
 	"github.com/openyurtio/openyurt/pkg/yurthub/server"
 	"github.com/openyurtio/openyurt/pkg/yurthub/tenant"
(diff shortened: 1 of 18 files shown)


cmd/yurthub/app/start.go
```diff
@@ -84,30 +81,10 @@ func NewCmdStartYurtHub(stopCh <-chan struct{}) *cobra.Command {
 
 // Run runs the YurtHubConfiguration. This should never exit
 func Run(cfg *config.YurtHubConfiguration, stopCh <-chan struct{}) error {
+	defer cfg.CertManager.Stop()
```
Congrool, line 84: Could we also use stopCh to stop CertManager?

Congrool:
Another question:
Originally, we will create 4 servers: proxy server, secure proxy server, dummyProxyServer and secure dummyProxyServer. 
But in this PR, we will only create 2 or 3 servers: 
- if enableDummyIf is set, there're proxy server, dummyProxyServer and secure dummyProxyServer
- if enableDummyIf is not set, there're proxy server and secure dummyProxyServer

If it was intentional，could you please also give some explanation?

cmd/yurthub/app/start.go
rambohe-ch: no, CertManager have stopCh that create by itself.

pkg/yurthub/network/iptables.go
```diff
@@ -57,14 +57,10 @@ func NewIptablesManager(dummyIfIP, dummyIfPort string) *IptablesManager {
...
-		// accept traffic to 169.254.2.1:10261
+		// accept traffic to 169.254.2.1:10261/169.254.2.1:10268
 		{iptables.Prepend, iptables.TableFilter, iptables.ChainInput, []string{"-p", "tcp", "-m", "comment", "--comment", "for container access hub agent", "--dport", ifPort, "--destination", ifIP, "-j", "ACCEPT"}},
-		// accept traffic from 169.254.2.1:10261
+		// accept traffic from 169.254.2.1:10261/169.254.2.1:10268
 		{iptables.Prepend, iptables.TableFilter, iptables.ChainOutput, []string{"-p", "tcp", "--sport", ifPort, "-s", ifIP, "-j", "ACCEPT"}},
-		// accept traffic to localhost:10261
-		{iptables.Prepend, iptables.TableFilter, iptables.ChainInput, []string{"-p", "tcp", "--dport", ifPort, "--destination", "localhost", "-j", "ACCEPT"}},
-		// accept traffic from localhost:10261
-		{iptables.Prepend, iptables.TableFilter, iptables.ChainOutput, []string{"-p", "tcp", "--sport", ifPort, "-s", "localhost", "-j", "ACCEPT"}},
```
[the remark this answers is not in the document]
rambohe-ch: traffic to localhost can be accepted in default status, so these iptables settings are dummy setting, so remove them.

rambohe-ch:
> Another question: Originally, we will create 4 servers: proxy server, secure proxy server, dummyProxyServer and secure dummyProxyServer. But in this PR, we will only create 2 or 3 servers:
> 
> * if `enableDummyIf` is set, there're proxy server, dummyProxyServer and secure dummyProxyServer
> * if `enableDummyIf` is not set, there're proxy server and secure proxy server
> 
> If it was intentional，could you please also give some explanation?
@Congrool originally, we have two https server: secure proxy server and secure dummyProxyServer, but in real scenario, only one secure server is used. in order to reduce the overhead of Yurthub, so we reduce the unused secure server.

btw: two http server: proxy server and dummyProxyServer(if `enableDummyIf` is true) both are used in real scenario, so we need to keep them unchanged.

PR #1144, opened 2023-01-13
Fix: dns bug for endpointslices v1beta1 version

LaurenceLiZhixin:


Congrool:
Could you please also create an issue to trace the problem? Since it will not be totally sovled in this pr.

rambohe-ch (approved):

PR #1145, opened 2023-01-13
Fix: set inElecting to false when handling follow hub, and add metrics

LaurenceLiZhixin:


pkg/yurthub/metrics/metrics.go
```diff
@@ -44,13 +44,16 @@ var (
...
+	serversHealthyCollector               *prometheus.GaugeVec
+	inFlightRequestsCollector             *prometheus.GaugeVec
+	inFlightRequestsGauge                 prometheus.Gauge
+	rejectedRequestsCounter               prometheus.Counter
+	closableConnsCollector                *prometheus.GaugeVec
+	proxyTrafficCollector                 *prometheus.CounterVec
+	proxyLatencyCollector                 *prometheus.GaugeVec
+	poolCoordinatorYurthubRoleCollector   *prometheus.GaugeVec
+	poolCoordinatorHealthyStatusCollector *prometheus.GaugeVec
+	poolCoordinatorReadyStatuaCollector   *prometheus.GaugeVec
```
Congrool, line 56: Statua -> Status

cmd/yurthub/app/start.go
```diff
@@ -330,7 +331,7 @@ func coordinatorRun(ctx context.Context,
 			return coordinatorTransportMgr
 		}, func() poolcoordinator.Coordinator {
 			return coordinator
-		}, returnErr
+		}, nil
```
Congrool, line 334: remove the returned error, since we always return nil

pkg/yurthub/metrics/metrics.go
LaurenceLiZhixin: Fixed

rambohe-ch (approved):

PR #1147, opened 2023-01-13
fix coredns cannot get /apis/discovery.k8s.io/v1 if cluster does not contain such resource

Congrool:

#### What type of PR is this?
/kind bug

If we use cordns of a high version (e.g. 1.8) but with a k8s cluster of low version (e.g. 1.20), the coredns cannot list/watch endpointslice/v1beta1, because it cannot discover the supported resource version through "GET /apis/discovery.k8s.io/v1" and "GET /apis/discovery.k8s.io/v1beta1".

The reason causing such problem is that yurthub nonresource handler will return http.StatusOK even the APIServer does not contain such resource. It will cause the coredns continuously "GET /apis/discovery.k8s.io/v1" which blocks the list/watch making it unavailable.

```go
func writeErrResponse(path string, err error, w http.ResponseWriter) {
	klog.Errorf("failed to handle %s non resource request, %v", path, err)
	status := responsewriters.ErrorToAPIStatus(err)
	output, err := json.Marshal(status)
	if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}
	writeRawJSON(output, w)
}

func writeRawJSON(output []byte, w http.ResponseWriter) {
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(http.StatusOK)
	w.Write(output)
}
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/server/nonresource.go
+++ b/pkg/yurthub/server/nonresource.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 
 	"github.com/openyurtio/openyurt/cmd/yurthub/app/config"
 	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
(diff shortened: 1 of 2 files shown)


Congrool:
Fix what we don't solve in #1144.  FYI @LaurenceLiZhixin.

BTW, an additional RBAC setting should be set to make it work. We should authorize yurthub to list/watch endpointslice. But it seems we do not need to modify the RBAC for coredns 1.8 in k8s 1.22. I don't know the reason.

rambohe-ch (approved):

Congrool:
/hold

need to fix unit test

LaurenceLiZhixin:
LGTM

PR #1148, opened 2023-01-14
Fix: add PoolScopeResource validation and dynamic configuration support

LaurenceLiZhixin:
1. add PoolScopeResource validation to support multi version of pool scope resources, like v1/v1beta1 version  of endpointslices.

2. Add pool scope resource type configuration, which read from yurt-hub-cfg cm's `pool_scope_resources` key.
For example:
we can add 
```
 pool_scope_resources: '[{"Group":"","Version":"v1","Resource":"pods"}]'
```
into yurt-hub-cfg config map to let pool-coordinator cache pods resources.

3. Add etcd transaction max num to 100k.
4. Add dynamic client for pool-coordinator to support crd informer.


pkg/yurthub/poolcoordinator/resources/resources.go
```diff
@@ -0,0 +1,167 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/util"
+)
+
+type PoolScopeResourcesManger struct {
+	factory                      informers.SharedInformerFactory
```
rambohe-ch, line 22: `factory` field is unused

pkg/yurthub/poolcoordinator/resources/resources.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (m *PoolScopeResourcesManger) getGroupVersionVerifyFunction(client kubernetes.Interface) func(gvr schema.GroupVersionResource) (bool, string) {
+	return func(gvr schema.GroupVersionResource) (bool, string) {
+		maxRetry := 3
+		duration := time.Second * 5
+		counter := 0
+		var err error
+		for counter <= maxRetry {
+			if _, err = client.Discovery().ServerResourcesForGroupVersion(gvr.GroupVersion().String()); err == nil {
```
rambohe-ch, line 141: if cloud-edge network is off, client can not verify the gvr exist or not. maybe we need to store the gvr info into local disk for yurthub restart use.

pkg/yurthub/poolcoordinator/resources/resources.go
LaurenceLiZhixin: fixed

pkg/yurthub/poolcoordinator/resources/resources.go
LaurenceLiZhixin: Yes, Shall we do that in the next release? At current release,  if cloud-edge network is off, the pool-coordinator would not work as there would be no leader. Let's do fix these issues together in the next release!
@rambohe-ch

rambohe-ch: ok

rambohe-ch (approved):

PR #1153, opened 2023-01-16
bugfix: ReplaceComponentList should not delete resources of other gvr

Congrool:

#### What type of PR is this?
/kind bug

`ReplaceComponentList` of etcd storage has a bug, which will delete resources of other gvr. For example, kubelet stored 3 keys in etcd, pod1, pod2 and configmap1 respectively. When using `ReplaceComponentList("kubelet",podGVR,"",newPodList)` to replace the pod keys in etcd, pod1 and pod2 will be replaced with newPodList, but the configmap1 will also be deleted. That's not what we want.

BTW, this pr also add more unit tests for keycache.go

Fixes #

```release-note

```


rambohe-ch (approved):

rambohe-ch:
@Congrool please fix the conflicting files.

rambohe-ch (approved):

PR #1164, opened 2023-01-19
bugfix: yurt-controller-manager cannot restart

Congrool:
#### What type of PR is this?
/kind bug

When yurt-controller-manager restart, it will encounter fatal error and exit because the mutatingwebhookconfiguration and validatingwebhookconfiguration has already exists.

Fixes #

```release-note

```


Congrool:
@gnunu @LindaYu17 PTAL

rambohe-ch (approved):

PR #1165, opened 2023-01-19
fix: leader yurthub cannot proxy heartbeat

Congrool:

#### What type of PR is this?
/kind bug

Yurthub cannot delegate node lease because the resourceVersion is 0. Thus, the update will always fail.

Note:
Even this pr is merged, leader yurthub may also cannot delegate heartbeat, beacuse of the noderestiction admission plugin. It only allows the node to update its own kube-node-lease with node certificate. See [noderestriction plugin](https://github.com/kubernetes/kubernetes/blob/v1.22.7/plugin/pkg/admission/noderestriction/admission.go#L542-L565)
To enable this capability, we should disable the admission plugin in APIServer command line options, like

```yaml
...
spec:
  containers:
  - command:
    - kube-apiserver
    - --advertise-address=172.19.0.3
    - --allow-privileged=true
    - --authorization-mode=Node,RBAC
    - --client-ca-file=/etc/kubernetes/pki/ca.crt
#   - --enable-admission-plugins=NodeRestriction
    - --enable-bootstrap-token-auth=true
    - --etcd-cafile=/etc/kubernetes/pki/etcd/ca.crt
    - --etcd-certfile=/etc/kubernetes/pki/apiserver-etcd-client.crt
    - --etcd-keyfile=/etc/kubernetes/pki/apiserver-etcd-client.key
    - --etcd-servers=https://127.0.0.1:2379
    - --kubelet-client-certificate=/etc/kubernetes/pki/apiserver-kubelet-client.crt
    - --kubelet-client-key=/etc/kubernetes/pki/apiserver-kubelet-client.key
    - --proxy-client-cert-file=/etc/kubernetes/pki/front-proxy-client.crt
    - --proxy-client-key-file=/etc/kubernetes/pki/front-proxy-client.key
    - --requestheader-allowed-names=front-proxy-client
    - --requestheader-client-ca-file=/etc/kubernetes/pki/front-proxy-ca.crt
    - --requestheader-extra-headers-prefix=X-Remote-Extra-
    - --requestheader-group-headers=X-Remote-Group
    - --requestheader-username-headers=X-Remote-User
    - --runtime-config=
    - --secure-port=6443
    - --service-account-issuer=https://kubernetes.default.svc.cluster.local
    - --service-account-key-file=/etc/kubernetes/pki/sa.pub
    - --service-account-signing-key-file=/etc/kubernetes/pki/sa.key
    - --service-cluster-ip-range=10.96.0.0/16
    - --tls-cert-file=/etc/kubernetes/pki/apiserver.crt
    - --tls-private-key-file=/etc/kubernetes/pki/apiserver.key
    image: k8s.gcr.io/kube-apiserver:v1.22.7
...
```

Fixes #

```release-note

```

### DIFF
--- a/charts/openyurt/templates/pool-coordinator.yaml
+++ b/charts/openyurt/templates/pool-coordinator.yaml
@@ -201,3 +201,29 @@ spec:
                     - secret:
                         name: pool-coordinator-static-certs
                 name: pool-coordinator-certs
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openyurt:pool-coordinator:node-lease-proxy-client 
+rules:
+  - apiGroups:
+      - "coordination.k8s.io"
+    resources:
+      - leases
+    verbs:
+      - get
+      - update
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: openyurt:pool-coordinator:node-lease-proxy-client
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: openyurt:pool-coordinator:node-lease-proxy-client
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: openyurt:pool-coordinator:node-lease-proxy-client
--- a/cmd/yurthub/app/start.go
+++ b/cmd/yurthub/app/start.go
@@ -270,7 +270,7 @@ func coordinatorRun(ctx context.Context,
 		}
 		go elector.Run(ctx.Done())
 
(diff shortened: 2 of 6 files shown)


rambohe-ch:
@Congrool we could not disable [noderestriction plugin](https://github.com/kubernetes/kubernetes/blob/v1.22.7/plugin/pkg/admission/noderestriction/admission.go#L542-L565) in kube-apiserver because the privilege of node certificate should be checked. so maybe we need to add another certificate or seriveaccount for proxying node leases.

maybe we can improve this feature in the next version like v1.2.1.

Congrool:
Add a new certificate called "node-lease-proxy-client" into `pool-coordinator-yurthub-certs`, which is binded to clusterrole "node-lease-proxy-client". Yurthub will use such certificate to build client to get/update kube node lease for other nodes.

rambohe-ch (approved):

Issue #1079, opened 2022-11-28
[feature request]support OTA and Auto upgrade model for static pod


rambohe-ch:
**What would you like to be added**:
As we discussed in proposal: https://github.com/openyurtio/openyurt/pull/1065, make a issue to track this work.

**Why is this needed**:
please reference the proposal link: [detail info ](https://github.com/openyurtio/openyurt/pull/1065)

**others**
/kind feature

PR #1168, opened 2023-01-19
add app to support upgrade static pod

xavier-hou:
Signed-off-by: hxcGit <houxc_mail@163.com>

Relate to: #1079 

```release-note

```


cmd/yurt-static-pod-upgrade/static-pod-upgrade.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	manifest string
+	mode     string
+)
+
+func main() {
+	rand.Seed(time.Now().UnixNano())
+	version := fmt.Sprintf("%#v", projectinfo.Get())
+	cmd := &cobra.Command{
+		Use: "yurt-static-pod-upgrade",
```
rambohe-ch, line 42: I think we don't need to add a new command, how about add a sub command for upgrading static pod in node-servant component?

xavier-hou: Agree! I will move it to node-servant.

pkg/static-pod-upgrade/upgrade.go
```diff
@@ -0,0 +1,143 @@
...
+	manifestPath string
+	// The backup manifest path, default `/etc/kubernetes/manifests/openyurtio-upgrade/manifestName.bak`
+	bakManifestPath string
+	// Default is `/data/podName`
+	configMapDataPath string
+	// The latest manifest path, default `/etc/kubernetes/manifests/openyurtio-upgrade/manifestName.upgrade`
+	upgradeManifestPath string
+}
+
+func New(manifest, mode string) (*Controller, error) {
```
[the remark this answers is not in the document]
xavier-hou: I found that I didn't fully get what you meant, at present auto and ota modes are using the same code to upgrade.
https://github.com/openyurtio/openyurt/pull/1168/files#diff-173ef2ba90c763e037ef16b2910795a45bf51cafa5cdc979a081600280792f66R70

cmd/yurt-static-pod-upgrade/static-pod-upgrade.go
xavier-hou: Done

pkg/node-servant/static-pod-upgrade/upgrade.go
```diff
@@ -0,0 +1,172 @@
...
+		namespace:   namespace,
+		manifest:    manifest,
+		hash:        hash,
+		upgradeMode: mode,
+	}
+
+	ctrl.manifestPath = filepath.Join(DefaultManifestPath, util.WithYamlSuffix(ctrl.manifest))
+	ctrl.bakManifestPath = filepath.Join(DefaultManifestPath, DefaultUpgradeDir, util.WithBackupSuffix(ctrl.manifest))
+	ctrl.configMapDataPath = filepath.Join(DefaultConfigmapPath, ctrl.manifest)
+	ctrl.upgradeManifestPath = filepath.Join(DefaultManifestPath, DefaultUpgradeDir, util.WithUpgradeSuffix(ctrl.manifest))
```
rambohe-ch, line 78: I think we should not pollute `/etc/kubernetes/manifests` directory, so how about place `manifest.bak` and `manifests.upgrade` file under `/tmp/manifests` directory?

pkg/node-servant/static-pod-upgrade/util/util.go
```diff
@@ -0,0 +1,112 @@
...
+
+		return false, false
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return false, fmt.Errorf("timeout waiting for static pod %s/%s to be running", namespace, name)
+		default:
+			pod, err := GetPodFromYurtHub(namespace, name)
```
rambohe-ch, line 98: it will cost some time before new static pod startup, so it is normal for `GetPodFromYurtHub` to get the old static pod or fail to find pod. 

and new static pod check policy should be:  we will wait for two minutes until new static pod is ready before timeout.

pkg/node-servant/static-pod-upgrade/upgrade.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	klog.Info("Auto upgrade replaceManifest success")
+
+	// (3) Verify the new static pod is running
+	ok, err := ctrl.verify()
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return fmt.Errorf("the latest static pod is not running")
```
rambohe-ch, line 124: maybe we need to fall into the backup manifest if upgrade failed.

pkg/node-servant/static-pod-upgrade/upgrade.go
```diff
@@ -0,0 +1,172 @@
...
+	"time"
+
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/node-servant/static-pod-upgrade/util"
+)
+
+const (
+	DefaultUpgradeDir                   = "openyurtio-upgrade"
+	DefaultStaticPodRunningCheckTimeout = 2 * time.Minute
```
rambohe-ch, line 32: how about add a new parameter for upgrade command to specify static pod check timeout? because sometimes two minutes maybe not enough.

xavier-hou: done

pkg/node-servant/static-pod-upgrade/upgrade.go
xavier-hou: done

pkg/node-servant/static-pod-upgrade/upgrade.go
xavier-hou: done

pkg/node-servant/static-pod-upgrade/util/util.go
xavier-hou: It will not return until new pod is running or work pod fails.

pkg/node-servant/static-pod-upgrade/util/util.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+			if pod != nil {
+				hasResult, result := checkPod(pod)
+				if hasResult {
+					return result, nil
+				}
+			}
+		}
+		<-ticker.C
+	}
```
rambohe-ch, line 112: line 95 ~111 can be optimized as following:

```
	for {
		select {
		case <-ctx.Done():
			return false, fmt.Errorf("timeout waiting for static pod %s/%s to be running", namespace, name)
		case <-ticker.C:
			pod, err := GetPodFromYurtHub(namespace, name)
			if err != nil {
				klog.V(4).Infof("Temporarily fail to get pod from YurtHub, %v", err)
			}
			if pod != nil {
				hasResult, result := checkPod(pod)
				if hasResult {
					return result, nil
				}
			}
		}
	}
```

pkg/node-servant/static-pod-upgrade/upgrade.go
```diff
@@ -0,0 +1,181 @@
...
+	// (2) Replace manifest and kubelet will upgrade the static pod automatically
+	if err := ctrl.replaceManifest(); err != nil {
+		return err
+	}
+	klog.Info("Auto upgrade replaceManifest success")
+
+	// (3) Verify the new static pod is running
+	ok, err := ctrl.verify()
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
xavier-hou: My bad. Done.

pkg/node-servant/static-pod-upgrade/util/util.go
xavier-hou: done

rambohe-ch (approved):

PR #1172, opened 2023-01-27
support ota upgrade for static pod

xavier-hou:
Signed-off-by: hxcGit <houxc_mail@163.com>

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/otaupdate/daemon/daemon_pod.go
+++ b/pkg/yurthub/otaupdate/daemon/daemon_pod.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2023 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package daemon
+
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+type DaemonPodUpgrader struct {
+	kubernetes.Interface
+	types.NamespacedName
+}
+
+// applyUpdate execute pod update process by deleting pod under OnDelete update strategy
+func (s *DaemonPodUpgrader) Apply() error {
+	err := s.CoreV1().Pods(s.Namespace).Delete(context.TODO(), s.Name, metav1.DeleteOptions{})
+	if err != nil {
+		klog.Errorf("Update pod %v/%v failed when delete, %v", s.Namespace, s.Name, err)
+		return err
+	}
+
+	klog.Infof("Start updating pod: %v/%v", s.Namespace, s.Name)
+	return nil
+}
--- a/pkg/yurthub/otaupdate/daemon/daemon_pod.go
+++ b/pkg/yurthub/otaupdate/daemon/daemon_pod.go
@@ -30,7 +30,7 @@ type DaemonPodUpgrader struct {
 	types.NamespacedName
 }
 
-// applyUpdate execute pod update process by deleting pod under OnDelete update strategy
+// Apply execute pod update process by deleting pod under OnDelete update strategy
 func (s *DaemonPodUpgrader) Apply() error {
 	err := s.CoreV1().Pods(s.Namespace).Delete(context.TODO(), s.Name, metav1.DeleteOptions{})
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/yurthub/otaupdate/ota.go
```diff
@@ -87,29 +97,48 @@ func UpdatePod(clientset kubernetes.Interface, nodeName string) http.Handler {
 		namespace := params["ns"]
 		podName := params["podname"]
 
-		err, ok := applyUpdate(clientset, namespace, podName, nodeName)
-		// Pod update failed with error
-		if err != nil {
-			WriteErr(w, "Apply update failed", http.StatusInternalServerError)
-			return
-		}
+		pod, ok := preCheck(clientset, namespace, podName, nodeName)
```
rambohe-ch, line 100: It seems that preCheck is not suitable for static pod upgrade. we need to add some verification for avoiding static pod upgrading reenter.

pkg/yurthub/otaupdate/ota.go
```diff
@@ -87,29 +97,48 @@ func UpdatePod(clientset kubernetes.Interface, nodeName string) http.Handler {
...
 }
 
-// applyUpdate execute pod update process by deleting pod under OnDelete update strategy
-func applyUpdate(clientset kubernetes.Interface, namespace, podName, nodeName string) (error, bool) {
+// preCheck will check the necessary requirements before apply upgrade
+// 1. target pod has not been deleted yet
+// 2. target pod belongs to current node
+// 3. check whether target pod is updatable
+// At last, return the target pod to do further operation
+func preCheck(clientset kubernetes.Interface, namespace, podName, nodeName string) (*corev1.Pod, bool) {
```
rambohe-ch, line 137: I think we can add static pod configmap loading as one of check items in preCheck.

pkg/yurthub/otaupdate/ota.go
xavier-hou: The new static pod won't have `PodNeedUpgrade` condition and it's not likely to re-enter the upgrade process.

pkg/yurthub/otaupdate/ota.go
xavier-hou: `preCheck` function is used to check whether the given pod can be upgraded or not. Configmap has nothing to do with this check. StaticPod controller will make sure the relevant configmap existed in the cluster. 

If you think necessary, we can use dynamic client to get targeted CR and then generate the manifest directly when we cannot find the configmap.

pkg/yurthub/otaupdate/static/static_pod.go
```diff
@@ -0,0 +1,101 @@
...
+	DefaultManifestPath = "/etc/kubernetes/manifests"
+	DefaultUpgradePath  = "/tmp/manifests"
+)
+
+type StaticPodUpgrader struct {
+	kubernetes.Interface
+	types.NamespacedName
+}
+
+func (s *StaticPodUpgrader) Apply() error {
```
rambohe-ch, line 45: how about use [controller.Upgrade()](https://github.com/openyurtio/openyurt/blob/master/pkg/node-servant/static-pod-upgrade/upgrade.go#L106) to complete static pod upgrade?

pkg/yurthub/otaupdate/static/static_pod.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		klog.V(5).Infof("Create upgrade dir %v", DefaultUpgradePath)
+	}
+
+	// 2. Get upgrade manifest from api-serv
+	cm, err := s.CoreV1().ConfigMaps(s.Namespace).Get(context.TODO(),
+		spctrlutil.WithConfigMapPrefix(spctrlutil.Hyphen(s.Namespace, s.Name)), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
```
rambohe-ch, line 63: how about move `getting configmap` into `preCheck`?

pkg/yurthub/otaupdate/static/static_pod.go
```diff
@@ -0,0 +1,101 @@
+/*
```
rambohe-ch, line 1: how about move `pkg/yurthub/otaupdate/static/static_pod.go` into `pkg/yurthub/otaupdate/upgrader/static_pod.go`

pkg/yurthub/otaupdate/daemon/daemon_pod.go
```diff
@@ -0,0 +1,43 @@
+/*
```
rambohe-ch, line 1: how about move `pkg/yurthub/otaupdate/daemon/daemon_pod.go` into `pkg/yurthub/otaupdate/upgrader/daemon_pod.go`

pkg/yurthub/otaupdate/ota.go
rambohe-ch: @xavier-hou I don't agree with you, and i think that configmap loading is one part of static pod upgrading, because if we fail to download the configmap, the static pod upgrade should not proceed.

pkg/yurthub/otaupdate/static/static_pod.go
xavier-hou: Done

pkg/yurthub/otaupdate/static/static_pod.go
xavier-hou: Done

pkg/yurthub/otaupdate/daemon/daemon_pod.go
xavier-hou: Done

pkg/yurthub/otaupdate/ota.go
xavier-hou: I prefer to check the existence of configmap in `Switch-case` struct. 
1. Daemon pods do not need this check.
2. It's hard to judge the pod type before get this pod and `preCheck` is also used to get the pod in the previous implementation.

rambohe-ch (approved):

PR #1177, opened 2023-01-29
imporve pool coordinator controller

rambohe-ch:
/kind enhancement

according to the discussion with @gnunu and @Congrool , improve pool coordinator controller as following:
1. pool-coordinator controller only need to monitor lease resource with delegate annotation, and taints the node with unschedulable taint.
2. add a new pod binding controller for adding tolerations for pods, in order to adapt node with autonomy annotation and pod with binding annotation.

Fixes #

```release-note

```


Congrool:
@rambohe-ch What about pool autonomy? If we set bind annotation, will this pod start on another node in same node pool when its original node break down?

rambohe-ch:
> @rambohe-ch What about pool autonomy? If we set bind annotation, will this pod start on another node in same node pool when its original node break down?

@Congrool All of pods in NodePool can get the ability for pool-coordinator delegating heartbeat, so it's not need for pod to add nodepool autonomy annotation.

pkg/controller/poolcoordinator/podbinding/pod_binding_controller.go
```diff
@@ -0,0 +1,327 @@
...
+		runtime.HandleError(err)
+		return true
+	}
+
+	c.podUpdateQueue.Forget(key)
+	return true
+}
+
+func (c *Controller) Run(stopCH <-chan struct{}) {
+	if !cache.WaitForCacheSync(stopCH, c.nodeSynced, c.podSynced) {
```
Congrool, line 268: stopCH -> stopCh

pkg/controller/poolcoordinator/podbinding/pod_binding_controller.go
```diff
@@ -0,0 +1,327 @@
...
+	nodeInformer := informerFactory.Core().V1().Nodes()
+	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: ctl.onNodeUpdate,
+	})
+	ctl.nodeSynced = nodeInformer.Informer().HasSynced
+	ctl.nodeLister = nodeInformer.Lister()
+
+	podInformer := informerFactory.Core().V1().Pods()
+	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: ctl.onPodUpdate,
```
Congrool, line 153: Do we need handle add event?

charts/openyurt/templates/yurt-controller-manager.yaml
```diff
@@ -135,24 +126,6 @@ rules:
...
-  - apiGroups:
-      - admissionregistration.k8s.io
-    resources:
-      - validatingwebhookconfigurations
-      - mutatingwebhookconfigurations
-    verbs:
-      - create
-      - delete
-      - get
-      - update
```
Congrool, line 155: Please also update Charts

pkg/yurtctl/constants/constants.go
```diff
@@ -224,7 +219,8 @@ spec:
       - name: yurt-controller-manager
         image: {{.image}}
         command:
-        - yurt-controller-manager	
+        - yurt-controller-manager
+        - --controllers=-servicetopologycontroller,-poolcoordinatorcertmanager,*
```
Congrool, line 223: These two controllers do not run by default?

pkg/controller/poolcoordinator/constant/constant.go
```diff
@@ -18,14 +18,14 @@ limitations under the License.
...
-	AnnotationKeyNodeAutonomy = "node.beta.openyurt.io/autonomy"
-	LabelKeyNodePool          = "apps.openyurt.io/nodepool"
-
 	DelegateHeartBeat = "openyurt.io/delegate-heartbeat"
 
 	// when node cannot reach api-server directly but can be delegated lease, we should taint the node as unschedulable
 	NodeNotSchedulableTaint = "node.openyurt.io/unschedulable"
+
+	// PodBindingAnnotation specify that pod should bind to the node that has been scheduled
+	PodBindingAnnotation = "apps.openyurt.io/binding"
```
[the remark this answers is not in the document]
rambohe-ch: fixed

pkg/controller/poolcoordinator/podbinding/pod_binding_controller.go
rambohe-ch: we need to wait pod phase become `Running` before changing tolerations for pod, so it doesn't need to handle add event.

charts/openyurt/templates/yurt-controller-manager.yaml
rambohe-ch: fixed

pkg/yurtctl/constants/constants.go
rambohe-ch: these two controllers depend on the yurt-app-manager component, but in e2e tests, yurt-app-manager component is not deployed, so disable these two controllers.

Congrool: OK

PR #1191, opened 2023-02-06
[Enhancement] improve image build efficiency

Congrool:

#### What type of PR is this?
/kind enhancement

Currently, if we build all images, it has to download dependent go package for several times, which makes the build too slow. 
In this pr, we will start a container `openyurt-build` which will build all binaries  first, and then using these built binaries to build images.

Also, we can still only build one of these image as we need. For example:
```bash
make docker-build WHAT=yurthub
```

Fixes #

```release-note

```


Congrool:
The time of build-image steps can be reduced from orginal 6m-9m to current 2m30s.

rambohe-ch:
@Congrool It's good job to improve image build efficiency, and cost time of e2e-tests has reduced from 25m to 17m.

but `make docker-push` command can not work after `hack/dockerfiles/xxx` modified, maybe we need to prepare two different dockerfiles for `make docker-push` and `make docker-build`.

Congrool:
This current improvement is only useful for build images, but not suitable for relasing images which is cross-platform.

rambohe-ch (approved):

xavier-hou:
When I use `make docker-build`, I find most of the time is taken up by downloading dependent go packages even though this process has now been reduced to just once. 

When testing locally, most of dependency libraries have been downloaded to local machine in advance, which will reduce the time it takes to execute `go build`. So does it make sense to build binaries locally and then copy them to container environment to build the images?

PR #1195, opened 2023-02-07
fix: re-list when target change

LaurenceLiZhixin:
Force re-list pool-scope resource when proxy change between loadbalancer and pool-coordinator

### DIFF
--- a/pkg/yurthub/proxy/remote/loadbalancer.go
+++ b/pkg/yurthub/proxy/remote/loadbalancer.go
@@ -178,10 +178,10 @@ func NewLoadBalancer(
 		algo = &rrLoadBalancerAlgo{backends: backends, checker: healthChecker}
 	}
 
-	return &loadBalancer{
-		backends: backends,
-		algo:     algo,
-	}, nil
+	lb.backends = backends
+	lb.algo = algo
+
+	return lb, nil
 }
 
 func (lb *loadBalancer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
@@ -211,19 +211,18 @@ func (lb *loadBalancer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			for {
 				select {
 				case <-t.C:
-					if lb.coordinatorGetter == nil {
-						continue
-					}
 					coordinator := lb.coordinatorGetter()
 					if coordinator == nil {
 						continue
 					}
 					if _, isReady := coordinator.IsReady(); isReady {
 						klog.Infof("notified the pool coordinator is ready, cancel the req %s making it handled by pool coordinator", hubutil.ReqString(req))
+						util.ReListWatchReq(rw, req)
 						cloudServeCancel()
 						return
 					}
 				case <-clientReqCtx.Done():
+					klog.Infof("watch req %s is canceled by client, when pool coordinator is not ready", hubutil.ReqString(req))
 					return
 				}
 			}
--- a/pkg/yurthub/proxy/pool/pool.go
+++ b/pkg/yurthub/proxy/pool/pool.go
@@ -164,6 +164,7 @@ func (pp *PoolCoordinatorProxy) poolWatch(rw http.ResponseWriter, req *http.Requ
 				case <-t.C:
 					if !pp.isCoordinatorReady() {
 						klog.Infof("notified the pool coordinator is not ready for handling request, cancel watch %s", hubutil.ReqString(req))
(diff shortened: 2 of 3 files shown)


Congrool:
I still have some questions for such problem:
1. client sent list/watch requests and they were handled by yurthub. Then the yurthub crashed and restarted. And at this time, the handler changed(possibly because pool-coordinator become not ready), will the client relist the resource?
2. When the long time watch connection breaks causing by the server-side(e.g. cloud-edge network breaks or pool-coordinator restarted),  will yurthub always make the client relist?

LaurenceLiZhixin:
Thanks for your great comments @Congrool 

1. After my tests, yurthub restart will cause all existing watch requests to be disconnected from client side, and the client will trigger a re-list-watch. If the restart process occurs with the flow change from edge to cloud, after the stability, yurthub got ready coordinator status, and yurthub will cut the cloud to the edge again , the client-side re-list-watch will still be triggered by the error event injection provided by this pr.

2.
- If the cloud edge network is disconnected: If yurthub is directly connected to the cloud, the edge pool-coordinator is unavailable. Therefore, there is no problem.
- If the pool-coordinator restarts, the existing watch requests are cut and the client re-list-watch. Then yurthub will determine that a pool-coordinator is abnormal, and trigger a re-list-watch for all existing watches through the error event injection provided by this pr, and then client side list-watch agent and yurthub proxy to the cloud. After the pool-coordinator restarted, the flow is changed to local and all re-list-watch are triggered again. There will be no problem after the experiment.

pkg/yurthub/proxy/remote/loadbalancer.go
```diff
@@ -211,19 +211,21 @@ func (lb *loadBalancer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			for {
 				select {
 				case <-t.C:
-					if lb.coordinatorGetter == nil {
-						continue
-					}
+					klog.Infof("lb.coordinatorGetter is %s", lb.coordinatorGetter)
```
Congrool, line 214: We may not need such log, please remove

pkg/yurthub/proxy/remote/loadbalancer.go
```diff
@@ -211,19 +211,21 @@ func (lb *loadBalancer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 					coordinator := lb.coordinatorGetter()
 					if coordinator == nil {
+						klog.Infof("lb.coordinatorGetter().coordinator is nil")
```
Congrool, line 217: ditto

pkg/yurthub/proxy/remote/loadbalancer.go
```diff
@@ -211,19 +211,21 @@ func (lb *loadBalancer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 						continue
 					}
 					if _, isReady := coordinator.IsReady(); isReady {
 						klog.Infof("notified the pool coordinator is ready, cancel the req %s making it handled by pool coordinator", hubutil.ReqString(req))
+						util.ReListWatchReq(rw, req)
 						cloudServeCancel()
 						return
 					}
+					klog.Infof("coordinator.IsReady() is false")
```
Congrool, line 226: ditto

Congrool:
Sounds good to me

Congrool:
Could you please squash your commits to make the commit message more readable?

Congrool:
Another question, is it possible that:
list request is first handled by one handler(for example cloud APIServer), but the later watch request is handled by another handler(for pool-coordinator becoming healthy)

Will it still have some problems in such situation?

pkg/yurthub/proxy/remote/loadbalancer.go
LaurenceLiZhixin: Fine!

LaurenceLiZhixin:
> Another question, is it possible that: list request is first handled by one handler(for example cloud APIServer), but the later watch request is handled by another handler(for pool-coordinator becoming healthy)
> 
> Will it still have some problems in such situation?

Yes, that's still a problem. But it's very rare. I think we can talk about it and fix it totally later.

LaurenceLiZhixin:
I've fixed your great comments above. @Congrool

Congrool:
LGTM as a solution for most cases, thanks for your contribution!

/lgtm

rambohe-ch (approved):

PR #1209, opened 2023-02-08
add nodeportisolation filter in yurthub

### LINKED ISSUE #1183: [feature request]Need support NodePort service isolate
we need NodePort service isolate function, when create a NodePort service we can choose which node or nodepool it worked!

**Why is this needed**:
when we use edge network ,we need to make service isolate, for example, when we use NodePort type service ,we just want it servers some specific node pool.

it needs some feature
1. pod must run at some specific node pool, we can use label to finish it.
2. service can only forward network flow to specific node pool pod, we can also use label selector to finish it.
3. NodePort service can be isolate, for example when we create a NodePort service open node 80 port, it only need to open nodes which in specific edge node pool, not all nodes in all edge.

**others**
/kind feature

rambohe-ch:
/kind feature

add a new filter named nodeportisolation in yurthub in order to make only nodes in specified nodepool can listen on nodePort of service.

Fixes #1183 

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/nodeportisolation/filter.go
+++ b/pkg/yurthub/filter/nodeportisolation/filter.go
@@ -1,0 +1,243 @@
+/*
+Copyright 2023 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeportisolation
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
+	"github.com/openyurtio/openyurt/pkg/yurthub/storage"
+	"github.com/openyurtio/openyurt/pkg/yurthub/storage/disk"
+	"github.com/openyurtio/openyurt/pkg/yurthub/util"
+	nodepoolv1alpha1 "github.com/openyurtio/yurt-app-manager-api/pkg/yurtappmanager/apis/apps/v1alpha1"
+)
+
+const (
+	ServiceAnnotationNodePortListen = "nodeport.openyurt.io/listen"
+)
+
+// Register registers a filter
+func Register(filters *filter.Filters) {
+	filters.Register(filter.NodePortIsolationName, func() (filter.ObjectFilter, error) {
+		return &nodePortIsolationFilter{}, nil
+	})
+}
+
+type nodePortIsolationFilter struct {
+	nodePoolName string
+	workingMode  util.WorkingMode
+	nodeGetter   filter.NodeGetter
+	nodeSynced   cache.InformerSynced
+	nodeName     string
+}
+
+func (nif *nodePortIsolationFilter) Name() string {
+	return filter.NodePortIsolationName
+}
+
+func (nif *nodePortIsolationFilter) SupportedResourceAndVerbs() map[string]sets.String {
+	return map[string]sets.String{
+		"services": sets.NewString("list", "watch"),
+	}
+}
+
+func (nif *nodePortIsolationFilter) SetNodePoolName(name string) error {
+	nif.nodePoolName = name
+	return nil
+}
+
+func (nif *nodePortIsolationFilter) SetNodeName(nodeName string) error {
+	nif.nodeName = nodeName
+	return nil
+}
+
+func (nif *nodePortIsolationFilter) SetWorkingMode(mode util.WorkingMode) error {
+	nif.workingMode = mode
+	return nil
+}
+
+func (nif *nodePortIsolationFilter) SetSharedInformerFactory(factory informers.SharedInformerFactory) error {
+	if nif.workingMode == util.WorkingModeCloud {
+		klog.Infof("prepare list/watch to sync node(%s) for cloud working mode", nif.nodeName)
+		nif.nodeSynced = factory.Core().V1().Nodes().Informer().HasSynced
+		nif.nodeGetter = factory.Core().V1().Nodes().Lister().Get
+	}
+
+	return nil
+}
+
+func (nif *nodePortIsolationFilter) SetStorageWrapper(s cachemanager.StorageWrapper) error {
+	if s.Name() != disk.StorageName {
+		return fmt.Errorf("nodePortIsolationFilter can only support disk storage currently, cannot use %s", s.Name())
+	}
+
+	if len(nif.nodeName) == 0 {
+		return fmt.Errorf("node name for nodePortIsolationFilter is not set")
+	}
+
+	// hub agent will list/watch node from kube-apiserver when hub agent work as cloud mode
+	if nif.workingMode == util.WorkingModeCloud {
+		return nil
+	}
+	klog.Infof("prepare local disk storage to sync node(%s) for edge working mode", nif.nodeName)
+
+	nodeKey, err := s.KeyFunc(storage.KeyBuildInfo{
+		Component: "kubelet",
+		Name:      nif.nodeName,
+		Resources: "nodes",
+		Group:     "",
+		Version:   "v1",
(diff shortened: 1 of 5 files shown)


rambohe-ch:
@Congrool @yuanmaomao PTAL

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -0,0 +1,238 @@
...
+	if len(nodePoolName) == 0 {
+		return svc
+	}
+
+	nsName := fmt.Sprintf("%s/%s", svc.Namespace, svc.Name)
+	if svc.Spec.Type == v1.ServiceTypeNodePort {
+		if len(svc.Annotations[ServiceAnnotationNodePortIsolation]) != 0 {
+			enabledNodePools, disabledNodePools := parseNodePoolFromAnnotation(svc.Annotations[ServiceAnnotationNodePortIsolation])
+			if disabledNodePools.Has(nodePoolName) {
+				klog.V(2).Infof("nodePort service(%s) is isolated in nodePool(%s) by nodePortIsolationFilter", nsName, nodePoolName)
```
Congrool, line 202: "nodePort service(%s) is disabled in nodePool(%s) by nodePortIsolationFilter"

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -0,0 +1,238 @@
...
+
+			if enabledNodePools.Has(nodePoolName) {
+				return svc
+			}
+
+			if enabledNodePools.Has("*") {
+				return svc
+			}
+
+			klog.V(2).Infof("nodePort service(%s) is isolated in nodePool(%s) by nodePortIsolationFilter", nsName, nodePoolName)
```
Congrool, line 214: ditto

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -0,0 +1,238 @@
...
+		if err != nil {
+			klog.Warningf("skip isolateNodePortService filter, failed to get node(%s), %v", nif.nodeName, err)
+			return svc
+		}
+		nodePoolName = node.Labels[nodepoolv1alpha1.LabelCurrentNodePool]
+	}
+
+	// node is not located in NodePool, skip this filter
+	if len(nodePoolName) == 0 {
+		return svc
```
Congrool, line 194: Is it suitable that skipping filter for node that does not in any nodepool?

1. If users only set enabled nodepools for svcs, it looks strange that such nodePort svcs will also work on these orphan nodes.

2. But if users only set disabled nodepools for svcs, it's intuitive that these svcs should work on all nodes in other nodepools and all orphan nodes. 

3. If users set both enable nodepools and disable nodepools, it maybe confused because such two sets may overlap. For example, "np1,np2,-np1,-np3", we have to make more clear definition for such case.

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -0,0 +1,238 @@
...
+			if enabledNodePools.Has(nodePoolName) {
+				return svc
+			}
+
+			if enabledNodePools.Has("*") {
+				return svc
+			}
+
+			klog.V(2).Infof("nodePort service(%s) is isolated in nodePool(%s) by nodePortIsolationFilter", nsName, nodePoolName)
+			return nil
```
Congrool, line 215: In other case, we return nil to disable the nodeport svc. It looks confusing. For example, we have "np1,-np2" which also implicitly disable this svc in np3, np4 and so on. Users may be confused when encountering this scenario.

Congrool:
Another problem, this filter seems cannot react to the nodepool change, for example the node is moved to another nodepool.

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -0,0 +1,238 @@
...
+	return nil
+}
+
+func (nif *nodePortIsolationFilter) SetStorageWrapper(s cachemanager.StorageWrapper) error {
+	if s.Name() != disk.StorageName {
+		return fmt.Errorf("nodePortIsolationFilter can only support disk storage currently, cannot use %s", s.Name())
+	}
+
+	if len(nif.nodeName) == 0 {
+		return fmt.Errorf("node name for nodePortIsolationFilter is not ready")
```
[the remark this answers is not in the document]
rambohe-ch: fixed

pkg/yurthub/filter/nodeportisolation/filter.go
rambohe-ch: fixed

pkg/yurthub/filter/nodeportisolation/filter.go
rambohe-ch: fixed

pkg/yurthub/filter/nodeportisolation/filter.go
rambohe-ch: I have updated nodepool definition in #1183 as following:

The details of nodeport.openyurt.io/isolation annotation of service:

key: nodeport.openyurt.io/isolation

value: you can combine [np, -np, *] these three string to configure value, the priority is: [ np > -np > * ]
- np: enable specified NodePort service listening in this nodePool
- -np: disable specified NodePort service listening in this nodePool
- *: enable specified NodePort service listening in all nodePools except specified by -np.

for example:

*: enable NodePort service listening in all nodePools
nodepool1, nodepool2: enable NodePort service listening in nodePool1 and nodePool2
-nodepool1, *: enable NodePort service listening in all nodePools except nodePool1
nodepool1,-nodepool2: only enable NodePort service listening in nodePool1
nodepool1, *: enable NodePort service listening in all nodePools
nodepool1, -nodepool1, *: enable NodePort service listening in of nodePools except nodePool1

pkg/yurthub/filter/nodeportisolation/filter.go
rambohe-ch: fixed

rambohe-ch:
> Another problem, this filter seems cannot react to the nodepool change, for example the node is moved to another nodepool.

@Congrool I suppose that node should be reset and re-join when node want to change the nodePool. or maybe we can adapt nodepool change with a new controller in the other pull request.

Congrool:
> I suppose that node should be reset and re-join when node want to change the nodePool. or maybe we can adapt nodepool change with a new controller in the other pull request.

If so, we may need to add such instruction to our manual.

pkg/yurthub/filter/nodeportisolation/filter.go
Congrool: > nodepool1, -nodepool1, *: enable NodePort service listening in of nodePools except nodePool1

If [ np > -np > * ], I think it should enable NodePort service on all nodepools.

Is there a better way to indicate orphan nodes? Using `!` is hard to read I think.

rambohe-ch: @Congrool As discussed in community meeting, the design has been changed as following:

I will name the new filter as `nodeportisolation`, and add a new annotation `nodeport.openyurt.io/listen` for NodePort service in order to trigger `nodeportisolation` filter in Yurthub.

The details of `nodeport.openyurt.io/listen` annotation of service:
- key:    `nodeport.openyurt.io/listen`
- value:  A list of nodepool names that separated by ','
   *  `foo`:  enable specified NodePort service listening on nodes of NodePool named foo
   * `-foo`: disable specified NodePort service listening on nodes of NodePool named fool
   * `*`: enable specified NodePort service listening on nodes of all NodePools
   * if NodePool name is duplicated in the configuration, we will take the first configuration
   * if NodePool name is not configured, we will disable NodePort listening on nodes of these unconfigured NodePools.
   * orphan nodes(don't locate in NodePool) will be kept the same as native Kubernetes, so NodePort will be listened on orphan nodes.

     for example:
     - `foo, bar`:  enable NodePort service listening on nodes in foo and bar NodePool.
     - `foo, *`: enable NodePort service listening on nodes of all NodePools
     - `-foo, -bar`: disable NodePort service listening on nodes of all NodePools
     - `-foo, *`: disable NodePort service listening only on nodes in foo NodePool
     - `foo,-foo`: enable NodePort service listening on nodes in foo NodePool
     - `-foo`: disable NodePort service listening on nodes of all NodePools(include foo NodePool).

Congrool:
Refer to the design of https://github.com/openyurtio/openyurt/issues/1183#issuecomment-1418426832:

> if NodePool name is not configured, we will disable NodePort listening on nodes of these unconfigured NodePools.

When the svc has the annotation key but with no value, like `nodeport.openyurt.io/listen:`, I think it's more suitable to disable this svc in all nodepools. But currently, it will enable this svc in all nodepools. What do you think ? @rambohe-ch

rambohe-ch:
> Refer to the design of [#1183 (comment)](https://github.com/openyurtio/openyurt/issues/1183#issuecomment-1418426832):
> 
> > if NodePool name is not configured, we will disable NodePort listening on nodes of these unconfigured NodePools.
> 
> When the svc has the annotation key but with no value, like `nodeport.openyurt.io/listen:`, I think it's more suitable to disable this svc in all nodepools. But currently, it will enable this svc in all nodepools. What do you think ? @rambohe-ch

@Congrool you are right, we need to disable this svc in all nodepools if no value configured. i have fixed it, PTAL

PR #1210, opened 2023-02-08
add ignore package for dependabot

rambohe-ch:
/kind enhancement

add ignore package for dependabot

Fixes #

```release-note

```


Congrool:
How about also ingore etcd? We'd better keep the etcd dependencies same version as etcd we use in pool-coordinator.

rambohe-ch:
> How about also ingore etcd? We'd better keep the etcd dependencies same version as etcd we use in pool-coordinator.

@Congrool it's a good idea, done.

Congrool (commented):
/lgtm

Congrool (approved):

PR #1237, opened 2023-02-16
Fix: make rename replace old dir

LaurenceLiZhixin:
Now, 'Rename' function can't handler the condition that 'If file of newPath has already existed, it will be replaced.' , and error would occurs. We should delete newPath if it's existed.

### DIFF
--- a/pkg/yurthub/util/fs/store.go
+++ b/pkg/yurthub/util/fs/store.go
@@ -253,6 +253,11 @@ func (fs *FileSystemOperator) Rename(oldPath string, newPath string) error {
 	if !IfExists(oldPath) {
 		return ErrNotExists
 	}
+	if IfExists(newPath) {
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@LaurenceLiZhixin It seems unit tests have errors, please fix it.

pkg/yurthub/util/fs/store.go
```diff
@@ -253,6 +253,11 @@ func (fs *FileSystemOperator) Rename(oldPath string, newPath string) error {
 	if !IfExists(oldPath) {
 		return ErrNotExists
 	}
+	if IfExists(newPath) {
```
Congrool, line 256: We should also check if newPath is a dir.

rambohe-ch (approved):

PR #1248, opened 2023-02-18
feat: upgrade golangci-lint

fengshunli:

Fixes #

```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,6 @@ on:
 env:
   # Common versions
   GO_VERSION: '1.17'
-  GOLANGCI_VERSION: 'v1.40'
   DOCKER_BUILDX_VERSION: 'v0.4.2'
 
   # Common users. We can't run a step 'if secrets.AWS_USR != ""' but we can run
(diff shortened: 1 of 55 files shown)


.github/workflows/ci.yaml
```diff
@@ -44,7 +43,7 @@ jobs:
       - name: Lint golang code
         uses: golangci/golangci-lint-action@v3
         with:
-          version: ${{ env.GOLANGCI_VERSION }}
+          version: latest
```
rambohe-ch, line 46: why not version v1.51.1 is used here?

rambohe-ch:
@fengshunli please fix unit tests error. you can click the link: https://github.com/openyurtio/openyurt/actions/runs/4219426618/jobs/7324769786 to check the error message.

.github/workflows/ci.yaml
fengshunli: It is convenient to use latest without modifying it every time

rambohe-ch: @fengshunli In file `Makefile`, you will install golangci-lint version v1.51.1, but here latest version is used, so installed version is not used.

fengshunli: Makefile has been updated @rambohe-ch

Issue #1271, opened 2023-02-27
[feature request] add a new release job for pushing linux/amd64 images only


rambohe-ch:
**What would you like to be added**:
cross-platforms images are released in OpenYurt at present, but can not supported in snyk for security check. so we need to add a new job for creating unique linux/amd64 images. for example: openyurt/yurthub-amd64:vx.x.x

**Why is this needed**:
for supporting image security check

**others**
/kind feature

rambohe-ch:
because snyk is replaced by trivy, so we don't need to push linux/amd64 images independently. so i will close this issue.

PR #1272, opened 2023-02-27
add arm64 qemu build

fengshunli:

Fixes #

https://github.com/openyurtio/openyurt/issues/1271

```release-note

```


rambohe-ch:
@fengshunli The current `docker-push` is used for releasing images for multiple platforms. so we should not modify it. how about add a new target named `docker-push-for-snyk` for releasing linux/amd64 images?

fengshunli:
> @fengshunli The current `docker-push` is used for releasing images for multiple platforms. so we should not modify it. how about add a new target named `docker-push-for-snyk` for releasing linux/amd64 images?

The images pushed by docker-push must be supported, or only scan openyurt itself @rambohe-ch

PR #1281, opened 2023-03-01
Chore: add static check to golangci-lint

Lan-ce-lot:
I add static check to golangci-lint and fixed some of the following problems

image

Fixes #

```release-note

```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -3,7 +3,6 @@
 
 # options for analysis running
 run:
-  go: '1.16'
   # default concurrency is a available CPU number
   concurrency: 4
 
@@ -39,10 +38,11 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - deadcode 
+    - gci
     - gofmt
     - goimports
-    - gci
     - ineffassign
     - misspell
+    - unused
     - vet
+    - staticcheck
--- a/pkg/yurthub/poolcoordinator/resources/resources.go
+++ b/pkg/yurthub/poolcoordinator/resources/resources.go
@@ -35,7 +35,6 @@ import (
 )
 
 type PoolScopeResourcesManger struct {
-	factory                      informers.SharedInformerFactory
 	validPoolScopedResources     map[string]*verifiablePoolScopeResource
 	validPoolScopedResourcesLock sync.RWMutex
 	k8sClient                    kubernetes.Interface
--- a/pkg/yurthub/proxy/util/util.go
+++ b/pkg/yurthub/proxy/util/util.go
@@ -198,17 +198,11 @@ func WithIfPoolScopedResource(handler http.Handler) http.Handler {
 type wrapperResponseWriter struct {
 	http.ResponseWriter
 	http.Flusher
-	http.CloseNotifier
 	http.Hijacker
 	statusCode int
 }
 
 func newWrapperResponseWriter(w http.ResponseWriter) *wrapperResponseWriter {
-	cn, ok := w.(http.CloseNotifier)
-	if !ok {
-		klog.Error("can not get http.CloseNotifier")
-	}
-
 	flusher, ok := w.(http.Flusher)
 	if !ok {
 		klog.Error("can not get http.Flusher")
@@ -222,7 +216,6 @@ func newWrapperResponseWriter(w http.ResponseWriter) *wrapperResponseWriter {
 	return &wrapperResponseWriter{
 		ResponseWriter: w,
 		Flusher:        flusher,
-		CloseNotifier:  cn,
 		Hijacker:       hijacker,
 	}
 }
--- a/pkg/yurthub/transport/transport.go
+++ b/pkg/yurthub/transport/transport.go
@@ -136,6 +136,7 @@ func (tm *transportManager) start() {
 		} else if lastCert != nil && curr != nil {
 			if lastCert == curr {
 				// cert is not rotate, just wait
+				klog.V(5).Infof("cert is not rotate, just wait")
 			} else {
 				// cert rotated
 				klog.Infof("cert rotated, so close all client connections for loading new cert")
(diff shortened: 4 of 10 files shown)


fengshunli (changes requested):
image
remove

pkg/yurthub/transport/transport.go
```diff
@@ -134,6 +134,7 @@ func (tm *transportManager) start() {
 			tm.closeAll()
 			lastCert = curr
 		} else if lastCert != nil && curr != nil {
+			//nolint:staticcheck
 			if lastCert == curr {
 				// cert is not rotate, just wait
 			} else {
```
YTGhost, line 140: Maybe we can avoid using the `//nolint:staticcheck` comment by using klog to output log in the empty if/else branch(we can set this log level to 5)?

YTGhost:
Looks like `codecov/patch` is failed?

Lan-ce-lot:
> 
But go type is a string.
```
        "go": {
          "description": "Targeted Go version.",
          "type": "string",
          "default": "1.17"
        }
```

fengshunli:
Refer to the ci of lint, it can be seen that this configuration is redundant，see https://github.com/golangci/golangci-lint-action/tree/v3

Lan-ce-lot:
> > > Default: use Go version from the go.mod file, fallback on the env var `GOVERSION`, fallback on 1.18
> 
> Refer to the ci of lint, it can be seen that this configuration is redundant

ok, I will remove it

pkg/yurthub/transport/transport.go
Lan-ce-lot: agree

pkg/yurthub/transport/transport.go
```diff
@@ -145,6 +146,7 @@ func (tm *transportManager) start() {
 		} else {
 			// lastCet != nil && curr == nil
 			// certificate expired or deleted unintentionally, just wait for cert updated by bootstrap config, do nothing
+			klog.V(5).Infof("certificate expired or deleted unintentionally")
```
Congrool, line 149: It's better to use Warning I think.

pkg/yurthub/poolcoordinator/resources/resources.go
```diff
@@ -35,6 +35,7 @@ import (
 )
 
 type PoolScopeResourcesManger struct {
+	//nolint:unused
 	factory                      informers.SharedInformerFactory
```
Congrool, line 39: I think the `factory` field here can be directly removed.

.golangci.yaml
```diff
@@ -39,10 +38,11 @@ linters-settings:
 linters:
   disable-all: true
   enable:
-    - deadcode
```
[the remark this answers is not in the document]
Lan-ce-lot: The linter 'deadcode' is deprecated since 14.9.8 and I use 'unused' to replace it.
https://github.com/golangci/golangci-lint/issues/1841

pkg/yurthub/poolcoordinator/resources/resources.go
Lan-ce-lot: Ok, I will remove it

pkg/yurttunnel/server/interceptor_test.go
```diff
@@ -66,10 +66,10 @@ func TestGetResponse(t *testing.T) {
...
 	}
 
-	if diffBytes(rbytes, wbytes) {
-		//t.Errorf("raw bytes is not equal\n")
-	}
-
+	//if diffBytes(rbytes, wbytes) {
+	//	t.Errorf("raw bytes is not equal\n")
+	//}
+	diffBytes(rbytes, wbytes)
```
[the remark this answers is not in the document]
Lan-ce-lot: Will fail, I think should be changed to this:
```golang
if !diffBytes(rbytes, wbytes) {
    t.Errorf("raw bytes is not equal\n")
}
```
Because `diffBytes` in the case of equal return `true`.

pkg/yurttunnel/trafficforward/dns/util_test.go
```diff
@@ -231,26 +231,26 @@ func TestAddOrUpdateRecord(t *testing.T) {
...
-			expect:  []string{"10.1.218.68\tk8s-xing-61"},
+			expect:  []string{"10.1.10.62\tk8s-xing-61"},
 		},
+
+		//{
+		//	desc:    "test idempotence",
+		//	records: []string{"10.1.218.68\tk8s-xing-61"},
+		//	record:  "10.1.10.62\tk8s-xing-61",
+		//	expect:  []string{"10.1.218.68\tk8s-xing-61"},
+		//},
```
Congrool, line 253: I don't know what is the purpose of this test case "test idempotence" 😕 . Maybe we can just delete the case.

Lan-ce-lot: Ok

rambohe-ch:
@Lan-ce-lot Very appreciate for your work to improve golangci-lint. it's a good job to add staticcheck for golangci-lint.

pkg/yurthub/proxy/util/util.go
```diff
@@ -198,17 +198,11 @@ func WithIfPoolScopedResource(handler http.Handler) http.Handler {
 type wrapperResponseWriter struct {
 	http.ResponseWriter
 	http.Flusher
-	http.CloseNotifier
 	http.Hijacker
 	statusCode int
 }
 
 func newWrapperResponseWriter(w http.ResponseWriter) *wrapperResponseWriter {
-	cn, ok := w.(http.CloseNotifier)
```
rambohe-ch, line 207: because Yurthub depends on `httputil.ReverseProxy` for proxying requests, and i have checked golang v1.18.10, `http.CloseNotifier` interface will be used for closing request when context done.
https://cs.opensource.google/go/go/+/refs/tags/go1.18.10:src/net/http/httputil/reverseproxy.go;l=221-233

so it's not a good idea to remove `http.CloseNotifier` field from wrapperResponseWriter, and we need to find a workaround.

Lan-ce-lot: Yes, this is a problem.

rambohe-ch: I have checked that request.Context is recommended to use for replacing http.CloseNotifier,  it's no problem to remove `http.CloseNotifier`. so i will merge this pull request.

rambohe-ch (approved):

PR #1312, opened 2023-03-16
【FIX】fix pool-coordinator running error when some error happened and abort…

JameKeal:
… parameter init

/kind bug

For example, in the first place where continue is added and executed, there is no way to initialize the parameter, and then you have to wait for statusChan to fetch the value again before you can initialize it again.

However, for an identity like leader, it will send to electorStatus only after it grabs the lock every time. If it is always the leader and the first initialization fails, then there is almost no chance to initialize the variable later. I wonder if there is any problem with my understanding in this way.

89A19270-EDD2-4296-8230-38373B27A0CF

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/poolcoordinator/coordinator.go
+++ b/pkg/yurthub/poolcoordinator/coordinator.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strconv"
 	"sync"
 	"time"
@@ -81,6 +82,11 @@ type Coordinator interface {
 	IsHealthy() (cachemanager.CacheManager, bool)
 }
 
+type statusInfo struct {
+	electorStatus int32
+	currentCnt    uint64
+}
+
 type coordinator struct {
 	sync.Mutex
 	ctx               context.Context
@@ -95,6 +101,8 @@ type coordinator struct {
 	etcdStorage       storage.Store
 	hubElector        *HubElector
 	electStatus       int32
+	cnt               uint64
+	statusInfoChan    chan statusInfo
 	isPoolCacheSynced bool
 	certMgr           *certmanager.CertManager
 	// cloudCAFilePath is the file path of cloud kubernetes cluster CA cert.
@@ -155,6 +163,7 @@ func NewCoordinator(
 		serializerMgr:      cfg.SerializerManager,
 		restMapperMgr:      cfg.RESTMapperManager,
 		hubElector:         elector,
+		statusInfoChan:     make(chan statusInfo, 10),
 	}
 
 	poolCacheSyncedDetector := &poolCacheSyncedDetector{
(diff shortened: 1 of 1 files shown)


pkg/yurthub/poolcoordinator/coordinator.go
```diff
@@ -95,6 +100,8 @@ type coordinator struct {
 	etcdStorage       storage.Store
 	hubElector        *HubElector
 	electStatus       int32
+	cnt               int32
```
Congrool, line 103: How about using uint64?

pkg/yurthub/poolcoordinator/coordinator.go
```diff
@@ -231,7 +239,21 @@ func (coordinator *coordinator) Run() {
 			}
 			metrics.Metrics.ObservePoolCoordinatorYurthubRole(electorStatus)
 
-			switch electorStatus {
+			coordinator.cnt++
```
Congrool, line 242: cnt may overflow, we'd better be aware of such case

Congrool: Any improvement?

pkg/yurthub/poolcoordinator/coordinator.go
```diff
@@ -231,7 +239,21 @@ func (coordinator *coordinator) Run() {
...
+			coordinator.statusInfoChan <- statusInfo{
+				electorStatus: electorStatus,
+				currentCnt:    coordinator.cnt,
+			}
+		case electorStatusInfo, ok := <-coordinator.statusInfoChan:
+			if !ok {
+				return
+			}
+			if electorStatusInfo.currentCnt < coordinator.cnt {
+				klog.Warningf("electorStatusInfo %+v is behind of current cnt %d", electorStatusInfo, coordinator.cnt)
```
Congrool, line 252: It's the case that we expect, I think we do not need to warning it.

pkg/yurthub/poolcoordinator/coordinator.go
```diff
@@ -245,20 +267,24 @@ func (coordinator *coordinator) Run() {
...
 					continue
 				}
 				klog.Infof("coordinator poolCacheSyncManager has ensure started")
+
+				nodeLeaseProxyClient, err := coordinator.newNodeLeaseProxyClient()
+				if err != nil {
+					klog.Errorf("cloud not get cloud lease client when becoming leader yurthub, %v", err)
+					coordinator.statusInfoChan <- electorStatusInfo
+					continue
+				}
```
Congrool, line 287: Why we move `newNodeLeaseProxyClient` to here?

JameKeal: nodeLeaseProxyClient is used by coordinator.delegateNodeLeaseManager.EnsureStartWithHandler, after the move from the code structure looks like a smooth logic, it`s a code block.

pkg/yurthub/poolcoordinator/coordinator.go
```diff
@@ -231,7 +240,37 @@ func (coordinator *coordinator) Run() {
...
+					for v := range coordinator.statusInfoChan {
+						klog.Infof("clean statusInfo data %+v when coordinator.cnt is reset", v)
+						i++
+						if i == length {
+							break
+						}
+					}
+				}
+			}
+			coordinator.cnt++
```
Congrool, line 259: Just a small bug(even not a bug currently).
If coordinator.cnt == math.MaxUint64, it will be firstly set as 0, and then self increase here. The cnt of final enqueue status is 1, 0 is skipped.

Congrool:
@JameKeal Would you update your pr?

JameKeal:
> @JameKeal Would you update your pr?

nothing, From the beginning when it was designed to be used, it didn't include 0, it was just 1

rambohe-ch (approved):

PR #1314, opened 2023-03-17
feat: add yurtadm renew certificate command

### LINKED ISSUE #1290: [feature request]improve bootstrap configuration for YurtHub
A bootstrap token should be passed to yurthub(--join-token parameter) for bootstrapping. sometime token without expire date is used for convenient,  so security problem maybe happen because the bootstrap token is always exist on the edge node.

so i think we can improve bootstrap configuration for yurthub, a bootstrap file instead of bootstrap token is passed into YurtHub, and bootstrap file will be removed after Yurthub bootstrap completed, so bootstrap token can not be leaked on the edge node.

the work we need do as following:
- YurtHub: 
      add a new paramter --boostrap-file, and use bootstrap file to bootstrap YurtHub, and we can keep the original `--join-token` parameter for several versions.

- Yurtadm: 
      - Yurtadm should create a bootstrap file for YurtHub boostrapping, and remove this file after YurtHub bootstrapping completed.
      - add a new sub command for Yurtadm to create a temp bootstrap file for YurtHub when YurtHub need to use bootstrap file to update certificate if certificate expired. on the other hand, the temp bootstrap file also should be removed after certificate is updated.

**Why is this needed**:
A bootstrap token should be passed to yurthub(--join-token parameter) for bootstrapping. sometime token without expire date is used for convenient,  so security problem maybe happen because the bootstrap token is always exist on the edge node.

**others**
/kind feature

YTGhost:
#### What type of PR is this?
/kind feature

1. add yurtadm renew certificate command
2. gather the yurthub-related functions into `pkg/yurtadm/util/yurthub/yurthub.go`
3. gather `CheckKubeletStatus` function into `pkg/yurtadm/util/kubernetes/kubernetes.go`
4. When building a test cluster, yurthub is also started with a bootstrap file

Fixes #1290 

/assign @rambohe-ch 

```release-note

```

### DIFF
--- a/pkg/yurtadm/cmd/renew/certificate/certificate.go
+++ b/pkg/yurtadm/cmd/renew/certificate/certificate.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2023 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package certificate
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"k8s.io/klog/v2"
+
+	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	yurtconstants "github.com/openyurtio/openyurt/pkg/yurtadm/constants"
+	"github.com/openyurtio/openyurt/pkg/yurtadm/util/yurthub"
+)
+
+type certificateOptions struct {
+	token                    string
+	caCertHashes             []string
+	unsafeSkipCAVerification bool
+	serverAddr               string
+	yurthubServer            string
+}
+
+// newCertificateOptions returns a struct ready for being used for creating cmd renew flags.
+func newCertificateOptions() *certificateOptions {
+	return &certificateOptions{
+		caCertHashes:             make([]string, 0),
+		unsafeSkipCAVerification: true,
+		yurthubServer:            yurtconstants.DefaultYurtHubServerAddr,
+	}
+}
+
+// NewCmdCertificate returns "yurtadm renew certificate" command.
+func NewCmdCertificate() *cobra.Command {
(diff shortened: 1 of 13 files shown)


pkg/yurtadm/constants/constants.go
```diff
@@ -220,6 +223,7 @@ spec:
     - --server-addr={{.kubernetesServerAddr}}
     - --node-name=$(NODE_NAME)
     - --join-token={{.joinToken}}
```
rambohe-ch, line 225: please don't left `join-token` in Yurthub.yaml.

pkg/yurtadm/cmd/renew/certificate/certificate.go
```diff
@@ -0,0 +1,155 @@
...
+	o := newCertificateOptions()
+
+	certificateCmd := &cobra.Command{
+		Use:   "certificate",
+		Short: "Create bootstrap file for yurthub to update certificate",
+		RunE: func(certificateCmd *cobra.Command, args []string) error {
+			if err := o.validate(); err != nil {
+				klog.Fatalf("validate options: %v", err)
+			}
+
```
rambohe-ch, line 61: I think we should verify the certificates of Yurthub expired or not at first. and if the certificate is valid, we can exit `yurtadm renew certificate` directly with some logs.

pkg/yurthub/certificate/token/token.go
```diff
@@ -274,6 +277,9 @@ func (ycm *yurtHubCertManager) getPkiDir() string {
 
 // getBootstrapConfFile returns the path of yurthub bootstrap conf file
 func (ycm *yurtHubCertManager) getBootstrapConfFile() string {
+	if len(ycm.bootstrapFile) != 0 {
+		return ycm.bootstrapFile
+	}
```
rambohe-ch, line 282: we need to check all codes that invoke `getBootstrapConfFile()` function can work or not if Yurthub restart and bootstrap file doesn't exist. 

The following places can not work:
1. verifyServerAddrOrCleanup()
2. prepareConfigAndCaFile()
3. generateCertClientFn()

pkg/yurtadm/constants/constants.go
YTGhost: Fixed

pkg/yurtadm/cmd/renew/certificate/certificate.go
YTGhost: Fixed

YTGhost:
@rambohe-ch PTAL

rambohe-ch (approved):

rambohe-ch:
@YTGhost this pull request depends on the #1333 , we need to rebase after #1333 merge.

YTGhost:
@rambohe-ch PTAL

rambohe-ch (approved):

PR #1320, opened 2023-03-21
Fix: remove deprecated rand

fengshunli:

that need a specific result sequence should use NewRand(NewSource(seed))
image

Fixes #

```release-note

```

### DIFF
--- a/cmd/yurt-controller-manager/controller-manager.go
+++ b/cmd/yurt-controller-manager/controller-manager.go
@@ -38,7 +38,8 @@ import (
 )
 
 func main() {
-	rand.Seed(time.Now().UnixNano())
+	newRand := rand.New(rand.NewSource(time.Now().UnixNano()))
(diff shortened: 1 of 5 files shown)


cmd/yurt-controller-manager/controller-manager.go
```diff
@@ -38,7 +38,8 @@ import (
 )
 
 func main() {
-	rand.Seed(time.Now().UnixNano())
+	newRand := rand.New(rand.NewSource(time.Now().UnixNano()))
+	newRand.Seed(time.Now().UnixNano())
```
rambohe-ch, line 42: would you be able to explain the reason of change rand seed?

fengshunli: see desc

rambohe-ch: good job.

rambohe-ch (approved):

Issue #1197, opened 2023-02-08
[feature request] `NodePool Governance Capability`: pool-coordinator controller need to deal with a sence that nodepool disconnect with cloud


JameKeal:
**What would you like to be added**:
pool-coordinator controller need to deal a sence when nodepool disconnect with cloud, the node in nodepool can be autonomy, the pod in node cannot be evict.

- node actually down, close the node autonomy, the pod binding controller cannot use, the pod in node can be evict.
- nodepool disconnect cloud, controller can add node autonomy for nodepool, then the pod binding controller will working, k8s cannot evict the pod.

**Why is this needed**:
On the current situation, when nodepool disconnect with cloud, if node does not open autonomy, the pod in node wil be evict. In another way, if the node open autonomy, when node actually down, the pod binding controller will not evict the pod, it`s different from nodepool governance.

**others**
/kind feature

rambohe-ch:
@JameKeal Thank you for raising issue. yes, we need to consider how to handle pods eviction when the network between cloud and all nodes in nodePool disconnected. I think we need to keep the pods un-evicted at that time.

rambohe-ch:
@gnunu @LindaYu17 do you have any comments?

AndyEWang:
Any progress? Should nodelifecycle be considered since zone disruption can make the minimal change to handle this case. (refer to partialDisruption and FullDisruption in kube node lifecycle controller)

rambohe-ch:
> Any progress? Should nodelifecycle be considered since zone disruption can make the minimal change to handle this case. (refer to partialDisruption and FullDisruption in kube node lifecycle controller)

@AndyEWang Do you have any suggestions how to integrate `original partialDisruption and FullDisruption` into `NodePool` in OpenYurt?

gnunu:
Remember we talked about this issue during our earlier discussion about pool coordinator design, then we decided to let apiserver decide what to do when whole pool is lost. 

However it's natural that pods should not be evicted on pool disconnection occurring. Anyway pool should be in automate mode if pods are pool binding. So the implementation should take care of this requirement.

How about we add this capability during our migration to yurt-controller?

AndyEWang:
Personally, OpenYurt can take charge of node lifecycle like old versions before v1.2. If not, we must find out many workarounds against kube-controller-manager. Should OpenYurt give an option to let users decide which node lifecycle to use?

rambohe-ch:
[...]

PR #1322, opened 2023-03-21
add nodepool autonomy into yurt-manager

JameKeal:
/kind feature

On the current situation, when nodepool disconnect with cloud, if node does not open autonomy, the pod in node wil be evict. In another way, if the node open autonomy, when node actually down, the pod binding controller will not evict the pod, it`s different from nodepool governance.

Fixes https://github.com/openyurtio/openyurt/issues/1197

/assign @rambohe-ch

```release-note

```


rambohe-ch:
@JameKeal please fix golangci-lint errors.

fengshunli (changes requested):
```suggestion
// Cluster implements a validating and defaulting webhook for Cluster.
```

PodHandler

pkg/webhook/pod/v1/pod_validation.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func validatePodDeletion(cli client.Client, pod *v1.Pod) field.ErrorList {
+	nodeName := pod.Spec.NodeName
+	if nodeName == "" {
+		return nil
+	}
+
+	node := v1.Node{}
+	if err := cli.Get(context.TODO(), client.ObjectKey{Name: nodeName}, &node); err != nil {
```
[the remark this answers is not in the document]
JameKeal: That client has a cache, once you request a resource, it parses it out of scheme, it automatically triggers the informer thing, and then it goes into the cache and then your get goes out of the cache.

The controller-runtime code show as follows.
image

image

pkg/webhook/pod/v1/pod_default.go
```diff
@@ -0,0 +1,37 @@
+/*
```
rambohe-ch, line 1: It seems that `mutate webhook` is not needed for nodepool autonomy, so how about remove this webhook for improving webhook efficiency?

pkg/webhook/pod/v1/pod_handler.go
```diff
@@ -0,0 +1,54 @@
...
+	v1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+
+	"github.com/openyurtio/openyurt/pkg/webhook/builder"
+	"github.com/openyurtio/openyurt/pkg/webhook/util"
+)
+
+// SetupWebhookWithManager sets up Cluster webhooks. 	mutate path, validatepath, error
```
rambohe-ch, line 29: we need to add some RBAC rules for this webhook, for example: list/watch/get nodepool, node, lease resource.

rambohe-ch: ok

pkg/webhook/pod/v1/pod_handler.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return util.GenerateMutatePath(gvk),
+		util.GenerateValidatePath(gvk),
+		builder.WebhookManagedBy(mgr).
+			For(&v1.Pod{}).
+			WithValidator(webhook).
+			Complete()
+}
+
+// +kubebuilder:webhook:path=/validate-core-openyurt-io-pod,mutating=false,failurePolicy=fail,sideEffects=None,admissionReviewVersions=v1;v1beta1,groups=core.openyurt.io,resources=pods,verbs=delete,versions=v1,name=validate.core.v1.pod.openyurt.io
```
rambohe-ch, line 46: path should be `/validate-core-openyurt-io-v1-pod` instead of `/validate-core-openyurt-io-pod`

charts/openyurt/templates/yurt-manager-auto-generated.yaml
```diff
@@ -804,3 +804,22 @@ webhooks:
     resources:
     - nodepools
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: kube-system
+      path: /validate-core-openyurt-io-pod
```
rambohe-ch, line 813: path should be /validate-core-openyurt-io-v1-pod instead of /validate-core-openyurt-io-pod

rambohe-ch (approved):

PR #1333, opened 2023-03-27
support pass bootstrap-file to yurthub

### LINKED ISSUE #1290: [feature request]improve bootstrap configuration for YurtHub

rambohe-ch:
/kind feature

1. support pass bootstrap file instead of join token for bootstrapping yurthub in order to improve security of yurthub. so yurtadm join will not leave `bootstrap token` on the edge node.
2. parameter `--join-token` is deprecated in yurthub, and plan to remove in OpenYurt version v1.5.
3.  add `/v1/readyz` endpoint for yurthub, add used by yurtadm command to check certificates are ready or not.

Fixes #1290 

```release-note

```

### DIFF
--- a/pkg/yurthub/certificate/token/token.go
+++ b/pkg/yurthub/certificate/token/token.go
@@ -68,6 +68,7 @@ type CertificateManagerConfiguration struct {
 	RootDir                  string
 	NodeName                 string
 	JoinToken                string
+	BootstrapFile            string
 	CaCertHashes             []string
 	YurtHubCertOrganizations []string
 	CertIPs                  []net.IP
@@ -86,6 +87,7 @@ type yurtHubCertManager struct {
 	hubRunDir                  string
 	hubName                    string
 	joinToken                  string
+	bootstrapFile              string
 	dialer                     *util.Dialer
 }
 
@@ -104,6 +106,7 @@ func NewYurtHubCertManager(cfg *CertificateManagerConfiguration) (hubCert.YurtCe
 		hubRunDir:     hubRunDir,
 		hubName:       projectinfo.GetHubName(),
 		joinToken:     cfg.JoinToken,
+		bootstrapFile: cfg.BootstrapFile,
 		caCertHashes:  cfg.CaCertHashes,
 		dialer:        util.NewDialer("hub certificate manager"),
 	}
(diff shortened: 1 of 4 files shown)


pkg/yurthub/certificate/token/token.go
```diff
@@ -274,6 +328,9 @@ func (ycm *yurtHubCertManager) getPkiDir() string {
 
 // getBootstrapConfFile returns the path of yurthub bootstrap conf file
 func (ycm *yurtHubCertManager) getBootstrapConfFile() string {
+	if len(ycm.bootstrapFile) != 0 {
+		return ycm.bootstrapFile
+	}
```
YTGhost, line 333: `generateCertClientFn` will invoke `getBootstrapConfFile()` function, dose it work well if yurthub restart and bootstrap file doesn't exist?

rambohe-ch: yes, we need to use bootstrap file to apply certificates when certificates expired, so getBootstrapConfFile() is used in generateCertClientFn, and yurthub will always wait for bootstrap file until it is prepared by `yurtadm renew certificate` command.

PR #1346, opened 2023-03-31
【FIX】fix node not-ready condition when leader disconnect with cloud

JameKeal:
/kind bug

When node ready condition is unknown, the native kubernetes will set node.kubernetes.io/unreachable taints in node,
and the pod will be evict after 300s, that's not what we're trying to do in delegate lease.
Up to now, it's only happen when leader in nodePool is disconnected with cloud, and this node will be not-ready,
because in an election cycle, the node lease will not delegate to cloud, after 40s, the kubernetes will set unknown.

Fixes #

@rambohe-ch 

```release-note

```

### DIFF
--- a/charts/openyurt/templates/yurt-controller-manager.yaml
+++ b/charts/openyurt/templates/yurt-controller-manager.yaml
@@ -127,6 +127,12 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - nodes/status
+    verbs:
+      - update
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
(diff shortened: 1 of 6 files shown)


charts/openyurt/templates/yurt-controller-manager.yaml
```diff
@@ -127,6 +127,12 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - nodes/status
+    verbs:
+      - update
```
rambohe-ch, line 135: please also add these roles for yurt-manager component.

rambohe-ch: I think i can close this comment. because these roles will be added for yurt-manager when @gnunu migrates this controller into yurt-manager.

rambohe-ch (approved):

Issue #1341, opened 2023-03-30
[feature request]support yurthub component work in specified namespace


rambohe-ch:
**What would you like to be added**:
yurthub component should be deployed in kube-system namespace. as you known, kube-system namespace is used for K8s components, so some users have requested that yurthub can be deployed in specified namespace.

**Why is this needed**:
improve the deployment flexibility of OpenYurt component.

**others**
/kind feature

YTGhost:
Does this mean we need to add a new flag named `ResourceNamespace` to `yurthub` and set the value to `LeaderElection.ResourceNamespace`? Is there anything else that needs to be changed?

rambohe-ch:
> Does this mean we need to add a new flag named `ResourceNamespace` to `yurthub` and set the value to `LeaderElection.ResourceNamespace`? Is there anything else that needs to be changed?

@YTGhost I think that one component works in a specified namespace, the following two jobs maybe included:
1. the deployment of this component should be improved, the namespace for this component can be specified in the helm values.yaml or command tool(like yurtadm) parameter.
2. improve `kube-system` hard code in the component, and the namespace can be specified as a parameter for the component.

YTGhost:
> > Does this mean we need to add a new flag named `ResourceNamespace` to `yurthub` and set the value to `LeaderElection.ResourceNamespace`? Is there anything else that needs to be changed?
> 
> @YTGhost I think that one component works in a specified namespace, the following two jobs maybe included:
> 
> 1. the deployment of this component should be improved, the namespace for this component can be specified in the helm values.yaml or command tool(like yurtadm) parameter.
> 2. improve `kube-system` hard code in the component, and the namespace can be specified as a parameter for the component.

I see. Maybe I can take this work after migrate the service topology controller.

huweihuang:
may I take this work? @rambohe-ch

huweihuang:
if we set a specified namespace for yurthub and the other openyurt component, What is the appropriate name of the namespace?  can use `openyurt` as default namespace which been set in yurthub.yaml? @rambohe-ch

rambohe-ch:
> if we set a specified namespace for yurthub and the other openyurt component, What is the appropriate name of the namespace? if the user didn't set the specified namespace , can use `openyurt` as default namespace which been set in yurthub.yaml? @rambohe-ch

@huweihuang I think that default namespace should be `kube-system`, so we keep the consistency with the old version.

huweihuang:
[...]

PR #1355, opened 2023-04-03
feat: support pool-coordinator component work in specified namespace

### LINKED ISSUE #1342: [feature request]support pool-coordinator component work in specified namespace
**What would you like to be added:**
pool-coordinator component should be deployed in kube-system namespace. as you known, kube-system namespace is used for K8s components, so some users have requested that OpenYurt components(like yurthub, pool-coordinator) can be deployed in specified namespace.

**Why is this needed:**
improve the deployment flexibility of OpenYurt components.

**others**
/kind feature

y-ykcir:
#### What type of PR is this?

> /kind feature

Fixes #1342 

/assign @rambohe-ch 

add a parameter for yurt-manager to specify namespace variable for PoolcoordinatorNS
modified helm: use value to specify namespace in pool-coordinator.yaml and add startup parameter in yurt-manager.yaml

### DIFF
--- a/charts/openyurt/templates/yurt-manager.yaml
+++ b/charts/openyurt/templates/yurt-manager.yaml
@@ -57,6 +57,7 @@ spec:
             - --health-probe-addr=:{{ .Values.yurtManager.healthProbe.port }}
             - --logtostderr=true
             - --v={{ .Values.yurtManager.log.level }}
+            - --working-namespace={{ .Release.Namespace | quote }}
           command:
             - /usr/local/bin/yurt-manager
           image: {{ .Values.yurtManager.image.repository }}:{{ .Values.yurtManager.image.tag }}
--- a/cmd/yurt-manager/app/options/generic.go
+++ b/cmd/yurt-manager/app/options/generic.go
@@ -36,6 +36,7 @@ func NewGenericOptions() *GenericOptions {
 			LeaderElectionNamespace: "kube-system",
 			RestConfigQPS:           30,
 			RestConfigBurst:         50,
+			WorkingNamespace:        "kube-system",
 		},
 	}
 }
@@ -61,9 +62,10 @@ func (o *GenericOptions) ApplyTo(cfg *config.GenericConfiguration) error {
 
 	cfg.HealthProbeAddr = o.HealthProbeAddr
 	cfg.EnableLeaderElection = o.EnableLeaderElection
-	cfg.LeaderElectionNamespace = o.LeaderElectionNamespace
+	cfg.LeaderElectionNamespace = o.WorkingNamespace
 	cfg.RestConfigQPS = o.RestConfigQPS
 	cfg.RestConfigBurst = o.RestConfigBurst
+	cfg.WorkingNamespace = o.WorkingNamespace
 
 	return nil
 }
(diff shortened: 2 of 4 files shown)


cmd/yurt-manager/app/options/options.go
```diff
@@ -24,18 +24,20 @@ import (
 
 // YurtManagerOptions is the main context object for the yurt-manager.
 type YurtManagerOptions struct {
-	Generic            *GenericOptions
-	NodePoolController *NodePoolControllerOptions
-	GatewayController  *GatewayControllerOptions
+	Generic                       *GenericOptions
+	NodePoolController            *NodePoolControllerOptions
+	GatewayController             *GatewayControllerOptions
+	PoolCoordinatorCertController *PoolCoordinatorCertControllerOptions
```
rambohe-ch, line 30: I think that all controllers and webhooks in yurt-manager will use this parameter for specifying namespace, so how about add a field `WorkingNamespace` in  field `GenericOptions` instead of adding `PoolCoordinatorCertControllerOptions `?

charts/openyurt/templates/pool-coordinator.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: pool-coordinator-apiserver
-  namespace: {{ .Release.Namespace | quote }}
+  namespace: {{ .Values.poolCoordinator.namespace | quote }}
```
rambohe-ch, line 5: I think the target is all components will be installed in the same namespace, so it is convenient to use `.Release.Namespace` to specify namespace for all components, and we don't need a unique namespace for each component.

charts/openyurt/templates/yurt-manager.yaml
```diff
@@ -57,6 +57,7 @@ spec:
             - --health-probe-addr=:{{ .Values.yurtManager.healthProbe.port }}
             - --logtostderr=true
             - --v={{ .Values.yurtManager.log.level }}
+            - --pool-coordinator-namespace={{ .Values.poolCoordinator.namespace | quote }}
```
rambohe-ch, line 60: .Values.poolCoordinator.namespace --> .Release.Namespace

charts/openyurt/values.yaml
```diff
@@ -16,6 +16,7 @@ poolCoordinator:
   apiserverResources:
     requests:
       cpu: 250m
+  namespace: kube-system
```
rambohe-ch, line 19: remove this unique namespace definition.

cmd/yurt-manager/app/options/options.go
```diff
@@ -24,18 +24,20 @@ import (
...
 }
 
 // NewYurtManagerOptions creates a new YurtManagerOptions with a default config.
 func NewYurtManagerOptions() (*YurtManagerOptions, error) {
 
 	s := YurtManagerOptions{
-		Generic:            NewGenericOptions(),
```
rambohe-ch, line 36: Default of field `LeaderElectionNamespace` in `Generic` is `kube-system`, so we also need to adapt this filed for `WorkingNamespace` parameter.

cmd/yurt-manager/app/options/options.go
y-ykcir: Okay I understand, what I thought before is that the namespace of poordinator can be specified separately, because poordinator and yurt manager are separate components. I will rework these.

rambohe-ch: @y-ykcir Sorry for i didn't make it clear in the issue before, and thanks for your understanding.

y-ykcir: @rambohe-ch Should I adapt all the components of yurt-manager to `working-namespace` in this pr?

rambohe-ch: @y-ykcir i think that we can add a issue for tracking yurt-manager(you can add if you want to). so you can handle pool-coordinator cert controller in this issue.

rambohe-ch (approved):

y-ykcir:
@rambohe-ch The completion of the poolcoordinator should also require synchronization changes in the yurthub like [here](	https://github.com/openyurtio/openyurt/blob/f531e4dfa3871e02a061f3668135c2e0f69179e3/pkg/yurthub/poolcoordinator/constants/constants.go#L33), which may have to be done in the #1341

y-ykcir:
@rambohe-ch The completion of the poolcoordinator should also require synchronization changes in the yurthub like [here](	https://github.com/openyurtio/openyurt/blob/f531e4dfa3871e02a061f3668135c2e0f69179e3/pkg/yurthub/poolcoordinator/constants/constants.go#L33), which may have to be done in the #1341

rambohe-ch:
> @rambohe-ch The completion of the poolcoordinator should also require synchronization changes in the yurthub like [here](https://github.com/openyurtio/openyurt/blob/f531e4dfa3871e02a061f3668135c2e0f69179e3/pkg/yurthub/poolcoordinator/constants/constants.go#L33), which may have to be done in the #1341

@y-ykcir yes, i think @huweihuang will handle it.

PR #1359, opened 2023-04-07
add a new way to create webhook

JameKeal:
/kind enhancement

At present, the webhook builder not support any other information, for example, like request info, we will use this parameter to some scenes, like use it to judge source of request, it can make our webhook more stronger and rigorous.

So we need to add other builder to create webhook.

Fixes #

```release-note

```

### DIFF
--- a/hack/make-rules/add_controller.sh
+++ b/hack/make-rules/add_controller.sh
@@ -36,6 +36,7 @@ function usage(){
     echo -e "\t-k, --kind\t crd kind name. It must be singular, such as [Sample]"
     echo -e "\t-sn, --shortname\t crd kind short name. such as [s]"
     echo -e "\t-s, --scope\t crd scoped , support [${SCOPE_NAMESPACE} ${SCOPE_CLUSTER}]."
+    echo -e "\t-sp, --special\t support webhook to handle AdmissionRequest."
     exit 1
 }
 
@@ -69,6 +70,11 @@ while [ $# -gt 0 ];do
       SHORTNAME=$1
       shift
       ;;
+    --special|-sp)
+      shift
+      SPECIAL=true
+      shift
+      ;;
 
     --help|-h)
       shift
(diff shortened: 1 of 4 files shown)


hack/make-rules/add_controller.sh
```diff
@@ -36,6 +36,7 @@ function usage(){
     echo -e "\t-k, --kind\t crd kind name. It must be singular, such as [Sample]"
     echo -e "\t-sn, --shortname\t crd kind short name. such as [s]"
     echo -e "\t-s, --scope\t crd scoped , support [${SCOPE_NAMESPACE} ${SCOPE_CLUSTER}]."
+    echo -e "\t-sp, --special\t a new way to create webhook."
```
rambohe-ch, line 39: a new way to create webhook --> support webhook to handle AdmissionRequest

pkg/builder/defaulter_custom.go
```diff
@@ -0,0 +1,98 @@
+/*
```
rambohe-ch, line 1: how about move this file into `pkg/webhook/builder/defaulter_custom.go`

pkg/builder/validator_custom.go
```diff
@@ -0,0 +1,112 @@
+/*
```
rambohe-ch, line 1: ditto

pkg/builder/webhook.go
```diff
@@ -0,0 +1,208 @@
+/*
```
rambohe-ch, line 1: ditto

hack/make-rules/add_controller.sh
```diff
@@ -812,7 +818,182 @@ func (o *${KIND_FIRST_UPPER}ControllerOptions) Validate() []error {
...
 build_options
 build_apis_frame
 build_controller_frame
 build_webhook_frame
+
+if [ $SPECIAL ]; then
+  rm -rf ${ADD_WEBHOOK_FILE}
+  rm -rf ${WEBHOOK_KIND_VERSION_DIR}
+  build_webhook_special_frame
+fi
```
rambohe-ch, line 999: how about optimize line 993~999 as following:
```
if [ $SPECIAL ]; then
  build_webhook_special_frame
else
  build_webhook_frame
fi
```

rambohe-ch (approved):

rambohe-ch (approved):

PR #1363, opened 2023-04-11
fix some misses in yurt-manager

rambohe-ch:
/kind enhancement

1. move controllers setup after manager creation
2. don't print some dummy logs in a interval
3. fix nodepool controller corner case
4. fix gateway webhook mutate/validate path
5. fix some typo errors

Fixes #

```release-note

```


Congrool (commented):
/lgtm

Issue #1361, opened 2023-04-10
[feature request] support yurt-manager work in specified namespace


y-ykcir:
**What would you like to be added**:
All controllers and webhooks in yurt-manager can be deployed in specified namespace.

**Why is this needed**:
improve the deployment flexibility of OpenYurt components.

**others**
/kind feature

The current yurt-manager needs to be adapted to the specified namespace by modifying the following:

- [x] Pool-coordinator cert controller
- [x] Gateway controller in raven
- [x] Yurt-manager webhooks
- [x] Other components not yet migrated...

rambohe-ch:
@y-ykcir would you be able to take this work?

rambohe-ch:
@y-ykcir yurt-manager component has been updated recently, would you be able to check all yurt-manager again whether can work in specified namespace or not?

y-ykcir:
> @y-ykcir yurt-manager component has been updated recently, would you be able to check all yurt-manager again whether can work in specified namespace or not?

Sure

y-ykcir:
@rambohe-ch It seems that the newly migrated controllers do not need to be modified to be compatible with the specified namespace, maybe we can close this issue?

rambohe-ch:
> @rambohe-ch It seems that the newly migrated controllers do not need to be modified to be compatible with the specified namespace, maybe we can close this issue?

@y-ykcir would you like to take a test for OpenYurt in specified namespace, and make sure OpenYurt can work in specified namespace?

y-ykcir:
> > @rambohe-ch It seems that the newly migrated controllers do not need to be modified to be compatible with the specified namespace, maybe we can close this issue?
> 
> @y-ykcir would you like to take a test for OpenYurt in specified namespace, and make sure OpenYurt can work in specified namespace?

@rambohe-ch Is it possible to test this thing with Kind on the m1 Mac? Sorry that I currently do not have an available Linux server, it may not be convenient for testing.

rambohe-ch:
> > > @rambohe-ch It seems that the newly migrated controllers do not need to be modified to be compatible with the specified namespace, maybe we can close this issue?
> > 
> > 
> > @y-ykcir would you like to take a test for OpenYurt in specified namespace, and make sure OpenYurt can work in specified namespace?
> 
> @rambohe-ch Is it possible to test this thing with Kind on the m1 Mac? Sorry that I currently do not have an available Linux server, it may not be convenient for testing.

[...]

PR #1367, opened 2023-04-13
feat: support yurt-manager work in specified namespace

y-ykcir:
#### What type of PR is this?
/kind feature

support yurt-manager deploy in specified namespace:
support Gateway controller and Yurt-manager webhooks

Fixes parts of #1361 

/assign @rambohe-ch 

```release-note

```

### DIFF
--- a/charts/openyurt/templates/yurt-manager.yaml
+++ b/charts/openyurt/templates/yurt-manager.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: yurt-manager
-  namespace: kube-system
+  namespace: {{ .Release.Namespace | quote }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -15,14 +15,14 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: yurt-manager
-  namespace: kube-system
+  namespace: {{ .Release.Namespace | quote }}
 ---
 
 apiVersion: v1
 kind: Service
 metadata:
   name: yurt-manager-webhook-service
-  namespace: kube-system
+  namespace: {{ .Release.Namespace | quote }}
 spec:
   ports:
     - port: 443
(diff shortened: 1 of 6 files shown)


y-ykcir:
It looks like the test use `yurt-manager-auto-generated.yaml` directly to install components [here](https://github.com/openyurtio/openyurt/blob/4f62983921dd52c69d3bbe5616d532f22baa4b50/test/e2e/cmd/init/converter.go#L251). If I change `webhooks.clientConfig.service.namespace` to be rendered through helm intead of hardcode, it cannot be used directly.

charts/openyurt/templates/yurt-manager-auto-generated.yaml
```diff
@@ -729,7 +729,7 @@ webhooks:
   clientConfig:
     service:
       name: webhook-service
-      namespace: kube-system
+      namespace: {{ .Release.Namespace | quote }}
```
rambohe-ch, line 732: `charts/openyurt/templates/yurt-manager-auto-generated.yaml` is generated automatically, it is not need to modify the contents in this file.

hack/make-rules/kustomize_to_chart.sh
```diff
@@ -261,6 +261,8 @@ EOF
     append_note $yurt_manager_templatefile
 
     kubectl kustomize ${output_default_dir} >> $yurt_manager_templatefile
+
+    sed "s/namespace: kube-system/namespace: {{ .Release.Namespace | quote }}/g" $yurt_manager_templatefile > tmpfile && mv tmpfile $yurt_manager_templatefile
```
rambohe-ch, line 265: don't need to modify `yurt-manager-auto-generated.yaml` file. because namespace in `yurt-manager-auto-generated.yaml` file will be overwrite in yurt-manager with running namespace, the detail code is here: 
https://github.com/openyurtio/openyurt/blob/master/pkg/webhook/util/configuration/configuration.go#L68
https://github.com/openyurtio/openyurt/blob/master/pkg/webhook/util/configuration/configuration.go#L93

y-ykcir: Thank you for your explanation. It has been corrected.

rambohe-ch (approved):

PR #1384, opened 2023-04-18
inject clusterrolebinding for apiserver in pool-coordinator to enable `kubectl logs` within nodepool scope

Congrool:
#### What type of PR is this?
/kind enhancement

Currently, `kubectl logs` to poolcoordinator cannot work because the apiserver in pool-coordinator is not authorized to access the kubelet server, in other words it cannot get sub-resources `proxy/logs`. This pr will inject relative rbac rule to enable the apiserver to access the kubelet server.

It works in the following step:

1. Using a `initContainer`, which is `alpine:3.14`,  to get necessary tools that are used to inject rbac. Tools include  `busybox.static`, which is a statically linked busybox that can run in kube-apiserver container without any dynamic link lib, and `kubectl` which is actually used to send request to apiserver to create rbac rules.
2. A new configmap is added to provide rbac we need. It will be mounted by kube-apiserver container.
3. kube-apiserver container will also mount the dir that contains tools we previously prepared. The command of this container is changed to run in a shell, which will attempt to use kubectl to create rbac and execute the kube-apiserver command.

#### why not use postStart hook of kube-apiserver container:

After adding postStart hook, the kube-apiserver is too slow to be ready(typically over 20s) and will exit when exceeding the internal [dial timeout](https://github.com/kubernetes/kubernetes/blob/release-1.22/staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go#L63-L66)(20s used to dial to etcd and it's unconfigurable). It will cause the pool-coordinator cannot be ready and, of course, the postStart hook cannot get successfully executed.

I've not figured out why adding postStart hook has influence on kube-apiserver. 

#### other Note

TODOs:
- [x] rbac rules can be injected into apiserver
- [ ] kubeclt logs can work within nodepool

Currently the test `kubeclt logs can work within nodepool` has not got done yet, but I think it can work since we tested it through manually creating rbac into pool-coordinator before releasing openyurt v1.12.

### DIFF
--- a/charts/openyurt/templates/pool-coordinator.yaml
+++ b/charts/openyurt/templates/pool-coordinator.yaml
@@ -63,31 +63,84 @@ spec:
             labels:
               k8s-app: pool-coordinator
           spec:
+        initContainers:
+          - image: alpine:3.14
+            name: init-tools
+            command:
+              - /bin/sh
+              - -c
+              - |
+                sed -i "s/dl-cdn.alpinelinux.org/mirrors.aliyun.com/g" /etc/apk/repositories              
+                apk add busybox-static curl
+                cp $(which busybox.static) /tmp/tools/busybox
+                ARCH=$(uname -m)
+                VERSION={{ .Values.poolCoordinator.apiserverImage.tag }}
+                case $ARCH in
+                  x86_64)
+                    URL="https://dl.k8s.io/release/$VERSION/bin/linux/amd64/kubectl"
+                    ;;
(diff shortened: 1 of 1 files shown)


Congrool:
/hold

Need to solve the situation that node is disconnected from network. The pool-coordinator may fail to restart.

Congrool:
New PR to solve this problem #1637 , close this one now.

PR #1386, opened 2023-04-18
feat: add specify namespace install

fengshunli:

Fixes #

```release-note

```


charts/openyurt/values.yaml
```diff
@@ -1,6 +1,9 @@
 imagePullSecrets: []
 imagePullPolicy: IfNotPresent
 
+##  if unspecified, will use built-in variable `.Release.Namespace`.
+namespace: openyurt-system
```
rambohe-ch, line 5: how about keep consistency with the old version, so rename `openyurt-system` to `kube-system`?

rambohe-ch (approved):

PR #1388, opened 2023-04-18
support to specify enabled controllers

### LINKED ISSUE #1379: [feature request]support enable or disable controllers for yurt-manager
Like kube-controller-manager component, we can add a parameter `--controllers=-foo,bar,*` for yurt-manager component in order to enable or disable some specified controllers.

at the same time, webhook corresponding to controller should be disabled when controller is disabled.

**Why is this needed**:
End user can disable or enable controllers in yurt-manager according to their requirements.

**others**
/kind feature

xavier-hou:

Fixes #1379 

```release-note

```

### DIFF
--- a/charts/openyurt/templates/yurt-manager.yaml
+++ b/charts/openyurt/templates/yurt-manager.yaml
@@ -58,8 +58,8 @@ spec:
             - --logtostderr=true
             - --v={{ .Values.yurtManager.log.level }}
             - --working-namespace={{ include "openyurt.namespace" . }}
-            {{- if .Values.yurtManager.enabledControllers }}
-            - --controllers={{ .Values.yurtManager.enabledControllers }}
+            {{- if .Values.yurtManager.controllers }}
+            - --controllers={{ .Values.yurtManager.controllers }}
             {{- end }}
           command:
             - /usr/local/bin/yurt-manager
--- a/charts/openyurt/templates/yurt-manager.yaml
+++ b/charts/openyurt/templates/yurt-manager.yaml
@@ -58,6 +58,9 @@ spec:
             - --logtostderr=true
             - --v={{ .Values.yurtManager.log.level }}
             - --working-namespace={{ include "openyurt.namespace" . }}
+            {{- if .Values.yurtManager.enabledControllers }}
+            - --controllers={{ .Values.yurtManager.enabledControllers }}
+            {{- end }}
           command:
             - /usr/local/bin/yurt-manager
           image: {{ .Values.yurtManager.image.repository }}:{{ .Values.yurtManager.image.tag }}
--- a/charts/openyurt/values.yaml
+++ b/charts/openyurt/values.yaml
@@ -51,8 +51,8 @@ yurtManager:
     port: 10270
   metrics:
     port: 10271
-  # format should be "foo,-bar"
-  enabledControllers: ""
+  # format should be "foo,-bar,*"
+  controllers: ""
   healthProbe:
     port: 10272
   # resources of yurt-manager container
--- a/charts/openyurt/values.yaml
+++ b/charts/openyurt/values.yaml
@@ -51,6 +51,8 @@ yurtManager:
     port: 10270
   metrics:
     port: 10271
+  # format should be "foo,-bar"
+  enabledControllers: ""
   healthProbe:
     port: 10272
   # resources of yurt-manager container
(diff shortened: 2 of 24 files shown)


charts/openyurt/values.yaml
```diff
@@ -51,6 +51,8 @@ yurtManager:
     port: 10270
   metrics:
     port: 10271
+  # format should be "foo,-bar"
+  enabledControllers: ""
```
rambohe-ch, line 55: enabledControllers --> controllers

charts/openyurt/values.yaml
```diff
@@ -51,6 +51,8 @@ yurtManager:
     port: 10270
   metrics:
     port: 10271
+  # format should be "foo,-bar"
```
rambohe-ch, line 54: "foo,-bar" --> "foo,-bar, *"

pkg/webhook/server.go
```diff
@@ -40,33 +41,47 @@ type SetupWebhookWithManager interface {
...
+var controllerWebhook map[string][]SetupWebhookWithManager
 var WebhookHandlerPath = make(map[string]struct{})
 
+func addWebhook(name string, handler SetupWebhookWithManager) {
+	if controllerWebhook == nil {
+		controllerWebhook = make(map[string][]SetupWebhookWithManager)
+	}
+
+	if controllerWebhook[name] == nil {
+		controllerWebhook[name] = make([]SetupWebhookWithManager, 1)
```
rambohe-ch, line 53: make([]SetupWebhookWithManager, 1) --> make([]SetupWebhookWithManager, 0)

rambohe-ch:
@xavier-hou some webhooks maybe have not connected controllers, so we need to add a parameter `--webhooks` for enabling or disabling specified webhooks.

xavier-hou:
> @xavier-hou some webhooks maybe have not connected controllers, so we need to add a parameter `--webhooks` for enabling or disabling specified webhooks.

@rambohe-ch 
1. For contollers who have corresponding webhooks, such as `staticpod`, if the controller is disabled, then the corresponding webhook will be disabled as well. Because we use a map `controllerWebhook` to check whether controller has been disabled.
https://github.com/openyurtio/openyurt/pull/1388/files#diff-c07be401391cf159ca42b3e3c6e8f31d3d0a0a6b1b48998069b64d585a264f35R64 
2. For controllers who don't have corresponding webhooks, then we do nothing.

So I think current PR is enough for enabling and disabling controllers. But If we just want to enable or disable webhooks, then we can add a new parameter `--webhooks`.

pkg/webhook/server.go
xavier-hou: fixed

charts/openyurt/values.yaml
xavier-hou: done

charts/openyurt/values.yaml
xavier-hou: done

xavier-hou:
@rambohe-ch  We have lots of files aimed at adding controllers to `yurt-manager`, such as `add_csrapprover.go`. They are generated by `newcontroller` command which is convenient for controllers migration but annoying to maintain in the future. How about removing these files and accomplish all controllers register in `controller.go`?

rambohe-ch:
@xavier-hou and i will merge this pr at first.

charts/openyurt/templates/yurt-manager.yaml
```diff
@@ -58,6 +58,9 @@ spec:
             - --logtostderr=true
             - --v={{ .Values.yurtManager.log.level }}
             - --working-namespace={{ include "openyurt.namespace" . }}
+            {{- if .Values.yurtManager.enabledControllers }}
```
rambohe-ch, line 61: .Values.yurtManager.enabledControllers --> .Values.yurtManager.controllers

charts/openyurt/templates/yurt-manager.yaml
```diff
@@ -58,6 +58,9 @@ spec:
...
+            - --controllers={{ .Values.yurtManager.enabledControllers }}
```
rambohe-ch, line 62: ditto

charts/openyurt/templates/yurt-manager.yaml
xavier-hou: fixed

charts/openyurt/templates/yurt-manager.yaml
xavier-hou: fixed

rambohe-ch (approved):

PR #1404, opened 2023-04-21
fix: set default podspec for staticpod podtemplatespec

y-ykcir:
#### What type of PR is this?
/kind bug

If not set default `PodSpec` for `staticpod` `PodTemplateSpec`, then when applying `staticpod`, it may be rejected due to `PodTemplateSpec` validation not passing.

Fixes #

```release-note

```

### DIFF
--- a/pkg/apis/apps/v1alpha1/default.go
+++ b/pkg/apis/apps/v1alpha1/default.go
@@ -222,6 +222,11 @@ func SetDefaultsStaticPod(obj *StaticPod) {
 		obj.Spec.RevisionHistoryLimit = new(int32)
 		*obj.Spec.RevisionHistoryLimit = 10
 	}
+
+	podSpec := &obj.Spec.Template.Spec
+	if podSpec != nil {
+		SetDefaultPodSpec(podSpec)
+	}
 }
 
 // SetDefaultsYurtAppDaemon set default values for YurtAppDaemon.


pkg/apis/apps/v1alpha1/default.go
```diff
@@ -222,6 +222,11 @@ func SetDefaultsStaticPod(obj *StaticPod) {
 		obj.Spec.RevisionHistoryLimit = new(int32)
 		*obj.Spec.RevisionHistoryLimit = 10
 	}
+
+	podSpec := &obj.Spec.Template.Spec
+	if podSpec != nil {
```
xavier-hou, line 227: Maybe `if podSpec == nil` ?

pkg/apis/apps/v1alpha1/default.go
```diff
@@ -222,6 +222,11 @@ func SetDefaultsStaticPod(obj *StaticPod) {
 		obj.Spec.RevisionHistoryLimit = new(int32)
 		*obj.Spec.RevisionHistoryLimit = 10
 	}
+
+	podSpec := &obj.Spec.Template.Spec
```
xavier-hou, line 226: Should we check the existence of `obj.Spec.Template`?

xavier-hou:
@y-ykcir Would you explain more about this case? From my point, I think we do need to set the `PodSpec` in order to pass the validation. When will we need staticpod just have `metav1.ObjectMeta`~?

y-ykcir:
@xavier-hou When I apply yurthub-staticpod.yaml. The following validation error occurred:
```
The StaticPod "yurt-hub" is invalid:
* template.spec.containers[0].livenessProbe.httpGet.scheme: Unsupported value: "": supported values: "HTTP", "HTTPS"
* template.spec.containers[0].livenessProbe.successThreshold: Invalid value: 0: must be 1
* template.spec.containers[0].terminationMessagePolicy: Required value: must be 'File' or 'FallbackToLogsOnError'
* template.spec.containers[0].env[0].valueFrom.fieldRef.apiVersion: Required value
* template.spec.restartPolicy: Required value
* template.spec.dnsPolicy: Required value
```
This should be because the validation requirement here is a complete template, and users may not always add default values when writing yaml. Therefore, it may be appropriate to add default values to fields that the user did not fill in during the mutate phase

xavier-hou:
@y-ykcir Really appreciate for your information~

rambohe-ch (approved):

PR #1417, opened 2023-04-25
improve controller names

rambohe-ch:
/kind enhancement

improve controller names as following:
- gateway and service controllers are sub controller for `ravenl3`
- servicetopology-endpoints and servicetopology-endpointslice controllers are sub controller for `servicetopology`

Fixes #

```release-note

```


pkg/controller/raven/gateway/gateway_controller.go
```diff
@@ -51,13 +51,9 @@ var (
...
 )
 
-const (
-	controllerName = "Gateway-controller"
-)
-
 func Format(format string, args ...interface{}) string {
 	s := fmt.Sprintf(format, args...)
-	return fmt.Sprintf("%s: %s", controllerName, s)
+	return fmt.Sprintf("%s: %s", common.ControllerName, s)
```
YTGhost, line 56: Perhaps we could output it like this? `fmt.Sprintf("%s: %s", fmt.Sprintf("%s-gateway", common.ControllerName), s)`
This should help us more effectively to know exactly which sub-controller is outputting the logs.

rambohe-ch: fixed

PR #1422, opened 2023-04-28
optimize staticpod controller for upgrade info

JameKeal:
/kind enhancement

There are some double counting in the existing code on upgrde info, and the calculation part is scattered and the parameter assignment is not uniform.
image
image

So this pr is used to optimized the upgrade info, and keep the computation in one place, rather than spreading it out.

Meanwhile, peel off the relevant calculation from staticpod_controller to upgrade_info, make the responsibilities clearer.

Fixes #

/assign @rambohe-ch 

```release-note

```

### DIFF
--- a/pkg/controller/staticpod/upgradeinfo/upgrade_info.go
+++ b/pkg/controller/staticpod/upgradeinfo/upgrade_info.go
@@ -18,6 +18,7 @@ package upgradeinfo
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
@@ -41,6 +42,9 @@ type UpgradeInfo struct {
 	// Upgrade worker pod running on the node
 	WorkerPod *corev1.Pod
 
+	// Indicate whether the static pod is ready
+	StaticPodReady bool
+
 	// Indicate whether the static pod on the node needs to be upgraded.
 	// If true, the static pod is not up-to-date and needs to be upgraded.
 	UpgradeNeeded bool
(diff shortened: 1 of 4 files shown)


rambohe-ch:
@JameKeal please fix the unit tests error

rambohe-ch:
@xavier-hou PTAL

pkg/controller/staticpod/upgradeinfo/upgrade_info.go
```diff
@@ -128,64 +179,52 @@ func ReadyUpgradeWaitingNodes(infos map[string]*UpgradeInfo) []string {
...
-	for _, info := range infos {
-		if info.StaticPod != nil {
-			if podutils.IsPodReady(info.StaticPod) {
-				readyNumber++
-			}
+		// sync worker pods info
+		if info.WorkerPodDeleteNeeded {
+			deletePods = append(deletePods, info.WorkerPod)
+		}
+		if info.WorkerPodStatusPhase != corev1.PodFailed && info.WorkerPodStatusPhase != corev1.PodSucceeded {
```
xavier-hou, line 225: Do we need to consider the existence of workerpod?
In this situation,`info.WorkerPodStatusPhase` may be "", and the result of `allSucceeded` is false

pkg/controller/staticpod/upgradeinfo/upgrade_info.go
```diff
@@ -128,64 +179,52 @@ func ReadyUpgradeWaitingNodes(infos map[string]*UpgradeInfo) []string {
 func ReadyNodes(infos map[string]*UpgradeInfo) []string {
```
xavier-hou, line 179: It seems that this function is not used. How about removing it?

pkg/controller/staticpod/upgradeinfo/upgrade_info.go
JameKeal: yes, you are right.

xavier-hou:
@rambohe-ch LGTM  @JameKeal Plz fix the UT

rambohe-ch (approved):

PR #1430, opened 2023-05-05
Update values.yaml

succw:

Fixes #

```release-note

```

### DIFF
--- a/charts/openyurt/values.yaml
+++ b/charts/openyurt/values.yaml
@@ -43,7 +43,7 @@ yurtManager:
     # log level for yurt-manager
     level: 4
   replicas: 1
-  image:
+  Image:
     registry: openyurt
     repository: yurt-manager
     tag: latest


charts/openyurt/values.yaml
```diff
@@ -43,7 +43,7 @@ yurtManager:
     # log level for yurt-manager
     level: 4
   replicas: 1
-  image:
```
Congrool, line 46: lgtm, but could you keep the same format which starts from lower case as other fields? I think it's better to replace `Image` as `image` in `charts/openyurt/templates/yurt-manager.yaml`

rambohe-ch: @succw Very appreciate you can make a pull request, but would you be able to add some message about the reason why we need to modify `image` to `Image`?

Congrool: @rambohe-ch I guess it's because what we use in [yurt-manager.yaml](https://github.com/openyurtio/openyurt/blob/77e2f61ef86157b03f85e6146d701183a4699ed7/charts/openyurt/templates/yurt-manager.yaml#LL78C53-L78C53) template is "Image" instead of "image".

rambohe-ch: @Congrool Thank you for the reminder. @succw yes, i agree with @Congrool , it's a good idea to modify `yurt-manager.yaml` instead of `values.yaml`.

rambohe-ch:
@succw helm charts have been updated for this issue: https://github.com/openyurtio/openyurt/issues/1433
so i will close this pull request. Thank you for your contribution again.

PR #1444, opened 2023-05-10
feat: prevent node movement by label modification

### LINKED ISSUE #1382: [feature request]can not move node to another nodepool by label modification
the nodepool label of node can not be updated after node creation, this means we can not move a node into another nodepool by changing node label simply. 

if end users want to move a node to another nodepool, they should drain all pods on the node at first, and then delete and reset this node from cluster, and re-join this node into the cluster.

so we will add a webhook in yurt-manager to prevent updating nodepool label of node.

**Why is this needed**:
Because the NodePool is the first class citizen in OpenYurt, and node must belong to a NodePool and only one NodePool, it is not secure to move a node into another NodePool by label changing.

**others**
/kind feature

y-ykcir:
#### What type of PR is this?

/kind feature

Fixes #1382 

add a validation webhook for node:
user can not change `apps.openyurt.io/desired-nodepool` if node has been added in noodpool
user can not change `apps.openyurt.io/nodepool` directly, only yurt-manger can change this label

```release-note

```

### DIFF
--- a/charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
+++ b/charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
@@ -590,6 +590,26 @@ webhooks:
   sideEffects: None
 - admissionReviewVersions:
   - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: yurt-manager-webhook-service
+      namespace: {{ .Release.Namespace }}
+      path: /validate-core-openyurt-io-v1-node
+  failurePolicy: Fail
+  name: validate.core.v1.node.openyurt.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - UPDATE
+    resources:
+    - nodes
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
   clientConfig:
     service:
       name: yurt-manager-webhook-service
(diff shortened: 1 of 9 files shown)


rambohe-ch:
@y-ykcir please fix github action's errors.

y-ykcir:
@rambohe-ch It seems that in e2e testing, the nodepool, yurtappdaemon, and yurtappset tests will remove nodes from the nodepool by deleting the NodePoolLabel at the beginning and end of each test. Should we modify such behavior in e2e to use join command and is there any obstacle to do this?

rambohe-ch:
> @rambohe-ch It seems that in e2e testing, the nodepool, yurtappdaemon, and yurtappset tests will remove nodes from the nodepool by deleting the NodePoolLabel at the beginning and end of each test. Should we modify such behavior in e2e to use join command and is there any obstacle to do this?

@y-ykcir would you like to have to try to improve e2e tests? but i think that `yurtadm join` command can not be used in kind cluster.

rambohe-ch:
@y-ykcir I think we only need to make sure `apps.openyurt.io/desired-nodepool` label of node can not be changed, no matter which component changes this label.

at the same time, we don't need to verify `apps.openyurt.io/nodepool` label, because NodePool controller will make sure `apps.openyurt.io/nodepool` label is equal to `apps.openyurt.io/desired-nodepool` label.

y-ykcir:
@rambohe-ch The code logic has been simplified. Is it possible to add reset and join way in the Kind cluster? If it's not feasible, I am currently thinking of the following handling methods:
1. Modify to run all tests in an unchanged nodepool distribution (nodepool tests may have to change nodepool)
2. Temporarily disable this webhook in e2e testing

rambohe-ch (approved):

PR #1453, opened 2023-05-12
rename auto to AdvancedRollingUpdate

### LINKED ISSUE #1450: [Question]How about rename auto upgrade strategy to AdvancedRollingUpdate
**What happened**:
 The newly added upgrade models: auto and ota are convenient way for user upgrade DaemonSet and StaticPod on edge nodes, but the name convention: auto is not easy to understand for users.
 
    - auto: rolling update pods of DaemonSet or StaticPod, but skip pods on NotReady nodes. And when NotReady nodes turn to Ready,  pods on these nodes will be upgraded automatically.

 so i think that `auto` is also a rolling update strategy, but with some advanced features. so how about rename `auto` to `AdvancedRollingUpdate`? i think that new name(AdvancedRollingUpdate) is more understandable for users.

by the way, `ota` upgrade strategy should be upper-cased, so how about rename `ota` to `OTA`?

rename newly added upgrade strategy as following:
- auto --> AdvancedRollingUpdate
- ota --> OTA

NA

NA

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**
/kind question

xavier-hou:

Fixes #1450 

```release-note

```

### DIFF
--- a/docs/proposals/20220718-workload-update-model.md
+++ b/docs/proposals/20220718-workload-update-model.md
@@ -78,7 +78,7 @@ Note: No modification to Kubernetes, keeping it non-invasive, no modification to
   <img src="../img/OTA.png" width=50% title="OTA">
 </div>
 
-**Pre-requisites**: Set application's updateStrategy to `type: OnDelete`; add a new annotation(`apps.openyurt.io/upgrade-strategy: ota`) for DaemonSet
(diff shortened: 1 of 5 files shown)


xavier-hou:
@rambohe-ch  PTAL

rambohe-ch (approved):

PR #1458, opened 2023-05-13
fix yurthub static pod isntallation in yurtadm join command

### LINKED ISSUE #1456: [feature request]how about rename StaticPod to StaticSet
The newly added CRD `StaticPod` is used as a workload for managing static pods on nodes, and this name is very confused for users, because the pods for this workload are called static pod in kubernetes, but the workload is also called `StaticPod`,  so it is very hard to distinguish which `static pod` you are talking about.

I think it is a good idea for us that we can refer to DaemonSet to name `static pod`, so how about rename `StaticPod` to `StaticSet`?

by the way, the name for the configmap that used for storing template of static pod can also be improved as `static-set-{ins.Name}`.

**others**
/kind feature

rambohe-ch:
/kind enhancement

1. modify `yurthub-staticpod.yaml` to `yurthub-yurtstaticset.yaml`
2. improve yurthub static pod installation in yurtadm join command

Fixes #1456 

```release-note

```


Congrool (commented):
/lgtm

PR #1470, opened 2023-05-18
add pod ready updater controller

### LINKED ISSUE #1426: [feature request] Pods should keep Ready when the edge node pool disconnect with kube-apiserver
When more than a certain percentage of nodes in an edge node pool cannot access kube-apiserver, those nodes become not ready and also their Pods become not ready.
Because nodes' keepalive time(renewTime in node lease) are different, they cannot become not ready at the same time.
Currently, the Pods' DELETE operation can be intercepted by ValidatingWebhook because of enough tolerationSeconds. But the Pods' status are updated to be not ready immediately after nodes become not ready. And then the related endpoints are dropped from service. This leads to that the online nodes in this node pool won't be able to access the Pods in offline nodes. Maybe at this time, the whole node pool disconnect with kube-apiserver and kube-proxy in this offline node pool cannot take any actions. But when the network recovers, the Pods will become not ready firstly and then become ready again. All services will stop working for seconds or longer.
So Pods' status is expected to be Ready when more than a certain percentage of nodes in an edge node pool cannot access kube-apiserver.

**Why is this needed**:
The edge side services are expected to continue working when the edge node pool cannot access kube-apiserver.

**others**
/kind feature

JameKeal:
/kind enhancement

When node turn to NotReady, the kubernetes will make the pod ready condition from True to False which running on the node, and the endpoints controller in kube-controller-manager will clean up the pod who is not ready, and update the endpoints info, the kube-proxy or other components will not use service to find the pod. 

Fixes #1426 

This pr will resolve some condition here:
1. When nodepool is disconnected with cloud, the pod ready condition is turn to false, we need to change it to true.
2. When node turn to NotReady, if some pod has pod_binding annotation, we need to change ready condition to true.


JameKeal:
I will add some unit test after a while

pkg/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,311 @@
...
+		klog.Errorf("Fail to get Pod %v, %v", request.NamespacedName, err)
+		return reconcile.Result{}, err
+	}
+
+	if instance.DeletionTimestamp != nil {
+		return reconcile.Result{}, nil
+	}
+
+	// if pod has binding annotation, deal it.
+	if instance.Annotations != nil && instance.Annotations[constant.PodBindingAnnotation] == "true" {
```
rambohe-ch, line 255: Pod Binding Annotation is removed from pod, and only can be added on the node.

pkg/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,311 @@
...
+		Reconciler: r, MaxConcurrentReconciles: concurrentReconciles,
+	})
+	if err != nil {
+		return err
+	}
+
+	// 1. Watch for changes to Pod
+	podReadyPredicate := predicate.Funcs{
+		CreateFunc: func(evt event.CreateEvent) bool {
+			return false
```
rambohe-ch, line 107: we should also handle `Pod Create` event, because of `add events` will be received when yurt-manager restart.

pkg/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,311 @@
...
+
+	if err := c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForObject{}, podReadyPredicate); err != nil {
+		return err
+	}
+
+	// 2. Watch for changes to NodePool
+	// only watch pods changes may leave out some pods, because nodepool status may not update as quick as pod
+	nodePoolPredicate := predicate.Funcs{
+		CreateFunc: func(evt event.CreateEvent) bool {
+			return false
```
rambohe-ch, line 128: ditto

pkg/webhook/pod/v1/pod_default.go
```diff
@@ -0,0 +1,70 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	nodeutil "github.com/openyurtio/openyurt/pkg/controller/util/node"
+	podutil "github.com/openyurtio/openyurt/pkg/controller/util/pod"
+	"github.com/openyurtio/openyurt/pkg/controller/yurtcoordinator/constant"
+)
+
+// Default satisfies the defaulting webhook interface.
+func (webhook *PodHandler) Default(ctx context.Context, obj runtime.Object, req admission.Request) error {
+	po, ok := obj.(*v1.Pod)
```
rambohe-ch, line 37: I think we should only handle pod status update requests initiated by kube-controller-manager

pkg/webhook/pod/v1/pod_default.go
```diff
@@ -0,0 +1,70 @@
...
+	if !ok {
+		return apierrors.NewBadRequest(fmt.Sprintf("expected a Pod but got a %T", obj))
+	}
+
+	// if pod have ready condition and condition is False
+	podReadyCondition := podutil.GetPodReadyCondition(po.Status)
```
rambohe-ch, line 43: we can use `nodeutil.GetPodCondition()` to get the pointer for ready condition, and then verify and modify it.

btw: line64~66 is not needed, we just need to modify ready condition.

pkg/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,325 @@
...
+// ReconcilePodReadyUpdater reconciles a PodReadyUpdater object
+type ReconcilePodReadyUpdater struct {
+	client.Client
+	scheme *runtime.Scheme
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(c *appconfig.CompletedConfig, mgr manager.Manager) reconcile.Reconciler {
+	return &ReconcilePodReadyUpdater{
+		Client: utilclient.NewClientFromManager(mgr, ControllerName),
```
rambohe-ch, line 87: utilclient.NewClientFromManager(mgr, ControllerName) --> mgr.GetClient()

pkg/webhook/pod/v1/pod_default.go
JameKeal: yes, i set the MutatingWebhookConfiguration for resource is pods/status, and the operations is update, so only the pods/status update will handle

pkg/yurtmanager/webhook/pod/v1/pod_default.go
```diff
@@ -0,0 +1,66 @@
...
+	// check the pod in which node and if node have pod_binding annotation
+	nodeName := po.Spec.NodeName
+	if nodeName == "" {
+		return nil
+	}
+	node := v1.Node{}
+	if err := webhook.Client.Get(context.TODO(), client.ObjectKey{Name: nodeName}, &node); err != nil {
+		return nil
+	}
+	if node.Annotations != nil && node.Annotations[constant.PodBindingAnnotation] == "true" {
```
rambohe-ch, line 59: we should also support the autonomy annotation in order to keep consistency.

pkg/yurtmanager/webhook/pod/v1/pod_default.go
```diff
@@ -0,0 +1,66 @@
...
+		// update pod ready condition to True
+		podReadyCondition.Status = v1.ConditionTrue
```
rambohe-ch, line 61: we should only mutate pod ready condition for pod status update requests from kcm component.

charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
```diff
@@ -126,6 +126,8 @@ rules:
   resources:
   - nodepools
   verbs:
+  - create
```
rambohe-ch, line 129: I think that nodepool create/delete rbac rights are not needed.

pkg/yurtmanager/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,319 @@
...
+		for _, nodeName := range nodePool.Status.Nodes {
+			podList := &corev1.PodList{}
+			selector := fields.OneTermEqualSelector("spec.nodeName", nodeName)
+			err := c.List(context.TODO(), podList, &client.ListOptions{FieldSelector: selector})
+			if err != nil {
+				continue
+			}
+			for i := range podList.Items {
+				// if pod Ready Condition is already True, skip.
+				if podutil.IsPodReadyConditionTrue(podList.Items[i].Status) {
```
rambohe-ch, line 150: we also need to skip pod that has no pod ready condition.

pkg/yurtmanager/controller/podreadyupdater/podreadyupdater_controller.go
```diff
@@ -0,0 +1,319 @@
...
+	if len(instance.Spec.NodeName) == 0 {
+		return reconcile.Result{}, nil
+	}
+
+	node := &corev1.Node{}
+	err = r.Get(context.TODO(), types.NamespacedName{Name: instance.Spec.NodeName}, node)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
rambohe-ch, line 275: If node has autonomy or pod-binding annotation, we can mark the pod ready status to ready.

charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
JameKeal: yes, of course, but the CI / verify was failed if i don't add it, i will try it again.

pkg/yurtmanager/controller/podreadyupdater/podreadyupdater_controller.go
JameKeal: This controller is only handler nodepool disconnect with cloud, the node operator is handler by webhook, not here.

Issue #1440, opened 2023-05-09
[BUG]There is a memory leak in the yurt-tunnel-server, consuming too much memory


huweihuang:
**What happened**:

There is a memory leak in the yurt-tunnel-server, consuming too much memory, Currently the memory over 80G.

```
 crictl ps
CONTAINER           IMAGE               CREATED             STATE               NAME                        ATTEMPT             POD ID              POD
554efaae142a4       acb209d36497c       4 months ago        Running             yurt-tunnel-server          0                   b4f2de2cd768d       yurt-tunnel-server-558cd556d8-mbbvf


crictl stats
CONTAINER           CPU %               MEM                 DISK                INODES
554efaae142a4       2.28                82.48GB             53.25kB             15
```

**Environment**:

version: yurt-tunnel-server   :       v1.1.0       

```
docker.io/openyurt/yurt-tunnel-server                             v1.1.0              acb209d36497c       15.3MB
```

**others**

/kind bug

River-sh:
Please provide the cluster environment, some special configurations, and pprof information

huweihuang:
Are there any input errors in this command? Error returning `unrecognized profile format`

``` 
go tool pprof http://HOST_IP/debug/pprof/profile
Fetching profile over HTTP from http://HOST_IP/debug/pprof/profile
http://HOST_IP/debug/pprof/profile: parsing profile: unrecognized profile format
```

rambohe-ch:
> Are there any input errors in this command? Error returning `unrecognized profile format`
> 
> ```
> go tool pprof http://HOST_IP/debug/pprof/profile
> Fetching profile over HTTP from http://HOST_IP/debug/pprof/profile
> http://HOST_IP/debug/pprof/profile: parsing profile: unrecognized profile format
> ```

@huweihuang The default meta port for tunnel-server is 10265

huweihuang:
```
(pprof) top 20
Showing nodes accounting for 47.47GB, 97.58% of 48.65GB total
Dropped 319 nodes (cum <= 0.24GB)
Showing top 20 nodes out of 31
      flat  flat%   sum%        cum   cum%
   25.91GB 53.27% 53.27%    25.91GB 53.27%  bufio.NewWriterSize (inline)
   12.91GB 26.54% 79.80%    12.91GB 26.54%  bufio.NewReaderSize (inline)
[...]

PR #1471, opened 2023-05-20
fix memory leak for yur-tunnel-server

huweihuang:
#### What type of PR is this?

> /kind bug

There is a memory leak in the yurt-tunnel-server, consuming too much memory.

The reason is that there is a goroutine memory leak in the dependency `apiserver-network-proxy`

Fixes # https://github.com/openyurtio/openyurt/issues/1440

@rambohe-ch


rambohe-ch:
@huweihuang please fix verify ci errors. maybe you need execute command `go mod tidy` before git push.

rambohe-ch (approved):

rambohe-ch:
Fixes #1440

PR #1481, opened 2023-05-22
Update init.go

guoguodan:
add condition: len(s)==2

### DIFF
--- a/test/e2e/cmd/init/init.go
+++ b/test/e2e/cmd/init/init.go
@@ -526,7 +526,7 @@ func (ki *Initializer) loadImagesToKindNodes(images, nodes []string) error {
 func validateKubernetesVersion(ver string) error {
 	s := strings.Split(ver, ".")
 	var originVer = ver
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@guoguodan very appreciate for your pull request. would you like to explain the reason why should improve the condition: `len(s)==2`?

rambohe-ch:
@guoguodan by the way, the e2e test has failed because of version check.

guoguodan:
@rambohe-ch Thinks for your reply. Overall context, the conditios is exists:len(s)<2 , len(s) >3 and len(s)==3 ,but leak the condition: len(s)==2

rambohe-ch:
> @rambohe-ch Thinks for your reply. Overall context, the conditios is exists:len(s)<2 , len(s) >3 and len(s)==3 ,but leak the condition: len(s)==2

@guoguodan I think that len(s) == 2 should be true in the condition, like version: v1.22 is allowed to run a kind cluster.

PR #1486, opened 2023-05-23
improve yurtstaticset template metadata

### LINKED ISSUE #1473: [feature request]improve yurtstaticset template metadata
Take https://github.com/openyurtio/openyurt/blob/master/charts/yurthub/templates/yurthub-yurtstaticset.yaml#L14-L15 as a example, it is not need to define name and namespace fields, and mutate webhook of yurtstaticset can set these fields in template metadata by yurtstaticset.name and yurtstaticset.namespace(https://github.com/openyurtio/openyurt/blob/master/charts/yurthub/templates/yurthub-yurtstaticset.yaml#L4-L5)

and we should use yurtstaticset name and namespace to replace name and namespace in template metadata even these fields are defined.

**Why is this needed**:
If user defines different namespace in pod template with yurtstaticset namespace, yurtstaticset controller don't know how to handle this case.

**others**
/kind feature

JameKeal:
/kind feature

If user defines different namespace in pod template with yurtstaticset namespace, yurtstaticset controller don't know how to handle this case.

Fixes #1473 

/assign @rambohe-ch 

```release-note

```

### DIFF
--- a/pkg/apis/apps/v1alpha1/default.go
+++ b/pkg/apis/apps/v1alpha1/default.go
@@ -227,6 +227,10 @@ func SetDefaultsYurtStaticSet(obj *YurtStaticSet) {
 	if podSpec != nil {
 		SetDefaultPodSpec(podSpec)
 	}
+
+	// use YurtStaticSet name and namespace to replace name and namespace in template metadata
+	obj.Spec.Template.Name = obj.Name
+	obj.Spec.Template.Namespace = obj.Namespace
 }
 
 // SetDefaultsYurtAppDaemon set default values for YurtAppDaemon.
--- a/charts/yurthub/templates/yurthub-cloud-yurtstaticset.yaml
+++ b/charts/yurthub/templates/yurthub-cloud-yurtstaticset.yaml
@@ -11,8 +11,6 @@ spec:
     metadata:
       labels:
         k8s-app: yurt-hub-cloud
-      name: yurt-hub-cloud
-      namespace: {{ .Release.Namespace }}
     spec:
       volumes:
         - name: hub-dir
--- a/charts/yurthub/templates/yurthub-yurtstaticset.yaml
+++ b/charts/yurthub/templates/yurthub-yurtstaticset.yaml
@@ -11,8 +11,6 @@ spec:
     metadata:
       labels:
         k8s-app: yurt-hub
-      name: yurt-hub
-      namespace: {{ .Release.Namespace }}
     spec:
       volumes:
         - name: hub-dir


pkg/apis/apps/v1alpha1/default.go
```diff
@@ -227,6 +227,10 @@ func SetDefaultsYurtStaticSet(obj *YurtStaticSet) {
 	if podSpec != nil {
 		SetDefaultPodSpec(podSpec)
 	}
+
+	// use YurtStaticSet name and namespace to replace name and namespace in template metadata
+	obj.Spec.Template.Name = obj.Name
```
fengshunli, line 232: Obj.Spec adds empty finger judgment

pkg/apis/apps/v1alpha1/default.go
```diff
@@ -227,6 +229,14 @@ func SetDefaultsYurtStaticSet(obj *YurtStaticSet) {
...
 		SetDefaultPodSpec(podSpec)
 	}
+
+	// use YurtStaticSet name and namespace to replace name and namespace in template metadata
+	if !reflect.DeepEqual(obj.Spec, reflect.Zero(reflect.TypeOf(obj.Spec)).Interface()) {
+		if _, ok := reflect.TypeOf(obj.Spec).FieldByName("Template"); ok {
+			obj.Spec.Template.Name = obj.Name
+			obj.Spec.Template.Namespace = obj.Namespace
+		}
+	}
```
rambohe-ch, line 239: Line234~239 can be optimized as following:
```
obj.Spec.Template.Name = obj.Name
obj.Spec.Template.Namespace = obj.Namespace
```

JameKeal: That's what I used to write, but @fengshunli suggest Obj.Spec adds empty finger judgment, so whether to modify it?

fengshunli: > That's what I used to write, but @fengshunli suggest Obj.Spec adds empty finger judgment, so whether to modify it?

@rambohe-ch WDYT？

rambohe-ch: @fengshunli @JameKeal I think we don't need to add condition for this code snippet, because `obj.Spec.Template` field must be present.

rambohe-ch: @fengshunli Take K8s code as example, you can check the default setting for daemonset creation in the following link: 
https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/apps/daemonset/strategy.go#L79-L81

JameKeal: > @fengshunli @JameKeal I think we don't need to add condition for this code snippet, because `obj.Spec.Template` field must be present.

I think so.

rambohe-ch (approved):

PR #1492, opened 2023-05-24
add proposal of nodepool with host network mode

Lan-ce-lot:
Add proposal of support nodepool with host network mode

Fixes #

```release-note

```


docs/proposals/20230524-support-nodepool-with-host-network-mode.md
```diff
@@ -0,0 +1,157 @@
...
+EOF
+```
+
+Restart kubelet
+
+```
+systemctl daemon-reload & systemctl restart kubelet
+```
+
+* Step2: Design a new field for NodePool called hostNetwork=true/false. True indicates that this is a NodePool of hostNetwork mode type.
```
rambohe-ch, line 124: The default value of field `hostNetwork` should be false.

docs/proposals/20230524-support-nodepool-with-host-network-mode.md
```diff
@@ -0,0 +1,157 @@
...
+* Step3: Modify the NodePool controller and add `nodepool.openyurt.io/host-network`=`true` tag to all nodes in the node-network-mode NodePool.
```
rambohe-ch, line 125: we should add `nodepool.openyurt.io/host-network=true` label to node in the yurtadm join when node is joined into the cluster. because flannel/kube-proxy/coredns daemon pod will be run on the node before the NodePool controller add this label on the node.

rambohe-ch:
@Lan-ce-lot Very appreciate for your proposal about host network NodePool. and would you like to share your proposal on the community meeting?

docs/proposals/20230524-support-nodepool-with-host-network-mode.md
Lan-ce-lot: Yes, we should determine the class of the nodepool when yurtamd join node, and label node if the node pool is in host network mode. 
BTW, should we still support Label to join the nodepool?

docs/proposals/20230524-support-nodepool-with-host-network-mode.md
```diff
@@ -0,0 +1,158 @@
...
+/var/lib/kubelet/kubeadm-flags.env
+
+`—network-plugin=cni`
+
+`—cni-conf-dir=/etc/cni/net.d`
+
+For example:
+
+```
+KUBELET_KUBEADM_ARGS="--network-plugin=cni —cni-conf-dir=/etc/cni/net.d --pod-infra-container-image=registry.cn-hangzhou.aliyuncs.com/google_containers/pause:3.5"
```
huiwq1990, line 103: set kubelet --network-plugin="" will make node work in host network mode.

huiwq1990: --network-plugin string                                    The name of the network plugin to be invoked for various events in kubelet/pod lifecycle. This docker-specific flag only works when container-runtime is set to docker. (DEPRECATED: will be removed along with dockershim.)
network-plugin only work with docker, so it's not a good way.

Lan-ce-lot: Thank you for your advice.

Yes, when --network-plugin="", we will use docker network bridge.

image


image

network-plugin only work with docker, That's a problem, because that means it's not a universal approach.

huiwq1990:
@Lan-ce-lot @rambohe-ch If cni config use `/etc/cni/net.d/0-loopback.conf`,  Did the pod deploy on the node must set hostnetwork=true ?

rambohe-ch:
> @Lan-ce-lot @rambohe-ch If cni config use `/etc/cni/net.d/0-loopback.conf`, Did the pod deploy on the node must set hostnetwork=true ?

@huiwq1990 yes, if the node belongs to a hostnetwork mode nodepool, The hostNetwork field of pod that is assigned to this node will be modified to true by the filter in yurthub.

docs/proposals/20230524-support-nodepool-with-host-network-mode.md
```diff
@@ -0,0 +1,158 @@
...
+
+Restart kubelet
+
+```
+systemctl daemon-reload & systemctl restart kubelet
+```
+We should add nodepool.openyurt.io/host-network=true label to node in the yurtadm join when node is joined into the cluster. Because flannel/kube-proxy/coredns daemon pod will be run on the node before the NodePool controller add this label on the node.
+
+* Step2: Design a new field for NodePool called hostNetwork=true/false. True indicates that this is a NodePool of hostNetwork mode type.
+* Step3: Modify the NodePool controller to manage the hostNetwork label of all nodes in the node-network-mode NodePool.
```
rambohe-ch, line 126: Because `yurtadm join` should specify the nodepool for the node, so we can add `nodepool.openyurt.io/host-network=true` label as kubelet parameter instead of labeling by NodePool controller.

docs/proposals/20230524-support-nodepool-with-host-network-mode.md
```diff
@@ -0,0 +1,158 @@
...
+* Step4: Add a webhook to verify if there is a nodepool named xxx for the pods with the `nodepool.openyurt.io/desired-nodepool: xxx` tag. If it exists, check if the hostNetwork field of the nodepool is true. If it is true, set the pod's hostNetwork to true. Users can add the pod label themselves or use YurtAppSet and YurtAppDaemon to deploy the application.
```
rambohe-ch, line 127: we can add a hostNetwork filter in yurthub to deal with changing hostNetwork field for pods that run on the hostNetwork mode node. so we can handle all pods even if pods are not controlled by YurtAppSet and YurtAppDaemon.

rambohe-ch (approved):

PR #1501, opened 2023-05-29
fix yurthub memory leak

JameKeal:
/kind bug

When yurthub connect with pool-coordinator etcd failed, and keep try again, the goroutine of grpc is increase all the while, and memory of yurthub also increase, and kubelet will restart it when is happen OOM.
0efec082db7f88f00fa4934ba4ba366

e50ee49fda9b174b93760f01bf9c10e

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/poolcoordinator/coordinator.go
+++ b/pkg/yurthub/poolcoordinator/coordinator.go
@@ -288,9 +288,9 @@ func (coordinator *coordinator) Run() {
 					continue
 				}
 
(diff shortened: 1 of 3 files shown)


rambohe-ch:
@Congrool PTAL

Congrool (commented):
/lgtm

Thanks!

Congrool (commented):
/lgtm

Thanks!

rambohe-ch (approved):

PR #1503, opened 2023-05-30
update gateway api, gateway should be distinguished into three modes，…

River-sh:
…and gateway can select multiple endpoints

> /kind enhancement
> /kind feature
> /sig network

Raven L7 need depend on gateway crd

Fixes #

```release-note

```


rambohe-ch:
@River-sh please fix golangci-lint errors.

PR #1514, opened 2023-06-05
improve yurthub get pods for cloud node

JameKeal:
/kind enhancement

User use yurtstaticset to upgrade pod, the worker pod will check the static pod status using  restful apis for yurthub like: /pods, when yurthub running in the cloud, it will not return pods because the yurthub not cache info for cloud mode.

So yurthub is cloud mode, when get pods, it can use informer cache to return pod list.

Fixes #

```release-note

```

### DIFF
--- a/cmd/yurthub/app/config/config.go
+++ b/cmd/yurthub/app/config/config.go
@@ -302,12 +302,20 @@ func registerInformers(options *options.YurtHubOptions,
 
 	if tenantNs != "" {
 		newSecretInformer := func(client kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
-
 			return coreinformers.NewFilteredSecretInformer(client, tenantNs, resyncPeriod, nil, nil)
 		}
 		informerFactory.InformerFor(&corev1.Secret{}, newSecretInformer)
 	}
 
+	if workingMode == util.WorkingModeCloud {
+		newPodInformer := func(client kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
+			listOptions := func(ops *metav1.ListOptions) {
+				ops.FieldSelector = fields.Set{"spec.nodeName": options.NodeName}.String()
+			}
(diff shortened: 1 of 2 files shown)


cmd/yurthub/app/config/config.go
```diff
@@ -308,6 +308,10 @@ func registerInformers(options *options.YurtHubOptions,
 		informerFactory.InformerFor(&corev1.Secret{}, newSecretInformer)
 	}
 
+	newPodInformer := func(client kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
+		return coreinformers.NewPodInformer(client, "", resyncPeriod, nil)
+	}
+	informerFactory.InformerFor(&corev1.Pod{}, newPodInformer)
```
rambohe-ch, line 314: we need new pod informer when working mode is cloud, so we can add a condition as following in order to reduce costs for yurthub.

```
if workingMode == util.WorkingModeCloud {
	newPodInformer := func(client kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
		return coreinformers.NewPodInformer(client, "", resyncPeriod, nil)
	}
	informerFactory.InformerFor(&corev1.Pod{}, newPodInformer)
}
```

cmd/yurthub/app/config/config.go
```diff
@@ -308,6 +308,12 @@ func registerInformers(options *options.YurtHubOptions,
 		informerFactory.InformerFor(&corev1.Secret{}, newSecretInformer)
 	}
 
+	if workingMode == util.WorkingModeCloud {
+		newPodInformer := func(client kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
+			return coreinformers.NewPodInformer(client, "", resyncPeriod, nil)
```
rambohe-ch, line 313: how about set the list options to a fieldSelector `obj.Spec.NodeName=nodeName`? so only pods for this node will be returned and efficiency can be improved.

rambohe-ch (approved):

Issue #1474, opened 2023-05-21
[feature request]improve yurthub artifacts installation


rambohe-ch:
**What would you like to be added**:
In the tutorial: https://openyurt.io/docs/next/installation/manually-setup#32-install-yurthub-artifacts, if user has not set `kubernetesServerAddr` value in yurthub artifacts installation, `yurtadm join` command should use the correct kubernetes server address to replace `https://127.0.0.1:6443`.

**Why is this needed**:
It is easy for user to forget to set `kubernetesServerAddr` value when install yurthub artifacts, and this will cause the execution of the `yurtadm join` command to fail.

**others**
/kind feature

Lan-ce-lot:
@rambohe-ch I want to try this.

PR #1517, opened 2023-06-05
feat: use real kubernetes server address to yurthub when yurtadm join

Lan-ce-lot:
In the scenario where the user forgets to set the yurthub server address for the master node, the yurtadm join command should replace the yurthub parameter with the address obtained from the join.

Fixes #
#1474 

```release-note

```

### DIFF
--- a/pkg/yurtadm/util/yurthub/yurthub.go
+++ b/pkg/yurtadm/util/yurthub/yurthub.go
@@ -17,6 +17,8 @@ limitations under the License.
 package yurthub
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"net/http"
@@ -73,6 +75,12 @@ func AddYurthubStaticYaml(data joindata.YurtJoinData, podManifestPath string) er
 	if err != nil {
 		return err
 	}
+
+	yurthubTemplate, err = useRealServerAddr(yurthubTemplate, kubernetesServerAddrs)
+	if err != nil {
+		return err
+	}
+
 	yurthubManifestFile := filepath.Join(podManifestPath, util.WithYamlSuffix(data.YurtHubManifest()))
 	klog.Infof("yurthub template: %s\n%s", yurthubManifestFile, yurthubTemplate)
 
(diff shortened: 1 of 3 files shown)


pkg/yurtadm/util/yurthub/yurthub.go
```diff
@@ -170,3 +177,31 @@ func CleanHubBootstrapConfig() error {
 	}
 	return nil
 }
+
+// useRealServerAddr check if the server-addr from yurthubTemplate is default value: 127.0.0.1:6443
+// if yes, we should use the real server addr
+func useRealServerAddr(yurthubTemplate *string, kubernetesServerAddrs string) error {
```
rambohe-ch, line 183: `func useRealServerAddr(yurthubTemplate *string, kubernetesServerAddrs string) error` --> `func useRealServerAddr(yurthubTemplate string, kubernetesServerAddrs string) (string, error)`

pkg/yurtadm/util/yurthub/yurthub.go
```diff
@@ -170,3 +177,31 @@ func CleanHubBootstrapConfig() error {
...
+		}
+	}
+
+	// Convert the modified podObj back to YAML
+	modifiedYaml, err := yaml.Marshal(podObj)
+	if err != nil {
+		return fmt.Errorf("Failed to marshal modified YAML: %v", err)
+	}
+
+	*yurthubTemplate = string(modifiedYaml)
```
rambohe-ch, line 205: I think it's not a good idea to use yaml marshal to replace server address, maybe you can reference the following code:
```
	scanner := bufio.NewScanner(bytes.NewReader(fileContent))
	var buffer bytes.Buffer
	for scanner.Scan() {
		line := scanner.Text()
		if strings.Contains(line, targetURL) {
			line = strings.Replace(line, targetURL, replacementURL, -1)
		}
		buffer.WriteString(line + "\n")
	}

	if err := scanner.Err(); err != nil {
		klog.Infof("Error scanning file: %v\n", err)
		return
	}
```

rambohe-ch (approved):

PR #1526, opened 2023-06-07
fix yurtstaticset workerpod reset error

JameKeal:
/kind bug

When user have more than one static pod want to upgrade, and the upgrade worker pod name has the same prefix in different yurtstaticset, when the upgradeinfo want to calculate information, it will confused.

So we need to add yurtstaticset name for worker pod to avoid this condition, and short for worker pod prefix to avoid the worker pod name is too long to make mistake

Fixes #

/assign @rambohe-ch 

```release-note

```


JameKeal:
@xavier-hou PTAL

rambohe-ch (approved):

rambohe-ch:
@JameKeal I think that it is a good idea to add name of yss for upgrade-worker pod. agree +1

PR #1554, opened 2023-06-18
feat: remove watch nodepool in service topology controller

### LINKED ISSUE #1553: [feature request]There is no need to list/watch nodepool in servicetoplogy controller
1. As discussed in the issue: #1382  and pull request #1444  , node pool name for a node can not be changed after joining into the cluster. if end user wants to change the nodepool name of node, node should be deleted from the cluster and then re-join the node for new nodepool into the cluster.

2. Because of node deletion and joining, endpoints and endpointslices will be updated explicitly, so there is no need to list/watch nodepool in service topology controller in order to update endpoints and endpointslices when nodepool is changed.

so i want to remove list/watch nodepool action in service topology controller.

**Why is this needed**:
The efficiency of service topology feature can be improved if nodepool list/watch is removed.
moreover, nodepool list/watch is a dummy function for service topology controller as above description.

**others**
/kind feature

rambohe-ch:
/kind feature

As discussed in #1553 , The efficiency of service topology feature can be improved if nodepool list/watch is removed.
moreover, nodepool list/watch is a dummy function for service topology controller. 

so i want to remove list/watch nodepool function in service topology controller.

Fixes #1553 

```release-note

```

### DIFF
--- a/pkg/controller/servicetopology/adapter/adapter.go
+++ b/pkg/controller/servicetopology/adapter/adapter.go
@@ -25,7 +25,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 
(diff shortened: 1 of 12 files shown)


rambohe-ch:
@YTGhost @JameKeal PTAL

YTGhost (commented):
/lgtm

PR #1558, opened 2023-06-19
update gateway api, gateway can enable ravenl3/ravenl7 and can select…

River-sh:
… multiple endpoints for ravenl7

> /kind enhancement
> /kind feature
> /sig network

1. support to enable/unenable raven l7 proxy 
2. support to enable/unenable raven l3 tunnel
3. support for electing multiple replicas
4. support watch lb service for update the public ip for exposed gateway

Fixes #

```release-note

```


pkg/apis/raven/v1alpha1/gateway_types.go
```diff
@@ -32,60 +29,94 @@ const (
...
+	// ActiveEndpoints is the reference of the active endpoint.
+	ActiveEndpoints []*Endpoint `json:"activeEndpoints,omitempty"`
 }
 
 // +genclient
 // +k8s:openapi-gen=true
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:scope=Cluster,path=gateways,shortName=gw,categories=all
-//+kubebuilder:printcolumn:name="ActiveEndpoint",type=string,JSONPath=`.status.activeEndpoint.nodeName`
```
njucjc, line 88: without this line, `ActiveEndponts` will not display when you get it by `kubectl`

luckymrwang:
@River-sh The gateway CRD has been modified some fields and type. And does it support older version not a new cluster?

River-sh:
@luckymrwang  The new gateway crd can support the old version, which only adds the Spec attribute

pkg/apis/raven/v1alpha1/gateway_types.go
River-sh: The gateway supports multiple ActiveEndpoints, therefore, how to display them is not considered for the time being

PR #1574, opened 2023-06-28
set file system permission via yurtadm to default

### LINKED ISSUE #1573: [BUG] yurtadm tries to set the incorrect permission, which leads to unexpected failure
**What happened**:

When using OpenYurt with Cilium CNI, cilium-agent (DaemonSets) initialization cannot be completed since it does not have the access permission to `/opt/cni/bin`.

see more details for https://github.com/cilium/cilium/issues/22933

OpenYurt can enable Cilium w/o any problems.

1. `yurtadm join` from edge device. (kubelet should not be installed, so that `yurtadm` creates the `/opt/cni/bin` with `0600` permission)
2. enable cilium with `cilium install`.
3. the edge node cannot be `READY` since cilium-agent will fail to initialize.

In general directory permission `0600` does not make sense to me, since the owner cannot change the directory in it.
In addition, OpenYurt should not put the different access permission in default, since this could lead the unexpected problems like this. saying, `/opt/cni/bin` is not dedicated directory for OpenYurt but anyone could read and execute.

**Environment**:
- OpenYurt version: v1.1.0
- Kubernetes version (use `kubectl version`): 1.22.13
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04
- Kernel (e.g. `uname -a`): 5.4.0-126-generic
- Install tools:
- Others:

**others**

/kind bug

fujitatomoya:
#### What type of PR is this?

/kind bug

address https://github.com/openyurtio/openyurt/issues/1573

It does use the default file access permission.

- `/opt/cni/bin`

```bash
tomoyafujita@~ >ls -lt /opt/cni/
total 4.0K
drwxr-xr-x 2 root root 4.0K Jun  9 15:43 bin/
```

- `/etc/systemd/system/kubelet.service.d/10-kubeadm.conf`

```bash
tomoyafujita@~ >ls -lt /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
-rw-r----- 1 root root 898 Sep 21  2022 /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```

Fixes #1573 

- I believe this makes sense, and avoid unexpected problem such as https://github.com/cilium/cilium/issues/22933
- Since we found this issue based on `v1.1.0`, current target to set so. I am happy to address master if you guys want me to do so, then we can backport the fix for each release.

### DIFF
--- a/pkg/yurtadm/util/kubernetes/util.go
+++ b/pkg/yurtadm/util/kubernetes/util.go
@@ -207,7 +207,7 @@ func SetKubeletUnitConfig() error {
 		}
 	}
 
-	if err := os.WriteFile(constants.KubeletServiceConfPath, []byte(constants.KubeletUnitConfig), 0600); err != nil {
+	if err := os.WriteFile(constants.KubeletServiceConfPath, []byte(constants.KubeletUnitConfig), 0640); err != nil {
 		return err
 	}
 
--- a/pkg/yurtadm/util/kubernetes/util.go
+++ b/pkg/yurtadm/util/kubernetes/util.go
@@ -161,7 +161,7 @@ func CheckAndInstallKubelet(kubernetesResourceServer, clusterVersion string) err
 		klog.V(1).Infof("Skip download cni, use already exist file: %s", savePath)
 	}
 
-	if err := os.MkdirAll(constants.KubeCniDir, 0600); err != nil {
+	if err := os.MkdirAll(constants.KubeCniDir, 0755); err != nil {
 		return err
 	}
 	if err := util.Untar(savePath, constants.KubeCniDir); err != nil {


fujitatomoya (commented):
I think same fixes can be applied to master branch as well. if we should fix that on master branch and backporting the fix for each release, that i can help with.

YTGhost:
> I think same fixes can be applied to master branch as well. if we should fix that on master branch and backporting the fix for each release, that i can help with.

Yes, we should fix that on master branch and backporting the fix for each release. And maybe you should submit the PR to the master branch, and then after merge we can have github-action automatically backport it via the backport command

fujitatomoya:
@YTGhost okay that is what i though.

my concern was that backport would not be straight-forward since there is huge difference between `master` and `v1.1.0`.

anyway, i understand the procedure, fix will be resubmitted targeted to master branch.

fujitatomoya:
@YTGhost i will go ahead to close this one since i created the replacement https://github.com/openyurtio/openyurt/pull/1576 targeted to master branch.

PR #1594, opened 2023-07-10
Features/dns controller

River-sh:
The DNS Controller maintains a record of the NodeName mapping to the Raven L7 Proxy IP address in a configmap and mounts it into CoreDNS to support cross pool communication

Fixes #

```release-note

```


pkg/controller/raven/dns/dns_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	for _, addr := range node.Status.Addresses {
+		if addr.Type == corev1.NodeInternalIP {
+			ip := net.ParseIP(addr.Address)
+			if ip != nil {
+				allIPs = append(allIPs, ip)
+				break
+			}
+		}
+	}
+	for _, addr := range node.Status.Addresses {
```
fengshunli, line 252: reuse the same for loop, add else if to handle the same logic, reduce redundant code

fengshunli (changes requested):

pkg/controller/apis/config/types.go
```diff
@@ -35,7 +36,7 @@ type YurtManagerConfiguration struct {
 	NodePoolController nodepoolconfig.NodePoolControllerConfiguration
 
 	// GatewayControllerConfiguration holds configuration for GatewayController related features.
-	GatewayController gatewayconfig.GatewayControllerConfiguration
+	GatewayController gatewayconfig.GatewayPickupControllerConfiguration
```
fengshunli, line 39: renmae GatewayPickupController

cmd/yurt-manager/app/options/ravendnscontroller.go
```diff
@@ -0,0 +1,60 @@
...
+		return
+	}
+
+	//fs.BoolVar(&n.CreateDefaultPool, "create-default-pool", n.CreateDefaultPool, "Create default cloud/edge pools if indicated.")
+}
+
+// ApplyTo fills up nodepool config with options.
+func (o *RavenDNSControllerOptions) ApplyTo(cfg *config.RavenDNSControllerConfiguration) error {
+	if o == nil {
+		return nil
```
fengshunli, line 47: return err

pkg/controller/raven/dns/dns_controller.go
```diff
@@ -0,0 +1,265 @@
...
+			if ip != nil {
+				allIPs = append(allIPs, ip)
+				break
+			}
+		}
+	}
+	if len(allIPs) == 0 {
+		return "", fmt.Errorf("host IP unknown; known addresses: %v", node.Status.Addresses)
+	}
+	return allIPs[0].String(), nil
```
fengshunli, line 264: add nil value judgment

fengshunli (changes requested):

rambohe-ch:
@River-sh I think it is not a good idea to add several controllers in only one pull request.  because it is not easy to review so much code in only one pull request. would you like to split this pull request and post one controller one by one?

Issue #1435, opened 2023-05-08
feat: configuration rendering engine for multi-regions


rambohe-ch:
## Motivation

OpenYurt is an open source project in the edge cloud native domain, which unifies the management of scattered massive edge heterogeneous resources (such as CDN sites, IoT machines and other edge computing power) from the control side (located in the cloud or central server room, etc.) to help users easily complete large-scale application delivery, operation and maintenance, and control on massive edge resources. At present, OpenYurt has implemented unified management of multi-territory workloads with CRD resources such as NodePool, YurtAppSet and YurtAppDaemon. Due to the objective existence of heterogeneous environment such as resource configuration and network topology in each region, it is necessary to render different configurations for each regional workload, including the number of instances, image version, ConfigMap, Secret, PVC, etc.

## Objectives
The primary objectives of this project are to:
1. Support rendering personalized configurations for multi-regions workloads (e.g. YurtAppSet, YurtAppDaemon)
2. Multi-regions configuration rendering engine has good scalability, when there is a new workload CRD for multi-regions management, the rendering engine can also be adapted in a friendly way.

## Output Requirements

1. Design documentation and code implementation of the multi-regions configuration rendering engine to support the rendering of YurtAppSet, YurtAppDaemon and other resources
2. Complete test cases for the rendering engine (including UT and E2E) to ensure that the rendering engine can run robustly
3. Complete the installation and deployment of the rendering engine in the OpenYurt cluster, and put it online in the OpenYurt community experience center.

## Related issues
1. https://github.com/openyurtio/openyurt/issues/1236
2. https://summer-ospp.ac.cn/org/prodetail/235fc0262?list=org&navpage=org

PR #1596, opened 2023-07-10
feat: provide users with the ability to customize the edgex framework

### LINKED ISSUE #1595: [feature request] Provide users with the ability to customize the edgex framework
Now that we are able to support the latest version of edgex, we need to export the current configuration to configmap so that users can customize the framework according to their requirements.

**Why is this needed**:
Give advanced users the ability to DIY their own framework.

**others**
/kind feature

LavenderQAQ:
#### What type of PR is this?
/kind feature
/sig iot

This pr will give users the ability to customize the edgex framework. We write the standard configuration collected by the auto-collector into a configmap called "platformadmin-framework". Users can customize their edgex by editing this configmap.

Fixes #1595

```release-note
# Support for customizing edgex configurations by modifying configmap
kubectl edit cm platformadmin-framework
```

### DIFF
--- a/pkg/controller/platformadmin/platformadmin_controller.go
+++ b/pkg/controller/platformadmin/platformadmin_controller.go
@@ -28,10 +28,12 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
+	"k8s.io/kubectl/pkg/scheme"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


LavenderQAQ:
At present, the isolation of PlatformAdmin is not enough. It can only be configured on a namespace basis. When two edgex versions appear in a namespace, overwriting problems may occur. For this problem we need to combine #1435 to solve.

pkg/controller/platformadmin/platformadmin_controller.go
```diff
@@ -600,3 +640,115 @@ func annotationToComponent(annotation map[string]string) ([]*config.Component, e
...
+	}
+
+	// Create the configmap that represents framework
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      platformAdminFramework.name,
+			Namespace: platformAdmin.Namespace,
+		},
+	}
+	cm.Labels = make(map[string]string)
```
Rui-Gan, line 739:
```
        cm.Labels[iotv1alpha2.LabelPlatformAdminGenerate] = LabelFramework
	cm.Data = make(map[string]string)
	cm.Data["framework"] = string(data)
```
The above part of the code needs to be placed in the CreateOrUpdate function, otherwise configmap cannot be updated.

Rui-Gan (changes requested):

Rui-Gan (approved):

rambohe-ch (approved):

Issue #1597, opened 2023-07-10
[OSPP 2024] Build iot system configuration isolation on nodepool


LavenderQAQ:
**What would you like to be added**:
Provides multiple PlatformAdmin deployment based on the same namespace, and allows users to customize the configuration of a nodepool.

**Why is this needed**:
We gave users the ability to customize iot systems in #1595, but it's currently not isolated enough for nodepool.
This is a very old need, refer to [multi edgex instance support](https://github.com/openyurtio/yurt-edgex-manager/issues/10). OpenYurt extends the concept of nodepools on top of k8s, so an ideal deployment is that users can configure each nodepool iot system independently. With #1435 we can manipulate yurtappset to configure each nodepool individually. And service topology allows us to separate the traffic from each nodepool. With these two capabilities we can take a step closer to idealizing the current deployment model.

_Finally, there is the tricky problem that we cannot split the configuration of services by nodepool (because edgex internal components access the same set of dns). At the moment we don't think there's enough incentive to solve this problem._

**others**
/kind feature

rambohe-ch:
@LavenderQAQ Do you have any progress about this issue?

LavenderQAQ:
@rambohe-ch I am really busy recently. I plan to write a proposal first, which may be after March.

WoShiZhangmingyu:
Hello，I am Zhangmingyu。Please

PR #1600, opened 2023-07-10
Proposal for Multi-region workloads configuration rendering engine

### LINKED ISSUE #1435: feat: configuration rendering engine for multi-regions

vie-serendipity:
kind feature

Fixes #1435 

```release-note

```


docs/proposals/20230706-yurtappconfigurationreplacement.md
```diff
@@ -0,0 +1,291 @@
...
+2. In order to inject item into the workloads, we should create a new CRD named YurtAppConfigurationReplacemnet, which consist of replacements. Each replacement replace a set of configuration for matching nodepools. 
+<img src = "../img/yurtappconfigurationreplacement/Inspiration.png" width="600">
+
+### YurtAppConfigurationReplacement API
+1. YurtAppConfigurationReplacement needs to be bound to YurtAppSet/YurtAppDaemon. 
+Considering that there are multiple Deployment/StatefulSet per nodepool, as shown below, it must be bound to YurtAppSet/YurtAppDaemon for injection. We use subject field to bind it to YurtAppSet/YurtAppDaemon. 
+2. YurtAppConfigurationReplacement is responsible for injection of Replacement. We only need to create a new YurtAppConfigurationReplacement resource for all nodepools under a YurtAppSet/YurtAppDaemon resource.
```
LavenderQAQ, line 67: Can I think the YurtAppConfigurationReplacement and YurtAppSet is a one-to-one relationship?

vie-serendipity: Yes

docs/proposals/20230706-yurtappconfigurationreplacement.md
```diff
@@ -0,0 +1,291 @@
...
+	// SecretSource represents secret name.
+	SecretSource string `json:"secretSource"`
+	// SecretTarget represents the Secret corresponding to the volume above.
+	// SecretTarget supprot advanced features like wildcard. 
+	// By naming secret as secretName-{{nodepool}}, all secret can be injected at once.
+	SecretTarget string `json:"secretTarget"`
+}
+
+// Item represents configuration to be injected.
+// Only one of its members may be specified.
```
LavenderQAQ, line 119: What does specifying one of its members mean here? Can I overwrite container images and environment variables at the same time?

docs/proposals/20230706-yurtappconfigurationreplacement.md
```diff
@@ -0,0 +1,291 @@
...
+// Replacement describe a set of nodepools and their shared or identical configurations
+type Replacement struct {
+	Pools []string `json:"pools"`
+	Items []Item   `json:"items"`
+}
+
+type Subject struct {
+	metav1.TypeMeta `json:",inline"`
+	// NameSpace is the namespace of YurtAppSet or YurtAppDaemon
+	NameSpace string `json:"nameSpace"`
```
LavenderQAQ, line 139: Can this be as consistent as possible with Metadata? I want this to be unified with the Namespace of k8s, so that we can call it more easily.
```go
// Namespace defines the space within which each name must be unique. An empty namespace is
// equivalent to the "default" namespace, but "default" is the canonical representation.
// Not all objects are required to be scoped to a namespace - the value of this field for
// those objects will be empty.
//
// Must be a DNS_LABEL.
// Cannot be updated.
// More info: http://kubernetes.io/docs/user-guide/namespaces
// +optional
Namespace string `json:"namespace,omitempty" protobuf:"bytes,3,opt,name=namespace"`
```

docs/proposals/20230706-yurtappconfigurationreplacement.md
vie-serendipity: The reason only one can be specified here is mainly for the sake of a clear hierarchy. Each Item is responsible for only one field. Otherwise, some items contain multiple fields and some have only one field, which is not user-friendly. The relevant validations are performed in the webhook. I think this will be better.

docs/proposals/20230706-yurtappconfigurationreplacement.md
vie-serendipity: You're right. It is more reasonable to use the Namespace of k8s.

docs/proposals/20230706-yurtappconfigurationreplacement.md
```diff
@@ -0,0 +1,291 @@
...
+Here is a scenario. Ten nodepools only need to be configured differently for configMap, it is not user-friendly to write ten replacements by using current scheme. 
+We add a new feature like wildcard to simplify the usage. We can name all the configMap as configMap-{{nodepool}} so that we can inject the configMap into nodepool in the suffix. 
+```yaml
+replacements:
+- pools:
+    *
+  items:
+  - configMap:
+      configMapSource: configMap-demo
+      configMapTarget: configMapName-{{nodepool}}
```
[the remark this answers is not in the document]
vie-serendipity: Here, the configMapSource is the name of ConfigMap in the original template, the configMapTarget is newly written externally and will be injected into. So I think you got it backwards. 
Do you think "source" should correspond to the newly written configMap?

docs/proposals/20230706-yurtappconfigurationreplacement.md
LavenderQAQ: Thank you😸. I understand. But I'm a little confused. What should webhook do if I specify multiple items for configmap? What if there are multiple image items?🧐

docs/proposals/20230706-yurtappconfigurationreplacement.md
LavenderQAQ: 👌

docs/proposals/20230706-yurtappconfigurationreplacement.md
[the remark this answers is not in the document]
LavenderQAQ: I got it. I got them backwards. Thank you for your explanation.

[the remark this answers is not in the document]
LavenderQAQ: This may confirm that it is easy to look backwards, and then we can discuss it together.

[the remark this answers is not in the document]
vie-serendipity: Ok, we can communicate it on Wednesday.

docs/proposals/20230706-yurtappconfigurationreplacement.md
vie-serendipity: Validation webhook will only do validation for an item containing one element. 
If there are multiple items that all are configMapItem, and their configMapSource are similar, the finally injected configMap will be the one that appears last. 
For image, we can have multiple items to configure different images based on different containerName. If their containerName are similar, the image appears last will work.

LavenderQAQ: Thank you. I get it.

rambohe-ch:
@vie-serendipity please fix the markdown misspell errors.

docs/proposals/20230706-yurtappconfigurationreplacement.md
LavenderQAQ: @vie-serendipity @rambohe-ch I think it's reasonable. While nodepool is cluster-level, the yurtappdaemons, yurtappsets, and yurtstaticsets that need to be modified are all namespace-level. If we assume YurtAppConfigurationReplacement is a namespace level of CRD, I don't think cross-namespace modification should be allowed.
By the way, we're not going to set it to cluster-level CRD, are we?

docs/proposals/20230706-yurtappconfigurationreplacement.md
rambohe-ch: @vie-serendipity I think it is a good idea to add this relationship description in the proposal.

LavenderQAQ:
@vie-serendipity @rambohe-ch Have we considered any other names? This one is so long that I don't even want to type it😹. Such as `YurtAppRender`? It makes the code seem less verbose. `YurtAppCoonfigurationReplacement` really is very long, and we are likely to be often call it in the code.

docs/proposals/20230706-yurtappconfigurationreplacement.md
vie-serendipity: @LavenderQAQ  It should be a namespace level CRD. I agree with @rambohe-ch , we can remove the namespace field in the Subject.

docs/proposals/20230706-yurtappconfigurationreplacement.md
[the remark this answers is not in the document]
y-ykcir: Does the mode of Source and resource only support the **replace** rendering way and do not support injecting new configmap or PVC? Do we not consider supporting the **add** rendering way？

vie-serendipity:
> @vie-serendipity @rambohe-ch Have we considered any other names? This one is so long that I don't even want to type it😹. Such as `YurtAppRender`, `YurtAppBinding` or something? It makes the code seem less verbose. `YurtAppConfigurationReplacement` really is very long, and we are likely to be often call it in the code. Or we could scale it down a little bit, `YurtAppConfigReplacement`?

@rambohe-ch @LavenderQAQ It does seem a bit too long, we can discuss it at the community meeting. We can think of a name that is both clear and brief. I propose YurtAppConfigRender as candidate.

my0sotis:
> @vie-serendipity @rambohe-ch Have we considered any other names? This one is so long that I don't even want to type it😹. Such as `YurtAppRender`, `YurtAppBinding` or something? It makes the code seem less verbose. `YurtAppConfigurationReplacement` really is very long, and we are likely to be often call it in the code. Or we could scale it down a little bit, `YurtAppConfigReplacement`?

Agree. Maybe we can name it according its purpose, such as `YurtRegionalConfig`?

y-ykcir:
> > @vie-serendipity @rambohe-ch Have we considered any other names? This one is so long that I don't even want to type it😹. Such as `YurtAppRender`, `YurtAppBinding` or something? It makes the code seem less verbose. `YurtAppConfigurationReplacement` really is very long, and we are likely to be often call it in the code. Or we could scale it down a little bit, `YurtAppConfigReplacement`?
> 
> @rambohe-ch @LavenderQAQ It does seem a bit too long, we can discuss it at the community meeting. We can think of a name that is both clear and brief. I propose YurtAppConfigRender as candidate.

Since `render` as a noun does not seem to have the meaning of 'xuanran', perhaps using `renderer` is more sensible😂

docs/proposals/20230706-yurtappconfigurationreplacement.md
rambohe-ch: @LavenderQAQ YurtAppConfigurationReplacement is used to handle namespace scoped resource(such as Deployment) in NodePool, so it is reasonable to define YurtAppConfigurationReplacement as namespace scoped resource.

LavenderQAQ:
I have a few more questions. I'm sorry if I have to clarify the background, currently every nodepool iot system is distributed through yurtappset, and we recently introduced the ability to customize iot systems, which will bring a lot of scenarios for nodepool configuration changes. I think the application scenario of this proposal should be here, but I found that it seems that these items can not cover all fields (for example, in our case, users can modify any field in the depolyment template, Examples include ports, ImagePullPolicy, hostname, etc.), and splitting the native deployment into individual items can also increase code complexity. This may be a potential problem, so we may only rely on nodepool's patch function (we will put forward a proposal to modify the current iot architecture in the future), which will be complicated to explain. I can demonstrate the iot scenario at the community meeting. Perhaps this proposal is not suitable for iot-sig, or we could think about adapting iot systems on Wednesday.

LavenderQAQ:
> > > @vie-serendipity @rambohe-ch Have we considered any other names? This one is so long that I don't even want to type it😹. Such as `YurtAppRender`, `YurtAppBinding` or something? It makes the code seem less verbose. `YurtAppConfigurationReplacement` really is very long, and we are likely to be often call it in the code. Or we could scale it down a little bit, `YurtAppConfigReplacement`?
> > 
> > 
> > @rambohe-ch @LavenderQAQ It does seem a bit too long, we can discuss it at the community meeting. We can think of a name that is both clear and brief. I propose YurtAppConfigRender as candidate.
> 
> Since `render` as a noun does not seem to have the meaning of 'xuanran', perhaps using `renderer` is more sensible😂

In my opinion, `renderer` sounds like a controller, but this crd is just a substitution operation, it's kind of like a binding relationship, not very independent😹.

docs/proposals/20230706-yurtappconfigurationreplacement.md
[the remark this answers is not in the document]
vie-serendipity: It is true that the current scheme does not support the injection of new configMap. The current API may not be applicable. 
On the one hand, users can add key-value pairs to the configMapTarget. On the other hand, if one deployment need injection all of PVC, secret, and configMap, would it be more appropriate to create a separate new deployment rather than deploying in a YurtAppSet&YurtAppConfigurationReplacement way?
This is our @rambohe-ch understanding.

LavenderQAQ:
> I have a few more questions. I'm sorry if I have to clarify the background, currently every nodepool iot system is distributed through yurtappset, and we recently introduced the ability to customize iot systems, which will bring a lot of scenarios for nodepool configuration changes. I think the application scenario of this proposal should be here, but I found that it seems that these items can not cover all fields (for example, in our case, users can modify any field in the depolyment template, Examples include ports, ImagePullPolicy, hostname, etc.), and splitting the native deployment into individual items can also increase code complexity. This may be a potential problem, so we may only rely on nodepool's patch function (we will put forward a proposal to modify the current iot architecture in the future), which will be complicated to explain. I can demonstrate the iot scenario at the community meeting. Perhaps we can consider on Wednesday how this proposal fits in with the needs of the current iot system.

@vie-serendipity I have looked all architecture and I think it can be solved. For fields in the proposal capability, we support user dynamic configuration. For fields not in the proposal capability, users can delete yurtappset to trigger our reconstruction.  If you are interested in this, you can also work with the iot-sig for this in the future.

vie-serendipity:
> @vie-serendipity I have looked all architecture and I think it can be solved. For fields in the proposal capability, we support user dynamic configuration. For fields not in the proposal capability, users can delete yurtappset to trigger our reconstruction. If you are interested in this, you can also work with the iot-sig for this in the future (refer to #1597).

Yeah, of course. But I don't understand this. 
> For fields not in the proposal capability, users can delete yurtappset to trigger our reconstruction

Could you explain more specifically?

LavenderQAQ:
> > @vie-serendipity I have looked all architecture and I think it can be solved. For fields in the proposal capability, we support user dynamic configuration. For fields not in the proposal capability, users can delete yurtappset to trigger our reconstruction. If you are interested in this, you can also work with the iot-sig for this in the future (refer to #1597).
> 
> @LavenderQAQ Yeah, of course. But I don't understand this.
> 
> > For fields not in the proposal capability, users can delete yurtappset to trigger our reconstruction
> 
> Could you explain more specifically?

@vie-serendipity We deploy iot systems according to a standard configuration, which is also written into a configmap. The entire iot system is refreshed in PlatformAdmin's control flow based on the configmap configuration, which users can modify to customize their iot framework. This architecture gives users a lot of configuration rights. They can mount PV, modify images, modify ports, and so on. The above means that when the proposal cannot dynamically update the configuration, the user can delete yurtappset once to trigger a refresh.

pkg/controller/yurtappconfigrender/yurtappconfigrender_controller_test.go
```diff
@@ -0,0 +1,22 @@
+/*
```
rambohe-ch, line 1: please remove this test file.

docs/proposals/20230706-yurtappoverrider.md
```diff
@@ -0,0 +1,334 @@
...
+
+// Item represents configuration to be injected.
+// Only one of its members may be specified.
+type Item struct {
+    // +optional
+    Image *ImageItem `json:"image,omitempty"`
+    // +optional
+    Replicas *int32 `json:"replicas,omitempty"`
+    // +optional
+    UpgradeStrategy *string `json:"upgradeStrategy,omitempty"`
```
rambohe-ch, line 95: how about remove `UpgradeStrategy`? because only override `UpgradeStrategy` is not enough, other settings like   MaxUnavailable and MaxSurge should also be overrided.

docs/proposals/20230706-yurtappoverrider.md
```diff
@@ -0,0 +1,332 @@
...
+// Only one of its members may be specified.
+type Item struct {
+    // +optional
+    Image *ImageItem `json:"image,omitempty"`
+    // +optional
+    Replicas *int32 `json:"replicas,omitempty"`
+}
+
+type Operator string
+
```
gnunu, line 97: How about use Operation? Operator is a little misleading.

vie-serendipity: Fine, that would be better.

docs/proposals/20230706-yurtappoverrider.md
```diff
@@ -0,0 +1,332 @@
...
+#### YurtAppOverrider Controller
+##### Task 1
+1. Get update events by watching the YurtAppOverrider resource
+2. Trigger the Deployment mutating webhook by modifying an annotation or label
+##### Task 2
+1. Get delete events(delete members of pools) by watching the YurtAppOverrider resource
+2. Render the configuration according to the YurtAppSet workload template and the watching YurtAppOverrider
+### User Stories
+#### Story 1 (General)
+Use YurtAppSet with YurtAppOverrider for customized configuration of each region. Create YurtAppOverrider first and then create YurtAppSet. If update is needed, modify Yu  resource directly. For YurtAppDaemon, the usage is similar. Users only need to do some  configurations in YurtAppOverrider and our rendering engine will inject all configurations into target workloads.
```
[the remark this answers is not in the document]
vie-serendipity: It should be YurtAppSet. Sorry for that.

docs/proposals/20230706-yurtappoverrider.md
```diff
@@ -0,0 +1,332 @@
...
+1. If the intercepted Deployment's ownerReferences field is empty, filter it directly
+2. Find the corresponding YurtAppOverrider resource by ownerReferences, if not, filter directly
+3. Find the entries involved, get the corresponding configuration, and inject them into workloads.
+
+Attention Points:
+1. Note that injection is implemented by recalculating the final configuration according to the YurtAppSet workload template and the watching YurtAppOverrider
+2. The latter configuration always relpace the former. So the last configuration will really work
+#### YurtAppOverrider Validating Webhook
+1. Verify that only one YurtAppOverrider can be bound to YurtAppSet/YurtAppDaemon
+2. Verify that value is empty when operator is REMOVE
```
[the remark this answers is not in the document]
vie-serendipity: Roger.

rambohe-ch (approved):

PR #1604, opened 2023-07-11
fix: add platformadmin crd to the manifests script

LavenderQAQ:
#### What type of PR is this?
/kind bug
/sig iot

Add platformadmin crd to the manifests script.


LavenderQAQ:
@rambohe-ch  Looks like there's something wrong with UT? But it shouldn't be here. Maybe it needs a rerun?

LavenderQAQ:
This problem has been resolved in #1607.

PR #1625, opened 2023-07-20
add kubelet certificate mode in yurthub

### LINKED ISSUE #1618: [feature request]Support for bootstrapping yurthub component using node certificates.
In addition to the current use of bootstrap tokens to bootstrap start yurthub, I'd like to support the use of node certificates to bootstrap start yurthub, so that yurthub can run on nodes as a regular pod rather than a static pod.

**Why is this needed**:
In some scenarios, kubelet may not need to use yurthub, but other components do, so that yurthub can use the kubelet's node certificates instead of having to apply for a new one.

**others**
/kind feature

rambohe-ch:
/kind feature

add a new bootstrap mode: `kubeletcertificate` for yurthub component, so yurthub can use kubelet certificate to work.

Fixes #1618 

```release-note

```


YTGhost (commented):
/lgtm

PR #1635, opened 2023-07-27
replace string var by const

dsy3502:
replace string var by const


pkg/webhook/yurtappdaemon/v1alpha1/yurtappdaemon_validation.go
```diff
@@ -39,6 +39,10 @@ import (
 	"github.com/openyurtio/openyurt/pkg/apis/apps/v1alpha1"
 )
 
+const (
+	YurtAppDaemonKind = "YurtAppDaemon"
```
rambohe-ch, line 43: how about replace all define `"YurtAppDaemon" string in the OpenYurt repo and define `YurtAppDaemonKind` in a proper file?

rambohe-ch (approved):

PR #1651, opened 2023-08-14
improve nodepool to support hostnetwork mode

### LINKED ISSUE #1620: [feature request] yurtadm add label to node with host network mode
When executing yurtadm join, first check the specified nodepool type, if it is `hostnetwork=true`, we can use yurtadm to add a parameter `--node-labels=nodepool` to `kubeadm-flags.env` .openyurt.io/host-network=true`.

**Why is this needed**:
Label nodepool.openyurt.io/hostnetwork=true add by yurtadm join.

**others**
/kind feature

### LINKED ISSUE #1622: [feature request] Add webhook to ensure the hostNetwork attribute of nodepool can't be modified
Add webhook to ensure that the hostNetwork attribute of nodepool cannot be modified  during the nodePool life cycle.

**Why is this needed**:
Just as the kubernetes pod does not allow modify the hostnetwork, we do not support midway modification of the hostnetwork of the node pool.

**others**
/kind feature

rambohe-ch:
/kind feature

Improve node pool as following:
1. add spec.HostNetwork field
2. remove spec.Selector field
3. improve node labels for nodepool(apps.openyurt.io/desired-nodepool label is removed)
4. support node conversion between v1alpha1 and v1beta1 version

Fixes #1622 
Fixes #1620 

```release-note

```


pkg/controller/nodepool/nodepool_controller.go
```diff
@@ -141,51 +130,30 @@ func (r *ReconcileNodePool) Reconcile(ctx context.Context, req reconcile.Request
 	// Note !!!!!!!!!!
 	// We strongly recommend use Format() to  encapsulation because Format() can print logs by module
 	// @kadisi
-	klog.Infof(Format("Reconcile NodePool %s/%s", req.Namespace, req.Name))
+	klog.Infof(Format("Reconcile NodePool %s", req.Name))
```
YTGhost, line 133: No need to output namespace?

pkg/yurthub/network/dummyif_test.go
```diff
@@ -19,6 +19,7 @@ limitations under the License.
 
 package network
 
+/*
```
YTGhost, line 22: Why is this unit test commented out?

pkg/controller/nodepool/nodepool_controller.go
rambohe-ch: yeah, because NodePool is cluster scope resource, so namespace is not set.

pkg/yurthub/network/dummyif_test.go
rambohe-ch: It seems that github action has changed the permission, so test case can not create dummy network interface, so in order to pass github action, i have no choice except comment this unit tests.

YTGhost (commented):
/lgtm

PR #1653, opened 2023-08-15
fix verify join token and bootstrapfile only in token bootstrap mode

rambohe-ch:
/kind bug

join token and bootstrap file are only checked when token bootstrap mode is selected.

Fixes #

```release-note

```


YTGhost (commented):
/lgtm

YTGhost (commented):
/lgtm

PR #1674, opened 2023-08-24
replace str var by const

dsy3502:
replace str var by const


YTGhost (approved):
/lgtm

rambohe-ch:
@dsy3502 please fix the conflicting files.

PR #1679, opened 2023-08-25
fix: yurt-iot-dock cannot be dynamically deployed in platformadmin

LavenderQAQ:
#### What type of PR is this?
/kind bug
/sig iot

Because the `yurt-iot-dock` is an internal component of `openyurt` and is not in the standard configuration collected by the `auto-collector`, a separate decision for the `yurt-iot-dock` is added when calculating the required components.

```release-note

```


Rui-Gan (changes requested):
We can remove the newYurtIoTDockComponent logic from initFramework as it is somewhat redundant.

Rui-Gan (approved):

rambohe-ch (approved):

PR #1680, opened 2023-08-25
[Proposal] use message-bus instead of REST to communicate with EdgeX

Pluviophile225:
In the iot scenario, OpenYurt is using EdgeX to accomplish the management of edge-side devices. EdgeX is essentially some microservices running in docker, and EdgeX itself contains related management of Device, DeviceService, and DeviceProfile. OpenYurt's current synchronization method is to traverse the device in EdgeX every five seconds, DeviceService, and DeviceProfile every five seconds, and then synchronize them to OpenYurt's CRD in Kubernetes.

The disadvantage of this approach is that if the number of devices increases, you need to spend a lot of resources on http requests every five seconds, which adds a lot of network bandwidth and puts an upper limit on scalability and the number of supported devices.

For this reason this Proposal aims to solve this problem through MessageBus.

Fixes #

```release-note

```


LavenderQAQ:
This pr corresponds to [this issue](https://github.com/openyurtio/yurt-device-controller/issues/60).

rambohe-ch (approved):

PR #1730, opened 2023-10-10
fix yurtappoverrider panic for handling normal deployments that has no ownerreferences

rambohe-ch:
/kind bug

panic will happen in yurtappoverrider when ordinary deployments without ownerreferences are handled.
so ownerreferences should be checked before yurtapporverrider deals with it.

Fixes #

```release-note

```


YTGhost (approved):
/lgtm

Issue #1545, opened 2023-06-14
[feature request] Move e2e test for PlatformAdmin from yurt-edgex-manager to yurt-manager


LavenderQAQ:
**What would you like to be added**:
add e2e test case for platformadmin_controller

**Why is this needed**:
improve e2e test coverage.

**others**
/kind feature

PR #1750, opened 2023-10-25
test: added e2e test for iot systems

LavenderQAQ:
#### What type of PR is this?
/kind enhancement
/sig iot

Fixes: #1545 

Added e2e test for iot systems.

### DIFF
--- a/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go
+++ b/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go
@@ -219,8 +219,20 @@ func (r *ReconcilePlatformAdmin) Reconcile(ctx context.Context, request reconcil
 	// resource are patched back to the API server.
 	defer func(isDeleted *bool) {
 		if !*isDeleted {
-			platformAdmin.Status = *platformAdminStatus
+			// Finally check whether PlatformAdmin is Ready
+			platformAdminStatus.Ready = true
+			if cond := util.GetPlatformAdminCondition(*platformAdminStatus, iotv1alpha2.ConfigmapAvailableCondition); cond.Status == corev1.ConditionFalse {
+				platformAdminStatus.Ready = false
+			}
+			if cond := util.GetPlatformAdminCondition(*platformAdminStatus, iotv1alpha2.ComponentAvailableCondition); cond.Status == corev1.ConditionFalse {
+				platformAdminStatus.Ready = false
+			}
+			if platformAdminStatus.UnreadyComponentNum != 0 {
+				platformAdminStatus.Ready = false
+			}
 
+			// Finally update the status of PlatformAdmin
+			platformAdmin.Status = *platformAdminStatus
 			if err := r.Status().Update(ctx, platformAdmin); err != nil {
 				klog.Errorf(Format("Update the status of PlatformAdmin %s/%s failed", platformAdmin.Namespace, platformAdmin.Name))
 				reterr = kerrors.NewAggregate([]error{reterr, err})
@@ -256,8 +268,6 @@ func (r *ReconcilePlatformAdmin) reconcileDelete(ctx context.Context, platformAd
 		return reconcile.Result{}, err
 	}
 	desiredComponents = append(desiredComponents, additionalComponents...)
-
-	//TODO: handle PlatformAdmin.Spec.Components
 
 	for _, dc := range desiredComponents {
 		if err := r.Get(
@@ -332,7 +342,6 @@ func (r *ReconcilePlatformAdmin) reconcileNormal(ctx context.Context, platformAd
 	util.SetPlatformAdminCondition(platformAdminStatus, util.NewPlatformAdminCondition(iotv1alpha2.ComponentAvailableCondition, corev1.ConditionTrue, "", ""))
 
 	// Update the metadata of PlatformAdmin
-	platformAdminStatus.Ready = true
 	if err := r.Client.Update(ctx, platformAdmin); err != nil {
 		klog.Errorf(Format("Update PlatformAdmin %s error %v", klog.KObj(platformAdmin), err))
 		return reconcile.Result{}, err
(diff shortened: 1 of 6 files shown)


test/e2e/e2e_test.go
```diff
@@ -74,6 +74,9 @@ var _ = ginkgo.BeforeSuite(func() {
 	_, err = ns.CreateNameSpace(c, constants.YurtE2ENamespaceName)
 	gomega.Expect(err).NotTo(gomega.HaveOccurred(), "fail to create namespace")
 
+	err = util.PrepareNodePoolWithNode(context.TODO(), yurtconfig.YurtE2eCfg.RuntimeClient, "openyurt-e2e-test-worker")
```
wangxye, line 77: For the sake of users or developers running e2e tests locally, I think the nodepool on the edge node needs to be destroyed during After Suite.

test/e2e/yurt/iot.go
```diff
@@ -0,0 +1,142 @@
...
+		createNamespace()
+	})
+
+	AfterEach(func() {
+		By("Cleanup resources after test")
+		By(fmt.Sprintf("Delete the entire namespaceName %s", namespaceName))
+		Expect(k8sClient.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespaceName}}, client.PropagationPolicy(metav1.DeletePropagationBackground))).Should(BeNil())
+	})
+
+	for _, testVersion := range testVersions {
```
wangxye, line 113: The version here is edgex, and the platformadmin version is supposed to follow openyurt, so is it possible to create only one platformadmin in the e2e test, and leave the adaptation for the different edgex versions to the iot-dock e2e test?

test/e2e/yurt/iot.go
```diff
@@ -0,0 +1,142 @@
...
+			})
+
+			It(fmt.Sprintf("The %s version of PlatformAdmin should be stable in ready state after it is created", version), func() {
+				By("verify the status of platformadmin")
+				Eventually(func() error {
+					testPlatfromAdmin := &iotv1alpha2.PlatformAdmin{}
+					if err := k8sClient.Get(ctx, types.NamespacedName{Name: platformAdminName, Namespace: namespaceName}, testPlatfromAdmin); err != nil {
+						return err
+					}
+					if testPlatfromAdmin.Status.Ready == true {
```
wangxye, line 133: Platformadmin deploys required components by default, but just platformadmin is ready does not mean that required components are ready.

test/e2e/e2e_test.go
LavenderQAQ: The premise here is that now if the openyurt cluster wants to delete a nodepool with a node, it needs to expel the node from the cluster and rejoin it. Implementing this set of operations in e2e introduces a lot of additional complexity and may introduce other errors. So, as a alternative solution, I need to create a nodepool with nodes globally and let it be destroyed along with the cluster so that other e2e that require nodepool to have nodes can use it without conflict with other nodepool that does not need nodes.

test/e2e/yurt/iot.go
LavenderQAQ: Thank you for reminding me. I don't understand what this means, but the Ready should determine the state of the component (default, optional, custom).
https://github.com/openyurtio/openyurt/blob/068fefe9e7b22d47d2dd5e13d4dbf7953f01e60d/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go#L303-L335
Here the component is synchronized from the framework.
https://github.com/openyurtio/openyurt/blob/068fefe9e7b22d47d2dd5e13d4dbf7953f01e60d/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go#L501C2-L501C2

wangxye: What I mean is that I'm aware of the fact that platformadmin is truly ready in the sense that all components under platformadmin are ready.
https://github.com/openyurtio/openyurt/blob/068fefe9e7b22d47d2dd5e13d4dbf7953f01e60d/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go#L321-L330
we use `ReadyComponentNum` and `UnreadyComponentNum` to indicate the number of ready and unready components respectively. Is it possible to use this as a way of evaluating the real situation of components under platformadmin.
https://github.com/openyurtio/openyurt/blob/068fefe9e7b22d47d2dd5e13d4dbf7953f01e60d/pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go#L376-L396

LavenderQAQ: I got it. Thank you for pointing it out.

wangxye (approved):

pkg/yurtmanager/controller/platformadmin/platformadmin_controller.go
```diff
@@ -393,6 +393,9 @@ func (r *ReconcilePlatformAdmin) reconcileComponent(ctx context.Context, platfor
 	defer func() {
 		platformAdminStatus.ReadyComponentNum = readyComponent
 		platformAdminStatus.UnreadyComponentNum = int32(len(platformAdminFramework.Components)) - readyComponent
+		if platformAdminStatus.UnreadyComponentNum != 0 {
+			platformAdminStatus.Ready = false
```
Rui-Gan, line 397: It would be more appropriate to use "return false, nil" here. Because if it returns true, when it gets back to the reconcileNormal function, it will continue to execute further, setting platformAdminStatus.Ready to true.

test/e2e/yurt/iot.go
```diff
@@ -0,0 +1,186 @@
...
+						return nil
+					} else {
+						return fmt.Errorf("The %s version of PlatformAdmin is not ready", version)
+					}
+				}, platformadminTimeout, 5*time.Second).Should(Succeed())
+			})
+
+			It(fmt.Sprintf("The %s version of PlatformAdmin should deploy the corresponding components after it is created", version), func() {
```
Rui-Gan, line 143: platformAdminStatus.Ready includes both ComponentAvailableCondition and ConfigmapAvailableCondition. Perhaps we can remove the verification test here.

LavenderQAQ:
@Rui-Gan The above issues have been fixed.

Rui-Gan (approved):

rambohe-ch (approved):

PR #1770, opened 2023-11-06
disable the iptables setting of yurthub component by default

rambohe-ch:
/kind enhancement

disable iptables setting of Yurthub component by default, and we will remove iptables setting in the future. this means we will remove parameter `--enable-iptables` of yurthub.

the reason as following:

> We originally expected that these iptables configurations would help clients in the container network to have access to yurthub's dummy interface, but in fact many situations prevented yurthub from running and caused confusion for users. At the same time, we found that without these iptables configurations, clients in the container network can still access the yurthub's dummy interface, so removing these configurations would be a better choice.

Fixes #

```release-note

```


YTGhost (approved):
/lgtm

PR #1775, opened 2023-11-07
disable iptable by default

huiwq1990:
/kind enhancement
> /kind good-first-issue
> /kind feature
> /kind question
> /kind design
> /sig ai
> /sig iot
> /sig network
> /sig storage

Fixes https://github.com/openyurtio/openyurt/pull/1770

```release-note

```


rambohe-ch:
@huiwq1990 Thanks for your pull request, but i don't think that we should configure `--enable-dummy-if` to false by default. Instead, I think it should default to true so that pods in the container network can access kube-apiserver via yurthub.

huiwq1990:
> @huiwq1990 Thanks for your pull request, but i don't think that we should configure `--enable-dummy-if` to false by default. Instead, I think it should default to true so that pods in the container network can access kube-apiserver via yurthub.

PR #1791, opened 2023-11-13
feat:fix error log format in cache manager

### LINKED ISSUE #1776: [feature request]fix error log format in cache manager

As the following code: var `key` is a struct, not a string, so print format is not match.

please change to `key.Key()` to fix it.

https://github.com/openyurtio/openyurt/blob/76a4cfb36a3cdb217d6431b18b06cfd7d8146c34/pkg/yurthub/cachemanager/cache_manager.go#L442-L444

**others**
/kind feature

NewKeyTo:
What type of PR is this?
/kind good-first-issue
/kind feature

fix error log format in cache manager.

Which issue(s) this PR fixes:
Fixes #1776


YTGhost (approved):
/lgtm

rambohe-ch (approved):

PR #1792, opened 2023-11-13
improve error message

### LINKED ISSUE #1781: [feature request] improve error message in order to up efficiency.
When unit tests failed, we need to search the keyword `fail` to find out the failed cases, but there are many error messages that include keyword `fail`(such as in the following image), this will affect our efficiency in finding failed cases.

image

so i think we can improve the error messages, this means change word `fail` to `couldn't` as following:

```
failed to update the DynamicRESTMapper failed to update cached dynamic RESTMapper, path does not exist --> could not update the DynamicRESTMapper failed to update cached dynamic RESTMapper, path does not exist
```

**others**
/kind feature

dzcvxe:
What type of PR is this?
> /kind good-first-issue
> /kind feature

When unit tests failed, we need to search the keyword fail to find out the failed cases, but there are many error messages that include keyword fail, this will affect our efficiency in finding failed cases.

Which issue(s) this PR fixes:
Fixes #1781

### DIFF
--- a/cmd/yurt-iot-dock/app/core.go
+++ b/cmd/yurt-iot-dock/app/core.go
@@ -97,22 +97,21 @@ func Run(opts *options.YurtIoTDockOptions, stopCh <-chan struct{}) {
 	// perform preflight check
 	setupLog.Info("[preflight] Running pre-flight checks")
 	if err := preflightCheck(mgr, opts); err != nil {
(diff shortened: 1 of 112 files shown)


rambohe-ch (approved):

rambohe-ch:
@dzcvxe Thank for very much, it is a good job.

YTGhost (approved):
/lgtm

Issue #1793, opened 2023-11-14
[BUG]fix cache leak in yurtappoverrider controller


rambohe-ch:
**What happened**:
1.  when a yurtappoverrider instance is deleted, the cache data in `CacheOverriderMap` should also be removed. so we can add cache cleanup in the following code snippet.

https://github.com/openyurtio/openyurt/blob/3636ec0f1373125ed6da76f61fc318cd1b621a0e/pkg/yurtmanager/controller/yurtappoverrider/yurtappoverrider_controller.go#L129

2. when entires of yurtappoverrider are changed, we should update the new instance into `CacheOverriderMap`.

https://github.com/openyurtio/openyurt/blob/3636ec0f1373125ed6da76f61fc318cd1b621a0e/pkg/yurtmanager/controller/yurtappoverrider/yurtappoverrider_controller.go#L162-L168

so we can improve the up codes as following:

```
	if cacheOverrider, ok := r.CacheOverriderMap[instance.Namespace+"/"+instance.Name]; ok {
		if reflect.DeepEqual(cacheOverrider.Entries, instance.Entries) {
			return reconcile.Result{}, nil
		}
	}
	r.CacheOverriderMap[instance.Namespace+"/"+instance.Name] = instance.DeepCopy()
```

**Environment**:
- OpenYurt version: v1.4.0
- Kubernetes version (use `kubectl version`):
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

YTGhost:
I think it can be closed.

PR #1795, opened 2023-11-14
fix the cache leak in yurtappoverrider controller

MeenuyD:
Add Cache Cleanup and Update on Deletion and Changes

When a `YurtAppOverrider` instance is deleted, the cache data in `CacheOverriderMap` should also be removed. This ensures that the cache stays synchronized with the actual instances.

Additionally, when entries of `YurtAppOverrider` are changed, the controller now updates the new instance into `CacheOverriderMap`. This guarantees that the cache reflects the latest changes in the `YurtAppOverrider` instances.
#1793


YTGhost (approved):
/lgtm

rambohe-ch (approved):

PR #1816, opened 2023-11-21
[Proposal] High Availability of Edge Services Proposal

Rui-Gan:
For multi-instance service applications within the same edge node pool, design and implement a high-availability solution at the network layer to ensure that client requests are not forwarded to faulty service instances. Moreover, it should be capable of distributing requests to healthy service endpoints according to a predetermined policy.

Fixes #

```release-note

```


LavenderQAQ (changes requested):
- We need to add instruction for modifying kernel parameters. 
- When synchronizing `keepalived`, try not to break user-defined configurations. 
- The use of `keepalived` containers could be reflected in the proposal, we need to use the `osixia/keepalived` image in the Pod.

rambohe-ch:
1. we need to define a new LoadBalancerClass for the service, and then YurtLB and speaker client only handle this kind of service.
2. we do not support user to modify vip directly through keepalived configuration file in the first stage.

rambohe-ch:
by the way, please fix the markdown file errors.

LavenderQAQ (approved):
/lgtm

docs/proposals/20231121-high-availability-of-edge-services.md
```diff
@@ -0,0 +1,282 @@
...
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: nginx
+  namespace: default
+spec:
+  type: LoadBalancer
+  loadBalancerClass: apps.openyurt.io/YurtLB
```
rambohe-ch, line 105: how about modify the class name from `apps.openyurt.io/YurtLB` to `service.openyurt.io/vip`, because vip will be used to implement LoadBlancer on edge node.

docs/proposals/20231121-high-availability-of-edge-services.md
```diff
@@ -0,0 +1,282 @@
...
+
+- YurtLB Controller selects the appropriate VIP for the load-balanced service using customized assignment policies (round robin, random, etc.). It's important that the VIP chosen from the nodepool's VIP set isn't selected repeatedly. In cases of repeated allocation, YurtLB controller will retain only the first successful allocation.
+- YurtLB Controller will deliver the selected VIP information to the nodes where the load balancer service is located, including the nodes where the pods reside.
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: nginx
+  namespace: default
```
rambohe-ch, line 102: I think we need to give users the options of which nodepool LBs are created in, not all nodepools with associated pods. so it is a good idea to use nodepool selector in annotations for selecting the target nodepools as following:

```
Annotations:
    service.openyurt.io/pool-labelselector: key1=val1,key2=val2
```

all nodepools with lables: key1: val1 and key2:val2 will be selected, and LB instance will be created in these nodepools.

rambohe-ch: @River-sh For the annotation definition, do you have any comments?

River-sh: I recommend adopting `service.openyurt.io/nodepool-labelselector: key1=val1,key2=val2`

rambohe-ch (approved):

Issue #1820, opened 2023-11-22
[feature request]improve logs of yurt-manager component


rambohe-ch:
**What would you like to be added**:
if a ValidatingWebhook is disabled,  the following logs will always be output continuously. and this affects the user's ability to parse yurt-manager's logs. 

```
Ignore webhook for xxx in configuration
```

so it is a good idea to modify the log level from warning to info5 as following:

https://github.com/openyurtio/openyurt/blob/299c5f97dcfbcb32ebacf997e1f4afb12432e759/pkg/yurtmanager/webhook/util/configuration/configuration.go#L89

=> klog.V(5).Infof("Ignore webhook for %s in configuration", path)

**others**
/kind feature

PR #1821, opened 2023-11-22
Added logs of yurt-manager component

MeenuyD:
if a ValidatingWebhook is disabled, the following logs will always be output continuously. and this affects the user's ability to parse yurt-manager's logs.

so modify the log level from warning to info5
#1820


rambohe-ch (approved):

rambohe-ch:
Fixes #1820

Issue #1813, opened 2023-11-20
[feature request]print out all registered controllers


rambohe-ch:
**What would you like to be added**:

- background:

yurt-manager component has already included 18 controllers as following:

https://github.com/openyurtio/openyurt/blob/754ccf85f6586337098d2330cffc0b21b0e399fe/pkg/yurtmanager/controller/controller.go#L78-L95

a part of controllers register function(`Add`) have printed out the controller info, such as: 

https://github.com/openyurtio/openyurt/blob/754ccf85f6586337098d2330cffc0b21b0e399fe/pkg/yurtmanager/controller/servicetopology/endpoints/endpoints_controller.go#L55

but some of register function(`Add`) of controllers have not printed out info, such as:

https://github.com/openyurtio/openyurt/blob/754ccf85f6586337098d2330cffc0b21b0e399fe/pkg/yurtmanager/controller/nodelifecycle/node_lifecycle_controller.go#L290-L396

- solution:

it is easy to verify which controllers are registered if register log is printed out, so we need to improve the log print of controller register in the entry point instead of in each controller register function.

so i think we should improve register log as following, and then remove `controller add info` in each `Add` function:

https://github.com/openyurtio/openyurt/blob/754ccf85f6586337098d2330cffc0b21b0e399fe/pkg/yurtmanager/controller/controller.go#L112-L118

```
		if err := fn(ctx, c, m); err != nil {
			if kindMatchErr, ok := err.(*meta.NoKindMatchError); ok {
				klog.Infof("CRD %v is not installed, its controller will perform noops!", kindMatchErr.GroupKind)
				continue
			}
			return err
		} else {
                        klog.Infof("controller %s is added", controllerName)
                }
```

**others**
/kind feature

PR #1824, opened 2023-11-23
print out all registered controllers

MeenuyD:
it is easy to verify which controllers are registered if register log is printed out, so  improve the log print of controller register in the entry point instead of in each controller register function.
#1813


rambohe-ch (approved):

rambohe-ch:
Fixes #1813

PR #1834, opened 2023-11-29
improve FieldIndexer setting in yurt-manager

### LINKED ISSUE #1794: [feature request]improve FieldIndexer setting in yurt-manager
In `NodeLifeCycle` and `PodBinding` controller of yurt-manager component, `spec.nodeName` IndexField has been used at present, and FieldIndexer setting is only needed to configured only once, and now is configured in `NodeLifeCycle` controller as following:

https://github.com/openyurtio/openyurt/blob/3636ec0f1373125ed6da76f61fc318cd1b621a0e/pkg/yurtmanager/controller/nodelifecycle/node_lifecycle_controller.go#L332-L344

This FieldIndexer setting is shared by two controllers, maybe it is a good idea to move this sharing setting as a public setting instead of configuring in `NodeLifeCycle` controller.

Because in some scenarios, users like to disable `NodeLifeCycle` controller, so FieldIndexer setting will not be configured, then  an error happens in `PodBinding` controller as following:

```
E1113 13:40:14.428554       1 podbinding_controller.go:180] pod-binding-controller: failed to get podList for node(izbp13108hhdvq7ver655tz), Index with name field:spec.nodeName does not exist
E1113 13:40:14.428590       1 controller.go:317] setup/controller/pod-binding-controller "msg"="Reconciler error" "error"="Index with name field:spec.nodeName does not exist" "name"="izbp13108hhdvq7ver655tz" "namespace"=""
```

**others**
/kind feature

2456868764:

improve FieldIndexer setting in yurt-manager

Fixes #1794

```release-note

```

### DIFF
--- a/pkg/yurtmanager/controller/controller.go
+++ b/pkg/yurtmanager/controller/controller.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/controller-manager/app"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/openyurtio/openyurt/cmd/yurt-manager/app/config"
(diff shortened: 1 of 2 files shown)


pkg/yurtmanager/controller/controller.go
```diff
@@ -120,5 +122,19 @@ func SetupWithManager(ctx context.Context, c *config.CompletedConfig, m manager.
 		}
 	}
 
+	if err := m.GetFieldIndexer().IndexField(context.TODO(), &v1.Pod{}, "spec.nodeName", func(rawObj client.Object) []string {
```
rambohe-ch, line 125: how about add the following if condition to check register is needed or not.

```
if app.IsControllerEnabled(names.NodeLifeCycleController, ControllersDisabledByDefault, c.ComponentConfig.Generic.Controllers) ||
app.IsControllerEnabled(names.PodBindingController, ControllersDisabledByDefault, c.ComponentConfig.Generic.Controllers) {
    // add index field register here
}
```

rambohe-ch (approved):

PR #1837, opened 2023-11-30
fix: yurtadm join ignorePreflightErrors could not set all

### LINKED ISSUE #1836: [BUG] yurtadm join ignorePreflightErrors could not set all
**What happened**:

When I exec the join commad which set ignore-preflight-errors to all, it will execute fail.

```shell
 yurtadm join ${apiserverAddress}:${apiserverPort} --node-name=${nodeName} --node-labels=${nodeLabels} \
--token=${bootstrapToken} --discovery-token-unsafe-skip-ca-verification --ignore-preflight-errors all --yurthub-image=${yurthubImage}
```

```
[ignorePreflightErrors: Invalid value: "all": 'all' cannot be used in configuration file, ignore-preflight-errors: Invalid value: "all,diravailable--etc-kubernetes-manifests,fileavailable--etc-kubernetes-kubelet.conf": don't specify individual checks if 'all' is used]
```

It should set to ignore-preflight-errors to all.

We should special process when values is all.

https://github.com/openyurtio/openyurt/blob/bc8b77dd968a498c3b4d98b746df314877981425/pkg/yurtadm/constants/constants.go#L183

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

YTGhost:
#### What type of PR is this?
/kind bug

Fixes #1836 

```release-note

```

### DIFF
--- a/pkg/yurtadm/cmd/config/config.go
+++ b/pkg/yurtadm/cmd/config/config.go
@@ -90,7 +90,7 @@ func runConfigPrintActionDefaults(out io.Writer, configBytesProc func() (string,
 
 func getDefaultNodeConfigBytes() (string, error) {
 	KubeadmJoinDiscoveryFilePath := filepath.Join(constants.KubeletWorkdir, constants.KubeadmJoinDiscoveryFileName)
-	ignoreErrors := sets.String{}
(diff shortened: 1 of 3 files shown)


YTGhost:
@rambohe-ch @huiwq1990 PTAL

rambohe-ch:
@huiwq1990 PTAL

PR #1850, opened 2023-12-11
feat: support v3 rest api client for edgex v3 api

wangxye:
#### What type of PR is this?
/sig iot
/kind feature
/kind enhancement

The api design of edgex has changed in v2 and v3 versions, the original v2 client does not support v3 api due to dependency libraries and api request paths, etc. In addition, the community intends to implement the v3 client based on MessageBus in the future. This PR is mainly for the preparation of message bus support.

The specific programs are as follows:
- Supports edgex list/get based on message bus (openyurt side will still use the rest api for create/update/delete). 
- Supports edgex create/update/delete based on message bus (all will be implemented using the message bus).

It will gradually replace the original implementation based on the rest API method.

/assign @LavenderQAQ @Rui-Gan


LavenderQAQ (approved):
/lgtm

PR #1857, opened 2023-12-13
feat: support yurt-manager work in specified namespace

### LINKED ISSUE #1845: [Question] cloud-edge network connection not work when openyurt components are deployed in specified namesapce
**What happened**:
I deploy openyurt components in specified namesapce _openyurt_, and join edge node by `yurtadm join --namespace=openyurt ...`, then create nodepool and gateway, but I cannot access pod running on edge nodes from cloud.
But when I redeploy openyurt components in kube-system namespace, access edge pods successfully.

Everything works well when openyurt components deployed in specified namespace.

By browsing yurt-manager logs, when deploying in specified namespace, yurt-manager don't print log in https://github.com/openyurtio/openyurt/blob/c1a47601312502adab12070b8ad42b48494239af/pkg/yurtmanager/controller/raven/gatewaypickup/gateway_pickup_controller.go#L302
It seems that _gateway_pickup_controller_ directly get raven-config info from kube-system namespace. 

**Environment**:
- OpenYurt version: v1.4.0
- Kernel (e.g. `uname -a`):
- Install tools: helm
- Others:

**others**
/kind question

tagGeeY:
#### What type of PR is this?
/kind feature

1. support gateway controllers work normally in specified namespace;
2. CRDs can be templated rather than hard code, so the namespace of conversion webhooks service can be configured in `gateways.raven.openyurt.io` and `nodepools.apps.openyurt.io`.

Fixes #1845

/assign @rambohe-ch 

```release-note
NONE
```

### DIFF
--- a/pkg/yurtmanager/controller/controller.go
+++ b/pkg/yurtmanager/controller/controller.go
@@ -39,6 +39,7 @@ import (
 	"github.com/openyurtio/openyurt/pkg/yurtmanager/controller/raven/gatewayinternalservice"
 	"github.com/openyurtio/openyurt/pkg/yurtmanager/controller/raven/gatewaypickup"
 	"github.com/openyurtio/openyurt/pkg/yurtmanager/controller/raven/gatewaypublicservice"
+	ravenutil "github.com/openyurtio/openyurt/pkg/yurtmanager/controller/raven/util"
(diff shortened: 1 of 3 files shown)


tagGeeY:
@rambohe-ch OK, I revoked modification in yurt-manager chart.

PR #1875, opened 2023-12-26
Feature: add name-length of dummy interface too long error

### LINKED ISSUE #1867: [feature request]the name length of dummy interface should be under 15 chars.
As we known, yurthub component will create a dummy interface on nodes, and end user can configure the name for this dummy interface by parameter `--dummy-if-name`.

At the same time,  the name length of network interface should be under 15 bytes, some discussion information can be found here: https://stackoverflow.com/questions/24932172/what-length-can-a-network-interface-name-have

so we need to add a `dummy interface name length` check in the following validate function, if the length of `options.HubAgentDummyIfName` is more than 15, an error will be returned.

https://github.com/openyurtio/openyurt/blob/74b0a474deee7a2305d679c3eeb270fb6245d24b/cmd/yurthub/app/options/options.go#L143

**others**
/kind feature

8rxn:
Adds a new validation for dummy interface length not to exceed 15 chars. 

Fixes #1867 

/@rambohe-ch

```release-note
NONE
```

### DIFF
--- a/cmd/yurthub/app/options/options.go
+++ b/cmd/yurthub/app/options/options.go
@@ -167,6 +167,10 @@ func (options *YurtHubOptions) Validate() error {
 		return fmt.Errorf("dummy ip %s is not invalid, %w", options.HubAgentDummyIfIP, err)
 	}
 
+	if len(options.HubAgentDummyIfName) > 15 {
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@8rxn Thank you for your good job.

rambohe-ch (approved):

PR #1888, opened 2023-12-29
feat: improve yurtstaticset upgrade model check (#1870)

### LINKED ISSUE #1870: [feature request]improve yurtstaticset upgrade models check
1. we have added two new upgrade models for static pod, AdvancedRollingUpdate and OTA. and the definition as following:

https://github.com/openyurtio/openyurt/blob/74b0a474deee7a2305d679c3eeb270fb6245d24b/pkg/apis/apps/v1alpha1/yurtstaticset_types.go#L38-L41

2. and only `AdvancedRollingUpdate, OTA` can be accepted when we use it as the following:

https://github.com/openyurtio/openyurt/blob/74b0a474deee7a2305d679c3eeb270fb6245d24b/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go#L113

3. we want to support more flexible for these upgrade models, for example: `advancedRollingUpdate` with small letter begin, or  `ota` with all small letters. so maybe we can use `strings.ToLower()` to change these upgrade models to entire small letters before models check.

4. we need to improve all places for yurtstaticset which these new upgrade models are used.

**others**
/kind feature

huangchenzhao:
Check two new upgrade models AdvancedRollingUpdate and OTA using case-insensitive string judgment instead of using "==".

By introducing `strings.EqualFold(stringA, stringB)`, we can ignore the difference between upper and lower case before models check.
 
Fixes #1870

### DIFF
--- a/pkg/apis/apps/v1alpha1/default.go
+++ b/pkg/apis/apps/v1alpha1/default.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strings"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	v1 "k8s.io/kubernetes/pkg/apis/core/v1"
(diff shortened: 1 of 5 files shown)


rambohe-ch:
@huangchenzhao Thanks for making the pull request. beside `https://github.com/openyurtio/openyurt/blob/master/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go` file, the following files also should be updated.

https://github.com/openyurtio/openyurt/blob/master/pkg/apis/apps/v1alpha1/default.go#L215

https://github.com/openyurtio/openyurt/blob/master/pkg/node-servant/static-pod-upgrade/upgrade.go#L83

https://github.com/openyurtio/openyurt/blob/master/pkg/yurtmanager/controller/yurtstaticset/util/util.go#L66

https://github.com/openyurtio/openyurt/blob/master/pkg/yurtmanager/controller/yurtstaticset/yurtstaticset_controller.go#L375

huangchenzhao:
Thanks for reminding, I will fix it.

rambohe-ch (approved):

PR #1896, opened 2024-01-07
fix: yurtadm join can't work when kubernetes version large than v1.27.0

### LINKED ISSUE #1675: [BUG] yurtadm does not work with kubernetes 1.27 due to flag removal
**What happened**:
When `yurtadm join` specifies `--cri-socket` that is not `/var/run/dockershim.sock`, the `--container-runtime` flag is added to `kubeletExtraArgs`.

The kubelet can't start, because the `--container-runtime=remote` flag has been dropped:
```
"command failed" err="failed to parse kubelet flag: unknown flag: --container-runtime"
```
It is only needed for versions before 1.24, and has been deprecated in 1.24 through 1.26

https://kubernetes.io/blog/2023/03/17/upcoming-changes-in-kubernetes-v1-27/

Compatible with this change.

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

2456868764:
#### What type of PR is this?

 /kind bug

fix: yurtadm join can't work when kubernetes version is large than v1.27.0

Fixes #1675

```release-note

```

### DIFF
--- a/pkg/yurtadm/util/kubernetes/kubernetes.go
+++ b/pkg/yurtadm/util/kubernetes/kubernetes.go
@@ -405,11 +405,11 @@ func SetKubeadmJoinConfig(data joindata.YurtJoinData) error {
 	if nodeReg.CRISocket == constants.DefaultDockerCRISocket {
 		ctx["networkPlugin"] = "cni"
 	} else {
-		v127, err := version.NewVersion("v1.27.0")
+		v124alpha, err := version.NewVersion("1.24.0-alpha.0")
 		if err != nil {
 			return err
 		}
-		if v1.LessThan(v127) {
+		if v1.LessThan(v124alpha) {
 			ctx["containerRuntime"] = "remote"
 		}
 		ctx["containerRuntimeEndpoint"] = nodeReg.CRISocket
--- a/pkg/yurtadm/util/kubernetes/kubernetes.go
+++ b/pkg/yurtadm/util/kubernetes/kubernetes.go
@@ -396,17 +396,25 @@ func SetKubeadmJoinConfig(data joindata.YurtJoinData) error {
 		"criSocket":              nodeReg.CRISocket,
 		"name":                   nodeReg.Name,
 	}
-	if nodeReg.CRISocket == constants.DefaultDockerCRISocket {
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@YTGhost PTAL

pkg/yurtadm/util/kubernetes/kubernetes.go
```diff
@@ -396,17 +396,25 @@ func SetKubeadmJoinConfig(data joindata.YurtJoinData) error {
...
 		ctx["networkPlugin"] = "cni"
 	} else {
-		ctx["containerRuntime"] = "remote"
+		v127, err := version.NewVersion("v1.27.0")
+		if err != nil {
+			return err
+		}
+		if v1.LessThan(v127) {
+			ctx["containerRuntime"] = "remote"
+		}
```
YTGhost, line 414: Since this flag only needed for versions before 1.24, and has been deprecated in 1.24 through 1.26. So I think we only need this flag if the k8s version is earlier than `1.24.0-alpha.0`, what do you think?

2456868764: This flag has been deprecated in 1.24 through 1.26, but it is removed in 1.27. So it is ok to compare with version 1.27.0.

2456868764: OK, I got it.

2456868764: i fixed to compare with 1.24.0-alpha.0

YTGhost (approved):
/lgtm
/approve

siredmar:
@rambohe-ch does this PR mean, that versions >1.27 are supported by openyurt?

rambohe-ch:
> @rambohe-ch does this PR mean, that versions >1.27 are supported by openyurt?

@siredmar yes, some users are using OpenYurt v1.4 with K8s v1.27. and this PR is used for improving `yurtadm join` tool.

siredmar:
Are there any use cases that do not work properly >1.22.17? Is it possible to make an official statement that OpenYurt works using 1.27?

PR #1902, opened 2024-01-09
fix: yad status should be updated according to the new created ControllerRevision (#1879)

### LINKED ISSUE #1879: [BUG] YurtAppDaemon status currentRevision is not updated when yad change
**What happened**:
After editing yurt-app-daemon(yad)'s spec, another controllerRevision(cr) will be created and currentRevision in yad's status should be updated according to the new created cr. Currently, the status will not be updated.

yad's status currentRevision is up to date with the latest cr.

**Environment**:
- OpenYurt version: v1.4 
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

huangchenzhao:
After editing yad's spec, there is a new created ControllerRevision. 

However, yad's status isn't updated because the function updateStatus use param `currentRevision`, instead of the updated value. We can see that in https://github.com/openyurtio/openyurt/blob/master/pkg/yurtmanager/controller/yurtappdaemon/yurtappdaemon_controller.go#L203

So, when invoking function updateStatus, we just need to replace `currentRevision` to `expectedRevision` to solve this problem.

Fixes #1879

### DIFF
--- a/pkg/yurtmanager/controller/yurtappdaemon/yurtappdaemon_controller.go
+++ b/pkg/yurtmanager/controller/yurtappdaemon/yurtappdaemon_controller.go
@@ -200,7 +200,7 @@ func (r *ReconcileYurtAppDaemon) Reconcile(_ context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
-	return r.updateStatus(instance, newStatus, oldStatus, currentRevision, collisionCount, templateType, currentNPToWorkload)
+	return r.updateStatus(instance, newStatus, oldStatus, expectedRevision, collisionCount, templateType, currentNPToWorkload)
 }
 
 func (r *ReconcileYurtAppDaemon) updateStatus(instance *unitv1alpha1.YurtAppDaemon, newStatus, oldStatus *unitv1alpha1.YurtAppDaemonStatus,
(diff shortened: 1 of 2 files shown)


rambohe-ch:
@YTGhost PTAL

pkg/yurtmanager/controller/yurtappdaemon/yurtappdaemon_controller.go
```diff
@@ -200,7 +200,7 @@ func (r *ReconcileYurtAppDaemon) Reconcile(_ context.Context, request reconcile.
 		return reconcile.Result{}, err
 	}
 
-	return r.updateStatus(instance, newStatus, oldStatus, currentRevision, collisionCount, templateType, currentNPToWorkload)
+	return r.updateStatus(instance, newStatus, oldStatus, expectedRevision, collisionCount, templateType, currentNPToWorkload)
```
YTGhost, line 203: I think your changes are correct, and I found the same problem in `yurtappset`, can you help to change it together? cc @luc99hen 
https://github.com/openyurtio/openyurt/blob/d7f64fc295d7a7378d800c37ae83caf5d0570c67/pkg/yurtmanager/controller/yurtappset/yurtappset_controller.go#L209

huangchenzhao: Yes, thanks for reminding!
I will fix another problem.

YTGhost:
> /LGTM

Let's wait for @huangchenzhao to fix the same problem at yurtappset.

huangchenzhao:
> > /LGTM
> 
> Let's wait for @huangchenzhao to fix the same problem at yurtappset.

I have updated the commit to solve the same problem at yurtappset, thanks for nicely reviewing :）

YTGhost:
@huangchenzhao Thank you for your contribution!

PR #1926, opened 2024-01-17
feat: improve discardcloudservice filter in yurthub component (#1924)

### LINKED ISSUE #1924: [feature request]improve discardcloudservice filter in yurthub component
Because clients on edges nodes can not connect cloud applications through pod ip, so LoadBalancer service for cloud application is discarded in yurthub component in order to make kube-proxy unaware of LoadBalancer service on edge nodes. and only LoadBalancer service with annotation `openyurt.io/skip-discard=true` will be skipped. 
the following are detail codes:

https://github.com/openyurtio/openyurt/blob/81d21b7f438be4fb912ebe9748f0c9a4a9296f5f/pkg/yurthub/filter/discardcloudservice/filter.go#L81-L86

but in the proposal: https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20231121-high-availability-of-edge-services.md, LoadBalancer service across NodePools will be supported, this means that LoadBalancer service should be aware of on edge nodes.

so we will improve discardcloudservice filter in yurthub component, and only discard LoadBalancer service which has annotation `svc.openyurt.io/discard=true`, this means that LoadBalancer service will not be discarded by default.

**others**
/kind feature

huangchenzhao:
have improved discardcloudservice filter in yurthub component

only discarding LoadBalancer service which has annotation `svc.openyurt.io/discard=true`, instead of which doesn't have `openyurt.io/skip-discard=true`.

Fixes #1924

### DIFF
--- a/pkg/yurthub/filter/constant.go
+++ b/pkg/yurthub/filter/constant.go
@@ -46,6 +46,11 @@ const (
 	// If end users want to use specified LB service at the edge side,
 	// End users should add annotation["openyurt.io/skip-discard"]="true" for LB service.
 	SkipDiscardServiceAnnotation = "openyurt.io/skip-discard"
+
+	// DiscardServiceAnnotation is annotation used by LB service.
+	// If end users want to discard specified LB service at the edge side,
+	// End users should add annotation["svc.openyurt.io/discard"]="true" for LB service.
+	DiscardServiceAnnotation = "svc.openyurt.io/discard"
 )
 
 var (
--- a/pkg/yurthub/filter/discardcloudservice/filter.go
+++ b/pkg/yurthub/filter/discardcloudservice/filter.go
@@ -79,7 +79,7 @@ func discardCloudService(svc *v1.Service) *v1.Service {
 	nsName := fmt.Sprintf("%s/%s", svc.Namespace, svc.Name)
 	// remove cloud LoadBalancer service
 	if svc.Spec.Type == v1.ServiceTypeLoadBalancer {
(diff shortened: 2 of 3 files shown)


huangchenzhao:
It seems unit tests in 'discardcloudservice filter' should be modified to ensure synchronization with the changed code, I will fix it soon.

pkg/yurthub/filter/constant.go
```diff
@@ -46,6 +46,11 @@ const (
 	// If end users want to use specified LB service at the edge side,
 	// End users should add annotation["openyurt.io/skip-discard"]="true" for LB service.
 	SkipDiscardServiceAnnotation = "openyurt.io/skip-discard"
+
+	// DiscardServiceAnnotation is annotation used by LB service.
+	// If end users want to discard specified LB service at the edge side,
+	// End users should add annotation["openyurt.io/discard"]="true" for LB service.
+	DiscardServiceAnnotation = "openyurt.io/discard"
```
rambohe-ch, line 53: openyurt.io/discard --> svc.openyurt.io/discard

rambohe-ch:
> It seems unit tests in 'discardcloudservice filter' should be modified to ensure synchronization with the changed code, I will fix it soon.

yes, unit tests also should be tuned.

pkg/yurthub/filter/constant.go
huangchenzhao: That's right, fixed.

rambohe-ch (approved):

PR #1949, opened 2024-02-20
don't cache csr and sar resource in yurthub

### LINKED ISSUE #1880: [feature request]yurthub doesn't need to cache certificatesigningrequests and subjectaccessreviews resource
1. Certain clients, such as kubelet, submit certificatesigningrequests (CSRs) and subjectaccessreviews (SARs) requests to the kube-apiserver via yurthub.

2. Yurthub is not required to cache the responses to these requests, as clients will not utilize cached CSRs or SARs data upon restart.

so we can add `check whether resource is CSRs or SARs` in the following function:

https://github.com/openyurtio/openyurt/blob/9fb701eb8b5d50c97674f441fede8273149fec73/pkg/yurthub/cachemanager/cache_manager.go#L709-L790

please add the following code in the above function:

```
var nonCacheableResources = map[string]struct{}{
    "certificatesigningrequests": {},
    "subjectaccessreviews": {},
}

if _, ok := nonCacheableResources[info.Resource]; ok {
    return false
}
```

at the same time, please add unit test cases into `TestCanCacheFor` func in `cache_manager_test.go` file.

**others**
/kind feature

rambohe-ch:
/kind feature

don't cache csr and sar resource on edge nodes

Fixes #1880

```release-note

```

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -52,6 +52,11 @@ import (
 
 var (
 	ErrInMemoryCacheMiss = errors.New("in-memory cache miss")
+
+	nonCacheableResources = map[string]struct{}{
+		"certificatesigningrequests": {},
+		"subjectaccessreviews":       {},
+	}
 )
 
 // CacheManager is an adaptor to cache runtime object data into backend storage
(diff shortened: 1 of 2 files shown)


YTGhost (approved):
/lgtm

PR #1950, opened 2024-02-20
fix cache manager panic in yurthub

rambohe-ch:
/kind bug

##### background:

yurt-hub-cfg configmap is used in multiple places, like:
- https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_agent.go#L72
- https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/filter/approver.go#L126

In the [cache manager](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/cache_agent.go#L72), configmap is read, but in the [approver](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/filter/approver.go#L126), configmap is modified, so a panic maybe triggered.

##### solution:

deepcopy configmap before modify to avoid panic.

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/approver.go
+++ b/pkg/yurthub/filter/approver.go
@@ -128,11 +128,13 @@ func (a *approver) updateConfigMap(oldObj, newObj interface{}) {
 	if !ok {
 		return
 	}
+	oldCM = oldCM.DeepCopy()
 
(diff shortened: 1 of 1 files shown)


YTGhost (approved):
/lgtm

PR #1959, opened 2024-02-27
feat: improve hostNetwork mode of NodePool by adding NodeAffinity to pods with specified annotation (#1935)

### LINKED ISSUE #1935: [feature request]improve hostNetwork mode of NodePool

- background:

A filter named [hostnetworkpropagation](https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/filter/hostnetworkpropagation/filter.go) in YurtHub component is used for mutating `pod.spec.HostNetwork` field to true in order to adapt hostNetwork mode NodePool. All pods on nodes in the hostNetwork mode NodePool should share the Host network namespace because cni plugin and component(like flannel) have not been installed in the hostNetwork mode NodePool.

But there is a scenario that pods with hostNetwork=false which use the same ports maybe scheduled to the same node in hostNetwork mode NodePool,  then the filter on the node mutate pod hostNetwork field from false to true, so ports conflict will happen and cause pods start failure. this problem is a out of expectation.
 
- solution:

1. It is not a good idea to mutate pod hostNetwork field to true directly, because this maybe cause pods failure. so i think that it is a good idea to add `NodeAffinity` to pods in order to avoid pods to be scheduled on the nodes in the hostNetwork mode NodePool.

2. Pods which don't want to be scheduled to nodes in hostNetwork mode NodePool should be specified by annotation explicitly, and the pod is `annotation["apps.openyurt.io/exclude-host-network-pool"] = true`

3. A webhook will be added in yurt-manager component for adding NodeAffinity to pods with `annotation["apps.openyurt.io/exclude-host-network-pool"] = true`.  Because all nodes in hostNetwork mode Nodepool have `labels[nodepool.openyurt.io/hostnetwork] = true`, so the NodeAffinity will be as following:

```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: nodepool.openyurt.io/hostnetwork
            operator: NotIn
            values:
            - "true"
```

**others**
/kind feature

huangchenzhao:
#### What type of PR is this?

> /kind feature

A mutating webhook is introduced in yurt-manager component for adding NodeAffinity to pods with `annotation["apps.openyurt.io/exclude-host-network-pool"] = true`. The NodeAffinity indicates that the pods don't want to be scheduled to nodes in hostNetwork mode NodePool.

Users should add `annotation ["apps.openyurt.io/exclude-host-network-pool"] = true` manually to pods if they don't want to run pods on edge nodes in hostNetwork mode nodepool.

Fixes #1935


charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
```diff
@@ -632,6 +632,27 @@ webhooks:
...
+  failurePolicy: Ignore
+  name: mutate.core.v1.pod.openyurt.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
```
rambohe-ch, line 652: Only intercept `CREATE` pod is enough, because of pods are managed by workloads like Deployment, so pods will be recreated if annotation is changed. please remove `UPDATE` operation.

pkg/yurtmanager/webhook/pod/v1alpha1/pod_default.go
```diff
@@ -0,0 +1,60 @@
...
+									Operator: corev1.NodeSelectorOpNotIn,
+									Values:   []string{"true"},
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+	}
```
rambohe-ch, line 58: we should not overwrite the existing `pod.Spec.Affinity`, this means we need to append affinity only.

rambohe-ch:
@huangchenzhao filter [hostnetworkpropagation](https://github.com/openyurtio/openyurt/tree/master/pkg/yurthub/filter/hostnetworkpropagation) also should be removed.

charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
```diff
@@ -632,6 +632,27 @@ webhooks:
...
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  - v1beta1
+  clientConfig:
+    service:
+      name: yurt-manager-webhook-service
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-core-openyurt-io-v1-pod
+  failurePolicy: Ignore
```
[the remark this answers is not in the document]
huangchenzhao: yes, already set failurePolicy to Ignore.

charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
huangchenzhao: reasonable, `UPDATE` is redundant, fixed.

pkg/yurtmanager/webhook/pod/v1alpha1/pod_default.go
huangchenzhao: that's right, fixed.

huangchenzhao:
> @huangchenzhao filter [hostnetworkpropagation](https://github.com/openyurtio/openyurt/tree/master/pkg/yurthub/filter/hostnetworkpropagation) also should be removed.

removed

rambohe-ch (approved):

PR #1964, opened 2024-03-04
fix: ensure local proxy can handle requests with the same path but with different selector

### LINKED ISSUE #1963: [BUG] the controller on edge node can not list same path but with different selector when edge node is offline
**What happened**:
the CanCacheFor() method filters out requests with the same path but different selector, so the controller on edge node can not list same path but with different selector when edge node is offline.

YurtHub is capable of correctly proxying requests that have the same path but different selectors when edge node is offline.

**Environment**:
- OpenYurt version:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

windnod:
#### What type of PR is this?
/kind bug

skip the check for the same path list requests, when the localproxy use the canCacheFor() method

Fixes #1963

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -50,6 +50,11 @@ import (
 	"github.com/openyurtio/openyurt/pkg/yurthub/util"
 )
 
+const (
+	CanCacheForRemote = "remote"
+	CanCacheForLocal  = "local"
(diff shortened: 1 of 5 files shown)


rambohe-ch:
@windnod Thanks for posting pull request. I have commented the issue #1963 , and we need to have more discussion about this case before we merge this pull request. sorry to response you lately.

PR #1974, opened 2024-03-11
fix: deployment render should only mutate workloads created by yurtap…

vie-serendipity:
…pset v1alpha1

/kind bug
Yurtappset is updated to v1beta1. Now deployment render will mutate all workloads created by yurtappset. Actually, yurtappoverrider should only manage workloads created by yurtappset v1alpha1. Otherwise, it will cause unexpected errors. So we should add some restrictions to deployment webhook of yurtappoverrider.

Fixes #

```release-note
None
```


rambohe-ch:
@vie-serendipity Thank you for posting the pull request to improve YurtAppSet workload. I agree with you that deployment render webhook should effect v1alpha1.YurtAppSet only.

rambohe-ch (approved):

PR #1979, opened 2024-03-18
fix panic: filter chain should be exited when object is nil

rambohe-ch:
/kind bug

when kube-proxy component list/watch LB service, filter `discardcloudservice` and `nodeportisolation` will be selected to create filter chain. if `discardcloudservice` filter returns a nil object, it will cause panic in `nodeportisolation` filter.

so we need to exit filter chain if object is nil.

Fixes #

```release-note

```


Congrool:
done in #1984

PR #1991, opened 2024-03-25
move list object handling from ObjectFilter into ResponseFilter

### LINKED ISSUE #1990: [feature request]extract list object handling from ObjectFilter to ResponseFilter
In each ObjectFilter(like masterservice, servicetopology, etc.), filter need to handle standalone object and List object. For example, `masterservice` filter need to handle `*v1.ServiceList` and `*v1.Service` as following:

https://github.com/openyurtio/openyurt/blob/acec197f3bee3348cec7d5d23d162ae3df4b3c21/pkg/yurthub/filter/masterservice/filter.go#L82-L95

And handling `list object` is just a collection of standalone object handling, so it is a good idea to handle `list object` in a common place like `ResponseFilter` instead of handling them in every `ObjectFilter`.

So i will make a pull request to improve `list object` handling in `ObjectFilter`, then `ObjectFilter` only need to handle standalone object(like *v1.Service).

**others**
/kind feature

rambohe-ch:
/kind feature

Move list object handling in every ObjectFilter(like masterservice, servicetopology, etc.) into `ResponseFilter`, so each ObjectFilter only need to handle single object.

Fixes #1990 

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/base/base_test.go
+++ b/pkg/yurthub/filter/base/base_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package base
 
 import (
+	"errors"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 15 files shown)


YTGhost (approved):
/lgtm
/approve

PR #1995, opened 2024-03-26
return back watch.Deleted event to clients when watch object is removed in OjbectFilters

### LINKED ISSUE #1994: [feature request] improve nodeportisolation filter，support listening reduction
Previously, the service was listened on   'nodeport.openyurt.io/listen=pool1,pool2'.   After the update, it is now only listening on  'nodeport.openyurt.io/listen=pool1',  but the nodeport of pool2 has not been released. 

**Why is this needed**:
To resolve the above issue, kube-proxy needs to be restarted.  This is not allowed in a production environment.

**others**
/kind feature

rambohe-ch:
/kind feature

ResponseFilter return `watch.Deleted` event back to clients when Standalone object is filtered to nil in ObjectFilters.

by the way, this feature will be helpful for all ObjectFilters, not only for `nodeportisolation` filter.

Fixes #1994 

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/nodeportisolation/filter.go
+++ b/pkg/yurthub/filter/nodeportisolation/filter.go
@@ -93,20 +93,19 @@ func (nif *nodePortIsolationFilter) Filter(obj runtime.Object, stopCh <-chan str
 }
 
 func (nif *nodePortIsolationFilter) isolateNodePortService(svc *v1.Service) *v1.Service {
-	nodePoolName := nif.resolveNodePoolName()
-	// node is not located in NodePool, keep the NodePort service the same as native K8s
-	if len(nodePoolName) == 0 {
-		return svc
-	}
-
-	nsName := fmt.Sprintf("%s/%s", svc.Namespace, svc.Name)
 	if svc.Spec.Type == v1.ServiceTypeNodePort || svc.Spec.Type == v1.ServiceTypeLoadBalancer {
 		if _, ok := svc.Annotations[ServiceAnnotationNodePortListen]; ok {
+			nodePoolName := nif.resolveNodePoolName()
+			// node is not located in NodePool, keep the NodePort service the same as native K8s
+			if len(nodePoolName) == 0 {
+				return svc
+			}
+
 			nodePoolConf := getNodePoolConfiguration(svc.Annotations[ServiceAnnotationNodePortListen])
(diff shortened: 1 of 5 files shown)


pkg/yurthub/filter/responsefilter/filter.go
```diff
@@ -179,11 +179,15 @@ func (frc *filterReadCloser) streamResponseFilter(rc io.ReadCloser, ch chan *byt
 			return err
 		}
 
-		newObj := obj
+		newObj := obj.DeepCopyObject()
```
Congrool, line 182: Is it necessary to `DeepCopyObject`?

pkg/yurthub/filter/nodeportisolation/filter.go
```diff
@@ -93,20 +93,19 @@ func (nif *nodePortIsolationFilter) Filter(obj runtime.Object, stopCh <-chan str
...
-
-	nsName := fmt.Sprintf("%s/%s", svc.Namespace, svc.Name)
 	if svc.Spec.Type == v1.ServiceTypeNodePort || svc.Spec.Type == v1.ServiceTypeLoadBalancer {
 		if _, ok := svc.Annotations[ServiceAnnotationNodePortListen]; ok {
+			nodePoolName := nif.resolveNodePoolName()
+			// node is not located in NodePool, keep the NodePort service the same as native K8s
+			if len(nodePoolName) == 0 {
+				return svc
+			}
+
```
Congrool, line 103: Why to move the block here? Any differences between them?

Congrool:
lgtm, but I still have questions for some of the modifications. @rambohe-ch PTAL

pkg/yurthub/filter/nodeportisolation/filter.go
rambohe-ch: because this filter is used for NodePort service or LB service, and we don't need to do anything for ClusterIP service such as resolve nodepool name.

pkg/yurthub/filter/responsefilter/filter.go
rambohe-ch: yes, there is really no need to DeepCopyObject. i will fix it.

rambohe-ch:
@Congrool I have fixed you comments, PTAL

Congrool (approved):

PR #1975, opened 2024-03-12
proposal: improve transparent management mechanism for control traffic from edge to cloud

rambohe-ch:
/kind enhancement
/kind documentation

#### What this PR does / why we need it:
make a proposal to improve transparent management mechanism for control traffic from edge to cloud.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #1996, opened 2024-03-26
add forwardkubesvctraffic filter and adapt server certificate

rambohe-ch:
/kind feature

add a new ObjectFilter named `forwardkubesvctraffic` in order to improve transparent management mechanism for control traffic from edge to cloud.

the detail design of this filter, the proposal link is here: https://github.com/openyurtio/openyurt/pull/1975

Fixes #

```release-note

```

### DIFF
--- a/pkg/yurthub/filter/forwardkubesvctraffic/filter.go
+++ b/pkg/yurthub/filter/forwardkubesvctraffic/filter.go
@@ -1,0 +1,121 @@
+/*
+Copyright 2024 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package forwardkubesvctraffic
+
+import (
+	"strconv"
+
+	discovery "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+	utilnet "k8s.io/utils/net"
+
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter"
+	"github.com/openyurtio/openyurt/pkg/yurthub/filter/base"
+)
+
+const (
+	// FilterName filter is used to mutate the default/kubernetes endpointslices
+	// in order to make pods on edge nodes can access kube-apiserver directly by default/kubernetes service.
+	FilterName = "forwardkubesvctraffic"
+
+	KubeSVCNamespace = "default"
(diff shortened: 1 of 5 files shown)


pkg/yurthub/filter/forwardkubesvctraffic/filter.go
```diff
@@ -0,0 +1,121 @@
...
+	port        int32
+}
+
+func (fkst *forwardKubeSVCTrafficFilter) Name() string {
+	return FilterName
+}
+
+func (fkst *forwardKubeSVCTrafficFilter) SupportedResourceAndVerbs() map[string]sets.String {
+	return map[string]sets.String{
+		"endpointslices": sets.NewString("list", "watch"),
```
Congrool, line 65: Does it need to support `endpoints`?

rambohe-ch: I think there is no need to support endpoints resource, because EndpointSlice is stable from K8s v1.21, and most  users are currently using K8s versions greater than v1.22.

Congrool: But there're still other filters that support endpoints, like serviceTopologyFilter. It's better to make them all support `endpoints` or not.

BTW, I think we can support the endpoints, since currently it does not take us much effort.

rambohe-ch: It is different with serviceTopologyFilter, because some other components like `nginx-ingress-controller` that are using endpoints resource also need servicetopology filter, but forwardkubesvctraffic filter is only by kube-proxy, so there is no need to deal with endpoints.

pkg/yurthub/filter/forwardkubesvctraffic/filter.go
```diff
@@ -0,0 +1,121 @@
...
+			if eps.Ports[j].Name != nil && *eps.Ports[j].Name == KubeSVCPortName {
+				eps.Ports[j].Port = &fkst.port
+				break
+			}
+		}
+		eps.Endpoints = []discovery.Endpoint{
+			{
+				Addresses: []string{fkst.host},
+				Conditions: discovery.EndpointConditions{
+					Ready: &trueCondition,
```
Congrool, line 114: It's better to replace it with following:
```go
    Ready: pointer.Bool(true)
```
which in pkg `"k8s.io/utils/pointer"`

rambohe-ch: fixed

Congrool:
e2e test error log:
```plain
+ kubectl wait --for=condition=Ready pod/yurt-e2e-test-nginx-openyurt-e2e-test-worker --timeout=240s
Error from server (NotFound): pods "yurt-e2e-test-nginx-openyurt-e2e-test-worker" not found
make: *** [Makefile:114: e2e-tests] Error 1
Error: Process completed with exit code 2.
```

pkg/yurthub/filter/forwardkubesvctraffic/filter.go
Congrool: seems not updated? @rambohe-ch

rambohe-ch: the definition of `trueCondition` is improved by `pointer.Bool(true)`.

Congrool (approved):

PR #2039, opened 2024-05-06
intercept kubelet get node request in order to reduce the traffic

vie-serendipity:
/kind feature

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -52,6 +52,7 @@ import (
 
 var (
 	ErrInMemoryCacheMiss = errors.New("in-memory cache miss")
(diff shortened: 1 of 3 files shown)


vie-serendipity:
@rambohe-ch I add a check whether it's the first request after reconnection.

pkg/yurthub/cachemanager/cache_manager.go
```diff
@@ -433,9 +428,13 @@ func (cm *cacheManager) saveWatchObject(ctx context.Context, info *apirequest.Re
 				} else {
 					updateObjCnt++
 				}
+				errMsg := cm.updateInMemoryCache(ctx, info, obj)
+				klog.ErrorS(errMsg, "failed to update cache")
```
rambohe-ch, line 432: only print error message only when error happened.

PR #2046, opened 2024-05-14
fix: close dst file

testwill:

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/util/file/file.go
+++ b/pkg/util/file/file.go
@@ -90,6 +90,7 @@ func backupFile(path string) error {
 	if err != nil {
 		return err
 	}
+	defer src.Close()
 
 	fileName := filepath.Base(path)
 	bakFile := filepath.Join("/tmp", fileName)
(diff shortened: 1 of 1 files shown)


pkg/util/file/file.go
```diff
@@ -97,6 +97,7 @@ func backupFile(path string) error {
 	if err != nil {
 		return err
 	}
+	defer dst.Close()
 
 	_, err = io.Copy(dst, src)
 	return err
```
rambohe-ch, line 103: how about optimize backupFile function as following:

```
func backupFile(path string) error {
	src, err := os.Open(path)
	if err != nil {
		return err
	}
        defer src.Close()

	fileName := filepath.Base(path)
	bakFile := filepath.Join("/tmp", fileName)
	dst, err := os.Create(bakFile)
	if err != nil {
		return err
	}
	defer dst.Close()

	if _, err = io.Copy(dst, src); err != nil {
	    return err
        }

        if err = dst.Sync(); err != nil {
             return err
        }
        return nil
}

```

rambohe-ch (approved):

PR #2054, opened 2024-05-26
improve restmapper for gateway resource

rambohe-ch:
/kind enhancement

because plural of gateway is gateways instead of gatewaies in raven, so we need to deal with gateway resource specially in restmapper of yurthub component.

Fixes #

```release-note

```


YTGhost (approved):
/lgtm
/approve

PR #2047, opened 2024-05-15
Apply OpenYurt to support K8s v1.28

rambohe-ch:
/kind feature

#### What this PR does / why we need it:
1. upgrade k8s.io/xxx to v0.28.9
2. upgrade controller-runtime to v0.16.5
3. upgrade go version from v1.18 to v1.20
4. upgrade kind to v0.22.0 to local up K8s v1.28 cluster.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2056, opened 2024-05-26
bump controller-gen to v0.13.0

Congrool:

Since we align the openyurt with kubernetes 1.28 in #2047, we should also bump the `controller-gen` to a newer version to avoid compatibility problem whose current version is v0.7.0 which is a bit stale.

`controller-gen` help us automatically generate code for CRDs. We've encountered the situation that the automatic generation cannot be done in release action.

Fixes #

```release-note

```


Congrool:
@rambohe-ch PTAL

Not sure if this can make the github release workflow run successfully, at least it works around at my host.

rambohe-ch (approved):

PR #2051, opened 2024-05-20
Proposal: separate yurt-manager clients

luc99hen:
#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it:
Propose a plan for separating yurt-manager clients.
#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2069, opened 2024-06-03
enhance: separate clients for yurt-manager 

luc99hen:
#### What type of PR is this?
 /kind enhancement

Fix https://github.com/openyurtio/openyurt/pull/2051

Fixes #

```release-note

```


luc99hen:
@rambohe-ch I have found the root cause of the blocking e2e tests are there are no permission for controller-manager's cache building process. As we know, `Cache` is shared among all controllers/clients in the `Manager` level.  And the process of building `Cache` need the full permissions from all clients. 

Therefore, in our case the base client used by yurt-manager need the full permission at the initialization stage.

luc99hen:
We can create a cache for each client used by individual controllers. However, I opt against this approach due to the escalating memory footprint.

pkg/yurtmanager/webhook/deploymentrender/v1alpha1/deploymentrender_handler.go
```diff
@@ -24,13 +24,15 @@ import (
...
+	yurtClient "github.com/openyurtio/openyurt/cmd/yurt-manager/app/client"
+	"github.com/openyurtio/openyurt/cmd/yurt-manager/names"
 	"github.com/openyurtio/openyurt/pkg/yurtmanager/webhook/util"
 )
 
 // SetupWebhookWithManager sets up Cluster webhooks. 	mutate path, validatepath, error
 func (webhook *DeploymentRenderHandler) SetupWebhookWithManager(mgr ctrl.Manager) (string, string, error) {
 	// init
-	webhook.Client = mgr.GetClient()
+	webhook.Client = yurtClient.GetClientByControllerNameOrDie(mgr, names.YurtAppOverriderController)
```
rambohe-ch, line 35: YurtAppOverrider is deprecated now, so maybe it is a good idea to use `YurtAppSet` client here.

luc99hen: Deploymentrender webhook comes with yurtappoverrider. Maybe we can also deprecate this webhook rather than use YurtAppSet client?

cmd/yurt-manager/app/client/client.go
```diff
@@ -0,0 +1,167 @@
...
+	configStore.lock.Lock()
+	defer configStore.lock.Unlock()
+
+	if cfg, ok := configStore.configByName[controllerName]; ok {
+		return cfg
+	}
+
+	// get base config
+	baseCfg := mgr.GetConfig()
+	baseClient, err := kubernetes.NewForConfig(baseCfg)
```
rambohe-ch, line 69: baseClient can be shared by all clients, so only one baseClient should be created for all tokens.

cmd/yurt-manager/app/client/token.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	// defaultExpirationSeconds defines the duration of a TokenRequest in seconds.
+	defaultExpirationSeconds = int64(3600)
+	// defaultLeewayPercent defines the percentage of expiration left before the client trigger a token rotation.
+	// range[0, 100]
+	defaultLeewayPercent = 20
+)
+
+// migrate from kubernetes/staging/src/k8s.io/controller-manager/pkg/clientbuilder/client_builder_dynamic.go
+// change client to controller-runtime client
```
rambohe-ch, line 44: it seems that `controller-runtime` client is not used, and kubernetes clientset client is still used.

luc99hen: done

cmd/yurt-manager/app/client/client.go
luc99hen: done

pkg/yurtmanager/webhook/deploymentrender/v1alpha1/deploymentrender_handler.go
rambohe-ch: ok, i got it.

rambohe-ch (approved):

PR #2074, opened 2024-06-06
use helm charts to install yurt-manager and yurthub components in a local kind cluster

rambohe-ch:
/kind enhancement

- improve `make local-up-openyurt` command:

   instead of using yaml constants, use helm charts to install yurt-manager and yurthub components in a local kind cluster.

Fixes #

```release-note

```


YTGhost (approved):
/lgtm
/approve

PR #2098, opened 2024-07-16
feat: cleanup cert/key info in tls config for ensuring service account will be used

rambohe-ch:
/kind feature

cleanup cert/key info in tls config of yurt-manager for ensuring serviceaccount will be used.

Fixes #

```release-note

```


zyjhtangtang (approved):

fungaren:
Why we need this? If we provide `--kubeconfig` option to  yurt-manager, this code makes yurt-manager fails to access the server.

rambohe-ch:
> Why we need this? If we provide `--kubeconfig` option to yurt-manager, this code makes yurt-manager fails to access the server.

@fungaren each controller will use unique service account to access kube-apiserver in order to manage rbac settings for each controller.

The case of `--kubeconfig` option can not work, I think this pull request maybe fixed this problem.
https://github.com/openyurtio/openyurt/pull/2193

PR #2111, opened 2024-08-03
fix(iot): the mount type of hostpath for localtime in napa

LavenderQAQ:
#### What type of PR is this?
/kind bug
/sig iot

In the napa version, the host's `/etc/localtime` file is mounted as a file, whereas previously the default bind type for `openyurt/auto-collector` was a folder. This pr is fixed to file.


YTGhost (approved):
/lgtm
/approve

PR #2119, opened 2024-08-07
[WIP] Issue 1597:Build iot system configuration isolation on nodepool

### LINKED ISSUE #1597: [OSPP 2024] Build iot system configuration isolation on nodepool

WoShiZhangmingyu:

Fixes #1597

@LavenderQAQ 

```release-note

```


rambohe-ch:
@WoShiZhangmingyu Thank you for posting the pull request, but please rebase the master branch before posting the pull request.

PR #2135, opened 2024-08-22
 Proposal: build iot system configuration isolation on nodepool(openyurtio#1597)

WoShiZhangmingyu:
#### Special notes for your reviewer:

@LavenderQAQ

PR #2147, opened 2024-09-09
feat: build iot system configuration isolation on nodepool

### LINKED ISSUE #1597: [OSPP 2024] Build iot system configuration isolation on nodepool

WoShiZhangmingyu:
#2135 
Fixes #1597 

@LavenderQAQ 
Yes, users need to change the original version of "pooName" to "nodePools" when creating platformadmin using "kubectl apply".


pkg/apis/iot/v1alpha1/platformadmin_conversion.go
```diff
@@ -36,7 +36,7 @@ func (src *PlatformAdmin) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.Version = src.Spec.Version
 	dst.Spec.Security = false
 	dst.Spec.ImageRegistry = src.Spec.ImageRegistry
-	dst.Spec.PoolName = src.Spec.PoolName
+	dst.Spec.NodePools = src.Spec.NodePools
```
LavenderQAQ, line 39: Instead of breaking the previous v1alpha1 and v1alpha2, we need a v1beta1 version here.

LavenderQAQ (changes requested):

pkg/yurthub/otaupdate/util/util_test.go
LavenderQAQ, line 1: Does this have nothing to do with this pr?

pkg/yurtmanager/webhook/platformadmin/v1alpha2/platformadmin_validation.go
```diff
@@ -143,4 +143,4 @@ func (webhook *PlatformAdminHandler) validatePlatformAdminWithNodePools(ctx cont
 	}
 
 	return nil
-}
+}
```
LavenderQAQ, line 146: It seems like there should be no modification here.

pkg/yurtmanager/webhook/platformadmin/v1alpha2/platformadmin_validation_test.go
```diff
@@ -383,4 +383,4 @@ func Indexer(rawObj client.Object) []string {
 		return []string{}
 	}
 	return []string{platformAdmin.Spec.PoolName}
-}
+}
```
LavenderQAQ, line 386: It seems like there should be no modification here.

LavenderQAQ (changes requested):

!
LavenderQAQ, line 1: Remove the files this pr don't need.

.swp
LavenderQAQ, line 1: Remove the files this pr don't need.

LavenderQAQ (changes requested):

test/e2e/cmd/init/converter.go
```diff
@@ -217,6 +217,7 @@ func (c *ClusterConverter) installYurtManagerByHelm() error {
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		klog.Errorf("couldn't install yurt-manager, %v", err)
+		klog.Errorf("Helm install output: %s", string(output))
```
LavenderQAQ, line 220: Remove unnecessary debug code.

LavenderQAQ (changes requested):

pkg/yurtmanager/webhook/server.go
```diff
@@ -70,14 +69,14 @@ func addControllerWebhook(name string, handler SetupWebhookWithManager) {
 	controllerWebhooks[name] = append(controllerWebhooks[name], handler)
 }
 
-func init() {
+func init() {
```
LavenderQAQ, line 72: Please remove this space.

LavenderQAQ (changes requested):

pkg/yurtmanager/controller/platformadmin/platform_admin_controller_test.go
LavenderQAQ, line 1: Lack of license.

LavenderQAQ (approved):
/lgtm

PR #2164, opened 2024-10-08
fix: create abspath dir in case that contents is empty

### LINKED ISSUE #2133: [BUG] diskStorage calls ReplaceComponentList function which deletes dir (not recreated again) when contents are empty
**What happened**:

diskStorage calls `ReplaceComponentList` function which deletes dir (not recreated again) when contents are empty.

For example, i installed `cilium` charts (version is 1.15.6). The cilium-agent component has a empty dir named `/etc/kubernetes/cache/cilium-agent/cilium2announcementpolicies.v2alpha1.cilium.io`. After diskStorage calls `ReplaceComponentList` function, the empty dir is renamed to `/etc/kubernetes/cache/cilium-agent/tmp_cilium2announcementpolicies.v2alpha1.cilium.io`.

Because the contents are empty, diskStorage does nothing and removes the `tmp` dir. So the original dir would not be recreated.

```golang
func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVersionResource, namespace string, contents map[storage.Key][]byte) error {
	rootKey, err := ds.KeyFunc(storage.KeyBuildInfo{
		Component: component,
		Resources: gvr.Resource,
		Group:     gvr.Group,
		Version:   gvr.Version,
		Namespace: namespace,
	})
	if err != nil {
		return err
	}
	storageKey := rootKey.(storageKey)

	for key := range contents {
		if !strings.HasPrefix(key.Key(), rootKey.Key()) {
			return storage.ErrInvalidContent
		}
	}

	if !ds.lockKey(storageKey) {
		return storage.ErrStorageAccessConflict
	}
	defer ds.unLockKey(storageKey)

	// 1. mv old dir into tmp_dir when rootKey dir already exists
	absPath := filepath.Join(ds.baseDir, storageKey.Key())
	tmpRootKey := getTmpKey(storageKey)
	tmpPath := filepath.Join(ds.baseDir, tmpRootKey.Key())
	if !fs.IfExists(absPath) {
		if err := ds.fsOperator.CreateDir(absPath); err != nil {
			return fmt.Errorf("could not create dir at %s", absPath)
		}
		if len(contents) == 0 {
			// nothing need to create, so just return
			return nil
		}
	}
	if ok, err := fs.IsDir(absPath); err == nil && !ok {
		return fmt.Errorf("%s is not a dir", absPath)
	} else if err != nil {
		return fmt.Errorf("could not check the path %s, %v", absPath, err)
	}
	// absPath exists and is a dir
	if err := ds.fsOperator.Rename(absPath, tmpPath); err != nil {
		return err
	}

	// 2. create new file with contents
	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
	for key, data := range contents {
		path := filepath.Join(ds.baseDir, key.Key())
		if err := ds.fsOperator.CreateDir(filepath.Dir(path)); err != nil && err != fs.ErrExists {
			klog.Errorf("could not create dir at %s, %v", filepath.Dir(path), err)
			continue
		}
		if err := ds.fsOperator.CreateFile(path, data); err != nil {
			klog.Errorf("could not write data to %s, %v", path, err)
			continue
		}
		klog.V(4).Infof("[diskStorage] ReplaceComponentList store data at %s", path)
	}

	//  3. delete old tmp dir
	return ds.fsOperator.DeleteDir(tmpPath)
```

it should be recreate the empty dir

**Environment**:
- OpenYurt version: v1.5.0
- Kubernetes version (use `kubectl version`): v1.27.4
- OS (e.g: `cat /etc/os-release`): centos linux
- Kernel (e.g. `uname -a`): 5.15.13-1.el7.elrepo.x86_64
- Install tools:
- Others:

**others**

/kind bug

vie-serendipity:
/kind bug
fix bug of issue 2133

Fixes #2133

```release-note
NONE
```

### DIFF
--- a/pkg/yurthub/storage/disk/storage.go
+++ b/pkg/yurthub/storage/disk/storage.go
@@ -384,11 +384,9 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 	// 2. create new file with contents
 	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
 	// create abspath dir in case that contents is empty
-	if len(contents) == 0 {
-		if err := ds.fsOperator.CreateDir(absPath); err != nil {
-			klog.Errorf("could not create dir at %s, %v", absPath, err)
-			return err
-		}
+	if err := ds.fsOperator.CreateDir(absPath); err != nil {
+		klog.Errorf("could not create dir at %s, %v", absPath, err)
+		return err
 	}
 	for key, data := range contents {
 		path := filepath.Join(ds.baseDir, key.Key())
--- a/pkg/yurthub/storage/disk/storage.go
+++ b/pkg/yurthub/storage/disk/storage.go
@@ -386,6 +386,7 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 	// create abspath dir in case that contents is empty
 	if err := ds.fsOperator.CreateDir(absPath); err != nil {
 		klog.Errorf("could not create dir at %s, %v", absPath, err)
+		return err
 	}
 	for key, data := range contents {
 		path := filepath.Join(ds.baseDir, key.Key())
--- a/pkg/yurthub/storage/disk/storage.go
+++ b/pkg/yurthub/storage/disk/storage.go
@@ -384,9 +384,11 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 	// 2. create new file with contents
 	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
 	// create abspath dir in case that contents is empty
-	if err := ds.fsOperator.CreateDir(absPath); err != nil {
-		klog.Errorf("could not create dir at %s, %v", absPath, err)
-		return err
+	if len(contents) == 0 {
+		if err := ds.fsOperator.CreateDir(absPath); err != nil {
+			klog.Errorf("could not create dir at %s, %v", absPath, err)
+			return err
+		}
 	}
 	for key, data := range contents {
 		path := filepath.Join(ds.baseDir, key.Key())
--- a/pkg/yurthub/storage/disk/storage.go
+++ b/pkg/yurthub/storage/disk/storage.go
@@ -383,6 +383,10 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 
 	// 2. create new file with contents
 	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
+	// create abspath dir in case that contents is empty
+	if err := ds.fsOperator.CreateDir(absPath); err != nil {
+		klog.Errorf("could not create dir at %s, %v", absPath, err)
+	}
 	for key, data := range contents {
 		path := filepath.Join(ds.baseDir, key.Key())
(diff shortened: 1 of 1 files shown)


rambohe-ch:
@vie-serendipity would you like to post a pull request to fix `Trivy Scan` error. you can check the error Details and find the error reason.

vie-serendipity:
@rambohe-ch ok

vie-serendipity:
It looks like the download of the vulnerability database failed due to some network issues. But after I tried again, everything is working fine now.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -383,6 +383,10 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 
 	// 2. create new file with contents
 	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
+	// create abspath dir in case that contents is empty
+	if err := ds.fsOperator.CreateDir(absPath); err != nil {
+		klog.Errorf("could not create dir at %s, %v", absPath, err)
+	}
```
rambohe-ch, line 389: If `absPath` directory can not be created, It should be short-circuit and fail immediately.
so please add `return err` here.

pkg/yurthub/storage/disk/storage.go
```diff
@@ -383,6 +383,10 @@ func (ds *diskStorage) ReplaceComponentList(component string, gvr schema.GroupVe
 
 	// 2. create new file with contents
 	// TODO: if error happens, we may need retry mechanism, or add some mechanism to do consistency check.
+	// create abspath dir in case that contents is empty
+	if err := ds.fsOperator.CreateDir(absPath); err != nil {
```
rambohe-ch, line 387: How about add the condition for verify content is empty as following:
```
if len(contents) == 0 {
    if err := ds.fsOperator.CreateDir(absPath); err != nil {
        // xxxx
    }
}
```

pkg/yurthub/storage/disk/storage.go
vie-serendipity: When the creation of the following contents fails, it chooses to continue. What's the difference between these two?
```go
if err := ds.fsOperator.CreateDir(filepath.Dir(path)); err != nil && err != fs.ErrExists {
    klog.Errorf("could not create dir at %s, %v", filepath.Dir(path), err)
    continue
}
```

pkg/yurthub/storage/disk/storage.go
vie-serendipity: ok

rambohe-ch: we can not cover all failed cases, and I think it is reasonable to create `absPath` directory at first even if content is not empty.  so I want to take back this comment.

rambohe-ch (approved):

pkg/yurthub/storage/disk/storage.go
rambohe-ch: If it failed to create the root directory, I think it's not need to create the sub files. so we can make a short-circuit here.

vie-serendipity: ok, it makes sense.

pkg/yurthub/storage/disk/storage.go
vie-serendipity: done

PR #2124, opened 2024-08-14
proposal: enhance operational efficiency of K8s cluster in user's IDC

huangchenzhao:
#### What this PR does / why we need it:
It is difficult for users to operate, manage and upgrade the K8s cluster in their own IDC (Internet Data Center). The proposal aims to enhance the operational efficiency of K8s cluster in user's IDC by adopting KOK (Kubernetes-On-Kubernetes) architecture shown in the following figure, where there are host-K8s and tenant-K8s. Host-K8s is located at cloud, provided by cloud service providers and can manage control plane of tenant-K8s in IDC.
image

PR #2190, opened 2024-11-18
feat: support yurtadm to join nodes to tenant-K8s in local mode

huangchenzhao:
#2124 
##### What is new:

1. Support deploying local mode yurthub in systemd, which list/watch endpoints to configure iptables rules for loadbalancing.
2. In yurtadm join, we first wait for local mode yurthub to be ready and iptables rules to be ready. Then we invoke `kubeadm join` join nodes to cluster, including kubelet bootstrap and so on. At last in the postcheck phase we modify kubelet.conf and set address to VIP then restart kubelet, so the traffic from kubelet to apiservers can be loadbalanced.

##### TODO and questions:

1. The implement of loadbalacing in yurtadm join relies on modifing kubelet.conf, which seems not elegant. The original idea is that we first do some preparation, then use `kubeadm join` to join the node. Because all apiservers' certificates are the same, invoking `kubeadm join` to join the cluster allows the node to do TLS bootstrap to obtain the kubelet certificate. Finally, modify kubelet.conf and restart kubelet, kubelet will do 4-layer load balancing and establish a TCP connection with one of the apiservers.
2. Yurtadm join is a one-time operation, so what will happen when kubelet rotate it's certificates, maybe loadbalance will be invalid. 
3. How do other components like kube-proxy, addons, pods access apiserver in loadbalance mode? Simply modifying yurtadm join seems to unable to afford these components for loadbalancing.
/assign @rambohe-ch

### DIFF
--- a/pkg/yurtadm/cmd/join/join.go
+++ b/pkg/yurtadm/cmd/join/join.go
@@ -20,6 +20,7 @@ package join
 import (
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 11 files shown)


huangchenzhao:
@rambohe-ch PTAL

PR #2209, opened 2024-11-28
fix: support cache response for partial object metedata watch request

### LINKED ISSUE #2199: [feature request]support partialobjectmetadata watch request
In this [pull request](https://github.com/openyurtio/openyurt/pull/2170), partialobjectmetadata get/list requests are supported in yurthub, but partialobjectmetadata watch request is not supported currently.

As we known, list/watch requests are used as a pair requests, so it is necessary to support partialobjectmetadata watch request in yurthub.

**others**
/kind feature

rambohe-ch:
/kind feature

improve saveWatchObject function in cache manager for supporting paritialobjectmetadata watch requests.

Fixes #2199 

```release-note

```

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -360,7 +360,18 @@ func (cm *cacheManager) saveWatchObject(ctx context.Context, info *apirequest.Re
 
 	comp, _ := util.ClientComponentFrom(ctx)
 	respContentType, _ := util.RespContentTypeFrom(ctx)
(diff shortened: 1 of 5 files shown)


vie-serendipity:
https://github.com/openyurtio/openyurt/blob/09078b4e23f17b313a8039ec7a2f9aea1d705f1c/pkg/yurthub/cachemanager/cache_manager.go#L356

BTW, I noticed that stopCh parameter is not used in saveWatchObject function. I think this could be optimzed.

zyjhtangtang (approved):

PR #2275, opened 2025-01-20
feat: improve yurthub configmap management

### LINKED ISSUE #2274: [feature request]improve yurt-hub-cfg configmap handling in yurthub component.

- background:

`kube-system/yurt-hub-cfg` configmap is used for managing dynamic configurations of Yurthub component, like cache agents and filter settings.

The current state of handling this configmap are scattered in different areas. the detailed handling as following:

- cache agents:

https://github.com/openyurtio/openyurt/blob/d9186066d30626229ed8b31a30d1ef08b08b8758/pkg/yurthub/cachemanager/cache_agent.go#L50

- filter settings:

https://github.com/openyurtio/openyurt/blob/d9186066d30626229ed8b31a30d1ef08b08b8758/pkg/yurthub/filter/approver/approver.go#L73 

- improvement:

If new configurations are added in this configmap for new requirements, new event handler will be added for processing these new configurations. There are a lot of duplicated codes for handling the same configmap.

so it is obvious that it is a good idea to new configuration manager for yurthub which is used for managing this configmap, and provide functions for each configuration.

**others**
/kind feature

rambohe-ch:
/kind feature

1. Add a new manager named configuration manager in yurthub for managing `kube-system/yurt-hub-cfg` configmap, and provide functions for cache manager and filter approver.

2. There is no change of API, it means this pull request is transparent to end users. and this pull request can make the code more readable and maintainable. the efficiency of yurthub is also improved.

3. The removal of cache should be managed by gc when cache agents configuration is changed. now this feature is not supported.

Fixes #2274 

```release-note

```


rambohe-ch:
@vie-serendipity  @zyjhtangtang PTAL

pkg/yurthub/cachemanager/cache_manager.go
```diff
@@ -77,25 +77,23 @@ type cacheManager struct {
 	storage               StorageWrapper
 	serializerManager     *serializer.SerializerManager
 	restMapperManager     *hubmeta.RESTMapperManager
-	cacheAgents           *CacheAgent
+	manager               *configuration.Manager
```
vie-serendipity, line 80: The name `manager` might sound vague, how about configManager, it sounds more detailed.

pkg/yurthub/configuration/manager.go
```diff
@@ -0,0 +1,289 @@
...
+	newAgents := make([]string, 0)
+	newAgents = append(newAgents, m.baseAgents...)
+	for _, agent := range strings.Split(cacheAgents, sepForAgent) {
+		agent = strings.TrimSpace(agent)
+		if len(agent) != 0 {
+			newAgents = append(newAgents, agent)
+		}
+	}
+
+	klog.Infof("%s cache agents: %v", action, newAgents)
```
vie-serendipity, line 194: When action is delete, this looks like newAgents are deleted. So I think this log should be optimzed.
```suggestion
	klog.Infof("After action %s, the cache agents are as follows: %v", action, newAgents)
```

pkg/yurthub/configuration/manager.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+		}
+	}
+
+	reqKeyToFilters := make(map[string][]string)
+	for key, filterSet := range reqKeyToFilterSet {
+		reqKeyToFilters[key] = filterSet.UnsortedList()
+	}
+
+	klog.Infof("%s filter settings: %v", action, reqKeyToFilters)
```
vie-serendipity, line 253: Same as `updateCacheAgents`, I think this log should optimzed.

pkg/yurthub/configuration/manager.go
```diff
@@ -0,0 +1,289 @@
...
+	configmapInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    m.addConfigmap,
+		UpdateFunc: m.updateConfigmap,
+		DeleteFunc: m.deleteConfigmap,
+	})
+	return m
+}
+
+// IsReady is used for checking that configuration of Yurthub has been loaded completed or not.
+// Yurthub should reject all requests from components on the node before it is ready.
```
vie-serendipity, line 99: I didn't find the relevant implementation. This function is called by approve, but if approve returns false, it only prevents the filter from taking effect.
https://github.com/openyurtio/openyurt/blob/470292f503df16c3ca4f8e553b4b6473ed0beb68/pkg/yurthub/proxy/remote/loadbalancer.go#L295

pkg/yurthub/cachemanager/cache_manager.go
rambohe-ch: fixed

pkg/yurthub/configuration/manager.go
rambohe-ch: fixed

pkg/yurthub/configuration/manager.go
rambohe-ch: fixed

pkg/yurthub/configuration/manager.go
rambohe-ch: yes, I will post another pull request to improve readinessprobe of yurthub which include this feature. so I remove this comment in this pull request.

Issue #2259, opened 2025-01-08
[feature request]add HubLeaderElection Controller in yurt-manager component.


rambohe-ch:
**What would you like to be added**:
According to the following proposal,  A new controller named HubLeaderElection should be added for electing a leader Yurthub in a nodepool according to election strategy.

- only elect a leader Yurthub for nodepool which field interConnectivity is true.
- write address of leader Yurthub into NodePool status.LeaderEndpoints.
- update NodePool status.conditions according to Leader Election results.

https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20241211-reuse-nodepool-list-watch-requests.md#reuse-listwatch-requests-in-nodepool

**others**
/kind feature

PR #2281, opened 2025-01-29
feat: add hub election leader controller

tnsimon:
#### What type of PR is this?
/kind feature

Add hub leader election controller

Fixes https://github.com/openyurtio/openyurt/issues/2259

/assign @rambohe-ch 

```release-note
NONE
```


pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	"github.com/openyurtio/openyurt/cmd/yurt-manager/names"
+	appsv1beta2 "github.com/openyurtio/openyurt/pkg/apis/apps/v1beta2"
+	"github.com/openyurtio/openyurt/pkg/projectinfo"
+	"github.com/openyurtio/openyurt/pkg/yurtmanager/controller/hubleader/config"
+	nodeutil "github.com/openyurtio/openyurt/pkg/yurtmanager/controller/util/node"
+)
+
+func init() {
+	flag.IntVar(
+		&concurrentReconciles,
```
rambohe-ch, line 49: move concurrentReconciles definition into `HubLeaderControllerConfiguration` struct.  please take nodepool configuration for example: https://github.com/openyurtio/openyurt/blob/d6908080a4f6ba70243b47210b60c644a6e34158/pkg/yurtmanager/controller/nodepool/config/types.go#L22

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+	controllerKind       = appsv1beta2.SchemeGroupVersion.WithKind("Nodepool")
+)
+
+// Add creates a new HubLeader Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(ctx context.Context, cfg *appconfig.CompletedConfig, mgr manager.Manager) error {
+	klog.Infof("hubleader-controller add controller %s", controllerKind.String())
+
+	reconciler := &ReconcileHubLeader{
+		Client:        mgr.GetClient(),
```
rambohe-ch, line 67: mgr.GetClient() --> yurtClient.GetClientByControllerNameOrDie(mgr, names. HubLeaderController), because this client will use the correct User-Agent which match the clusterrole of HubLeader controller.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+// ReconcileHubLeader reconciles a HubLeader object
+type ReconcileHubLeader struct {
+	client.Client
+	scheme        *runtime.Scheme
+	recorder      record.EventRecorder
+	Configuration config.HubLeaderControllerConfiguration
+}
+
+// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepool,verbs=get;list;watch;update;patch
```
rambohe-ch, line 144: please delete `list; watch`, because rbac settings of list/watch resource is managed by base controller: https://github.com/openyurtio/openyurt/blob/d6908080a4f6ba70243b47210b60c644a6e34158/pkg/yurtmanager/controller/base/controller.go#L108

btw: list/watch requests are posted by `mgr.GetClient()` instead of each controller client.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepool/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list
```
rambohe-ch, line 146: please delete `list`

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+	// Watch for changes to NodePool
+	err = c.Watch(
+		source.Kind[client.Object](
+			mgr.GetCache(),
+			&appsv1beta2.NodePool{},
+			&handler.EnqueueRequestForObject{},
+			poolPredicate,
+		),
+	)
```
rambohe-ch, line 126: we also need to list/watch nodes for the following reasons:
1. if leader node becomes not ready, we need to re-elect a new leader.
2. if the current leader replicas doesn't satisfy the nodePool.Spec.LeaderReplicas, we need to reconcile the nodePool when node add/update.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+			if numLeaders == 0 {
+				break
+			}
+		}
+	case string(appsv1beta2.ElectionStrategyRandom):
+		// Randomly select a leader node
+		// Iterating slice in Go is non-deterministic enough to be considered random
+		// for this purpose
+		for endpoint := range candidates {
+			leaderEndpoints = append(leaderEndpoints, endpoint)
```
rambohe-ch, line 298: we also need to elect numLeaders even in random strategy.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,307 @@
...
+		_, ok := endpointsMap[endpoint]
+		return !ok
+	}
+	updatedLeaders := slices.DeleteFunc(updatedNodePool.Status.LeaderEndpoints, leaderDeleteFn)
+
+	// If there is no leader, or number of leaders are less than desired elect a new leader
+	if len(updatedLeaders) == 0 ||
+		len(updatedLeaders) < int(nodepool.Spec.LeaderReplicas) &&
+			nodepool.Spec.LeaderElectionStrategy == string(appsv1beta2.ElectionStrategyMark) {
+		// No leader, it should elect a new leader
```
rambohe-ch, line 225: the condition should be `len(updatedLeaders) != int(nodepool.Spec.LeaderReplicas)` for random and mark strategy.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Removed

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Removed

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: 1. Added if there's a change in UnreadyNodeNum/ReadyNodeNum we will elect new leaders. 
2. We cannot check whether we satisy the leader replicas without reconciling as this controller is responsible for updating leader endpoints in the nodepool.Status. So instead we're covered by #1 if there's nodes that became ready/unready.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Ok design spec indicated random would be 1. Updated the logic to cover both for leader replicas.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleader/config/types.go
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+// HubLeaderControllerConfiguration contains elements describing HubLeaderController.
+type HubLeaderControllerConfiguration struct {
+	ConcurrentNodePoolWorkers int32
```
rambohe-ch, line 21: ConcurrentNodePoolWorkers --> ConcurrentHubLeaderElectionWorkers

btw: we also need to configure this field in the options file: cmd/yurt-manager/app/options/hubleadercontroller.go

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
rambohe-ch: It's a good idea to monitor UnreadyNodeNum/ReadyNodeNum field change of nodepool.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
rambohe-ch: list/watch is still left, please remove them from rbac settings.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,291 @@
...
+	leaderDeleteFn := func(endpoint string) bool {
+		_, ok := endpointsMap[endpoint]
+		return !ok
+	}
+	updatedLeaders := slices.DeleteFunc(updatedNodePool.Status.LeaderEndpoints, leaderDeleteFn)
+
+	// If the number of leaders is not equal to the desired number of leaders
+	if len(updatedLeaders) != int(nodepool.Spec.LeaderReplicas) {
+		// No leader, it should elect a new leader
+		leaders, ok := electLeaders(nodepool.Spec.LeaderElectionStrategy, nodepool.Spec.LeaderReplicas, endpointsMap)
```
rambohe-ch, line 216: we should make sure original valid leaders(in updatedLeaders var) are kept in the new elected leaders. this means we can only delete redundant leaders or add new leaders, but original valid leaders should not be changed.

rambohe-ch: change a leader will cost a lot on the edge side.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -0,0 +1,291 @@
...
+
+// Add creates a new HubLeader Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(ctx context.Context, cfg *appconfig.CompletedConfig, mgr manager.Manager) error {
+	klog.Infof("hubleader-controller add controller %s", controllerKind.String())
+
+	reconciler := &ReconcileHubLeader{
+		Client:        yurtClient.GetClientByControllerNameOrDie(mgr, names.HubLeaderController),
+		scheme:        mgr.GetScheme(),
+		recorder:      mgr.GetEventRecorderFor(names.HubLeaderController),
```
rambohe-ch, line 59: It seems that fields(scheme and recorder) are unused by reconciler. how about remove unused fields.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Got it - removed list/watch for nodepools.

pkg/yurtmanager/controller/hubleader/config/types.go
tnsimon: Renamed and added to cmd/yurt-manager/app/options.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: Yeah good point - I updated the algorithm to keep already elected leaders that are still valid.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
tnsimon: I kept recorder as it's writing an event for reconcile error. Do we want to keep the recorder events?

rambohe-ch: ok. If the controller report a large mount of events, we can optimize in future version.

rambohe-ch (approved):

PR #2285, opened 2025-02-01
feat: improve direct clientsets for yurthub

rambohe-ch:
/kind enhancement

Use Clientset directly instead of using rest config in order to improve the performance of yurthub. because yurthub.conf file  should be loaded when rest config is used each time.

Fixes #

```release-note

```


rambohe-ch:
@zyjhtangtang @vie-serendipity PTAL

rambohe-ch:
@zyjhtangtang  @vie-serendipity PTAL

zyjhtangtang (approved):

Issue #2295, opened 2025-02-03
[feature request] Store node name in v1beta2.Nodepool spec


tnsimon:
**What would you like to be added**:
Store node name in v1beta2.Nodepool spec.

**Why is this needed**:
When yurt-hub starts, it needs to know whether it is the leader. Currently only the node Internal Ip is stored. Keeping the node name also will help. 

**others**
/kind feature

rambohe-ch:
@tnsimon Yes, It's a good idea to add node name information into leader endpoints, then Yurthub component can resolve this information for deciding the forwarded destination of requests for pool scope requests(like list/watch services, endpointslices).

PR #2297, opened 2025-02-04
feat: add leader node names to v1beta2.nodepool

tnsimon:
#### What type of PR is this?
/kind feature

Store node name with leader for yurthub.

Fixes https://github.com/openyurtio/openyurt/issues/2295

/assign @rambohe-ch 

```release-note
NONE
```


pkg/apis/apps/v1beta2/nodepool_types.go
```diff
@@ -107,14 +107,23 @@ type NodePoolStatus struct {
...
 	Conditions []NodePoolCondition `json:"conditions,omitempty"`
 }
 
+// Leader represents the hub leader in a nodepool
+type Leader struct {
+	// The node name of the leader yurthub
+	NodeName string `json:"nodeName"`
+
+	// The address of the leader yurthub
+	Endpoint string `json:"endpoint"`
```
rambohe-ch, line 124: how about rename Endpoint to Address?

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -270,9 +273,9 @@ func hasLeadersChanged(old, new []string) bool {
 func electNLeaders(
 	strategy string,
 	numLeaders int,
-	candidates map[string]*corev1.Node,
-) ([]string, bool) {
-	leaderEndpoints := make([]string, 0, len(candidates))
+	candidates map[appsv1beta2.Leader]*corev1.Node,
+) ([]appsv1beta2.Leader, bool) {
+	leaderEndpoints := make([]appsv1beta2.Leader, 0, len(candidates))
```
rambohe-ch, line 278: make([]appsv1beta2.Leader, 0, len(candidates)) --> make([]appsv1beta2.Leader, 0)

because only a part of candidates will be elected as leaders.

tnsimon: I sized it to the numLeaders to be elected

rambohe-ch: but candidates maybe less then numLeaders, so there is also dummy memory.

tnsimon: I added return early if there are no candidates to elect we won't pre-allocate the slice.

Pre-allocation of slices wherever possible is a positive optimisation. There's an article below about the positive performance for doing it where possible. Our figures here should be small anyway for the dummy memory.

**Let's presume:**
* A character in a string takes **1 byte** of memory (ASCII)
* Kubernetes allows **253** characters max for a node name
* An IPv4 address represented as a string with **15** characters e.g. "255.255.255.255"
* A user has set **5000** desired leader replicas i.e. need to make room for 5000 leaders in slice.
 
**To store both NodeName and IP:**
5000 (nodes) x (253 (node name characters) x 1 byte (per character) + 15 (number of characters to represent IPv4 address as a string) x 1 byte (per character))  ~ **1.34 MB**

There's an article with benchmarks why pre-allocation is preferred. I can move it to 0 capacity if still preferred. 

https://oilbeater.com/en/2024/03/04/golang-slice-performance/#:~:text=Overall%2C%20pre%2Dallocating%20slice%20memory,can%20still%20significantly%20improve%20performance.

rambohe-ch: @tnsimon Thank you for your clarification. I have learned the reason of pre-allocate capacity for slice, in order to get rid of reallocating memory for slice.

rambohe-ch (approved):

Issue #2260, opened 2025-01-08
[feature request]add a controller in yurt-manager component for managing address of leader Yurthub and leader election configurations.


rambohe-ch:
**What would you like to be added**:
Leader election configurations(like interConnectivity and poolScopeMetadata) and address of leader Yurthub is managed in the NodePool, but NodePool metadata is changed frequently because of status change. so if Yurthub on edge nodes get these information through list/watch NodePool, it will cause large network traffic between cloud-edge. moreover, These information can not be shared in the nodepool and each Yurthub should get this information from cloud kube-apiserver directly.

so we want to define Endpoints to manage these information, and all Yurthub only need list/watch this specified Endpoints resource to resolve these information about leader election configuration and address of leader yurthub.

- Input for controller: NodePool
- Output for controller: Endpoints
  1.  each NodePool has a related Endpoints metadata, the name of Endpoints is: `leader-hub-{nodepool-name}`
  2. NodePool.Spec.InterConnectivity/NodePool.Spec.PoolScopeMetadata are stored in the Endpoints annotation
  3. NodePool.Status.LeaderEndpoints are store in the Endpoints.Subsets.

**others**
/kind feature

PR #2299, opened 2025-02-05
feat: add hub leader config controller

tnsimon:
#### What type of PR is this?
/kind feature

Adds hubleader config controller to create configmap for the nodepool leader and pool scoped metadata

Fixes https://github.com/openyurtio/openyurt/issues/2260

/assign @rambohe-ch 

```release-note
NONE
```

#### other Note

An internal CRD would be an improvement than a configmap to have a structured schema. Endpoints resource was not used due to character limitations using annotations to store poolscoped metadata. It was decided to use a config map instead.

### DIFF
--- a/cmd/yurt-manager/app/options/hubleaderconfigcontroller.go
+++ b/cmd/yurt-manager/app/options/hubleaderconfigcontroller.go
@@ -26,10 +26,11 @@ type HubLeaderConfigControllerOptions struct {
 	*config.HubLeaderConfigControllerConfiguration
 }
 
-func NewHubLeaderConfigControllerOptions() *HubLeaderConfigControllerOptions {
+func NewHubLeaderConfigControllerOptions(hubleaderNamespace string) *HubLeaderConfigControllerOptions {
 	return &HubLeaderConfigControllerOptions{
 		&config.HubLeaderConfigControllerConfiguration{
 			ConcurrentHubLeaderConfigWorkers: 3,
+			HubLeaderNamespace:               hubleaderNamespace,
 		},
 	}
 }
@@ -55,6 +56,7 @@ func (h *HubLeaderConfigControllerOptions) ApplyTo(cfg *config.HubLeaderConfigCo
 	}
 
 	cfg.ConcurrentHubLeaderConfigWorkers = h.ConcurrentHubLeaderConfigWorkers
+	cfg.HubLeaderNamespace = h.HubLeaderNamespace
 
 	return nil
 }
--- a/cmd/yurt-manager/app/options/hubleaderconfigcontroller.go
+++ b/cmd/yurt-manager/app/options/hubleaderconfigcontroller.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2025 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the License);
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an AS IS BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"github.com/spf13/pflag"
+
+	"github.com/openyurtio/openyurt/pkg/yurtmanager/controller/hubleaderconfig/config"
+)
+
+type HubLeaderConfigControllerOptions struct {
+	*config.HubLeaderConfigControllerConfiguration
+}
+
+func NewHubLeaderConfigControllerOptions() *HubLeaderConfigControllerOptions {
+	return &HubLeaderConfigControllerOptions{
+		&config.HubLeaderConfigControllerConfiguration{
+			ConcurrentHubLeaderConfigWorkers: 3,
+		},
+	}
+}
+
+// AddFlags adds flags related to hubleader for yurt-manager to the specified FlagSet.
+func (h *HubLeaderConfigControllerOptions) AddFlags(fs *pflag.FlagSet) {
+	if h == nil {
+		return
+	}
+
+	fs.Int32Var(
+		&h.ConcurrentHubLeaderConfigWorkers,
+		"concurrent-hubleaderconfig-workers",
+		h.ConcurrentHubLeaderConfigWorkers,
+		"The number of nodepool objects that are allowed to reconcile concurrently.",
+	)
+}
+
+// ApplyTo fills up hubleader config with options.
+func (h *HubLeaderConfigControllerOptions) ApplyTo(cfg *config.HubLeaderConfigControllerConfiguration) error {
+	if h == nil {
+		return nil
+	}
+
+	cfg.ConcurrentHubLeaderConfigWorkers = h.ConcurrentHubLeaderConfigWorkers
+
+	return nil
+}
+
+// Validate checks validation of HubLeaderControllerOptions.
+func (h *HubLeaderConfigControllerOptions) Validate() []error {
+	if h == nil {
+		return nil
+	}
+	errs := []error{}
+	return errs
+}
--- a/pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
+++ b/pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
@@ -137,8 +137,9 @@ type ReconcileHubLeaderConfig struct {
 	Configuration config.HubLeaderConfigControllerConfiguration
 }
 
-// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepool,verbs=get
-// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepool/status,verbs=get
+// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepools,verbs=get
+// +kubebuilder:rbac:groups=apps.openyurt.io,resources=nodepools/status,verbs=get
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;update;patch;create
 
 // Reconcile reads that state of the cluster nodepool leader status and updates the leader configmap object
 func (r *ReconcileHubLeaderConfig) Reconcile(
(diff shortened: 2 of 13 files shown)


cmd/yurt-manager/app/options/hubleaderconfigcontroller.go
```diff
@@ -0,0 +1,71 @@
...
+
+type HubLeaderConfigControllerOptions struct {
+	*config.HubLeaderConfigControllerConfiguration
+}
+
+func NewHubLeaderConfigControllerOptions(hubleaderNamespace string) *HubLeaderConfigControllerOptions {
+	return &HubLeaderConfigControllerOptions{
+		&config.HubLeaderConfigControllerConfiguration{
+			ConcurrentHubLeaderConfigWorkers: 3,
+			HubLeaderNamespace:               hubleaderNamespace,
```
rambohe-ch, line 33: we can use `config.GenericConfiguration.WorkingNamespace` directly instead of defining a new one.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+
+	poolPredicate := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			newPool, ok := e.Object.(*appsv1beta2.NodePool)
+			if !ok {
+				return false
+			}
+
+			return newPool.Spec.InterConnectivity &&
+				len(newPool.Status.LeaderEndpoints) > 0
```
rambohe-ch, line 83: we need to create a configmap for each nodepool in order to store leader election information, regardless of InterConnectivity and status.LeaderEndpoints.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+				return false
+			}
+
+			// Only update if the leader has changed or the pool scope metadata has changed
+			return !slices.EqualFunc(
+				oldPool.Status.LeaderEndpoints,
+				newPool.Status.LeaderEndpoints,
+				func(s1, s2 appsv1beta2.Leader) bool {
+					return s1.NodeName == s2.NodeName && s1.Address == s2.Address
+				},
```
rambohe-ch, line 104: if contents are the same but Leader order is different, It's no need to reconcile nodepool for configmap.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+			) || !slices.EqualFunc(
+				oldPool.Spec.PoolScopeMetadata,
+				newPool.Spec.PoolScopeMetadata,
+				func(s1, s2 metav1.GroupVersionKind) bool {
+					return s1.Group == s2.Group && s1.Version == s2.Version && s1.Kind == s2.Kind
```
rambohe-ch, line 109: ditto

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+		if err != nil && !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+
+	if !nodepool.Spec.InterConnectivity {
+		// If the NodePool is not using interconnectivity, it should not reconcile
+		return reconcile.Result{}, nil
+	}
```
rambohe-ch, line 175: Regardless of InterConnectivity field, we need to create a configmap for each nodepool for storing leader election information.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileHubLeaderConfig) reconcileHubLeaderConfig(
+	ctx context.Context,
+	nodepool *appsv1beta2.NodePool,
+) error {
+	if len(nodepool.Status.LeaderEndpoints) == 0 {
+		return nil
+	}
```
rambohe-ch, line 192: If no LeaderEndpoints, we need to create or update configmap with no Leaders.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+
+		// Create the ConfigMap resource
+		return r.Create(ctx, leaderConfigMap)
+	}
+
+	// If the ConfigMap exists, update it with the new data
+	leaderConfigMap.Data = data
+
+	// Update the ConfigMap resource
+	return r.Update(ctx, leaderConfigMap)
```
rambohe-ch, line 244: we update configmap only when configmap is changed.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	// Add pool scope metadata
+	poolScopedMetadata := make([]string, 0, len(nodepool.Spec.PoolScopeMetadata))
+	for _, metadata := range nodepool.Spec.PoolScopeMetadata {
+		poolScopedMetadata = append(poolScopedMetadata, getGVKString(metadata))
+	}
+
+	// Prepare data
+	data := map[string]string{
+		"leaders":              strings.Join(leaders, ","),
+		"pool-scoped-metadata": strings.Join(poolScopedMetadata, ","),
```
rambohe-ch, line 223: InterConnectivity field should also stored in the configmap, because yurthub should decides the forward destination of requests for pool scope metadata according to this field.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	// If the ConfigMap does not exist, create it
+	if errors.IsNotFound(err) {
+		leaderConfigMap = &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapName,
+				Namespace: r.Configuration.HubLeaderNamespace,
+			},
```
rambohe-ch, line 232: how about add a labels for the configmap as following:

```
Labels: map[string]string{
    "openyurt.io/configmap-name": "leader-hub-{pool-name}"
}
```

Yurthub will use label selector to list/watch two configmaps, the other one is yurt-hub-cfg configmap.

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Moved hasLeaderChanged to nodepool util and created differ for poolscoped metadata to check this

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Ok

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Fixed

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Added

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Added

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -0,0 +1,238 @@
...
+				},
+			},
+			Data: data,
+		}
+
+		// Create the ConfigMap resource
+		return r.Create(ctx, leaderConfigMap)
+	}
+
+	if !maps.Equal(leaderConfigMap.Data, data) {
```
rambohe-ch, line 226: If the contents of slices `Status.LeaderEndpoints` or `Spec.PoolScopeMetadata` are the same but the element positions are different, they are essentially unchanged. However, the comparison result here would indicate that the data has changed. Therefor, These two slices should be sorted before constructing the data to eliminate the impact of element position differences.

Additionally, although the Update event predicate already includes a check, restarting yurt-manager triggers an Add event, which may still lead to unnecessary ConfigMap updates. The predicate primarily reduces unnecessary reconcile operations, while the check within reconcile is intended to minimize unnecessary ConfigMap updates.

pkg/yurtmanager/controller/util/nodepool/nodepool.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+
+	return false
+}
+
+// HasPoolScopedMetadataChanged checks if the metadata has changed
+// between the old and new.
+// It returns true if the metadata has changed, otherwise false irrespective of the order of the metadata.
+func HasPoolScopedMetadataChanged(old, new []metav1.GroupVersionKind) bool {
```
rambohe-ch, line 50: how to use Generics to merge these two functions as follwing:

```
func HasSliceContentChanged[T comparable](old, new []T) bool {
    if len(old) != len(new) {
        return true
    }

    oldSet := make(map[T]struct{}, len(old))
    for _, v := range old {
        oldSet[v] = struct{}{}
    }

    for _, v := range new {
        if _, ok := oldSet[v]; !ok {
            return true
        }
    }

    return false
}
```

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Ah I see it's the order of the string structure built that is out of order.

rambohe-ch: I think we need to sort `Status.LeaderEndpoints` and `Spec.PoolScopeMetadata` before building leaders and poolScopedMetadata string.

tnsimon: Yeah I added that in. The leaders and pools are sorted before config map data is created/compared for update. 

```go
	// Prepare data
	data := map[string]string{
		"leaders":              strings.Join(slice.SortStrings(leaders), ","),
		"pool-scoped-metadata": strings.Join(slice.SortStrings(poolScopedMetadata), ","),
		"interconnectivity":    strconv.FormatBool(nodepool.Spec.InterConnectivity),
	}
```

rambohe-ch (approved):

PR #2301, opened 2025-02-06
feat: improve proxy handler of yurthub

rambohe-ch:
/kind feature

1. remove latency metrics(proxyLatencyCollector) of yurthub, because this metric is not useful for end users.
2. remove WithMaxInFlightLimit handler, because requests are aggregated by multiplexer in yurthub, so requests limit is not needed for controlling requests volume.
3.  remove WithCacheHeaderCheck handler, because Edge-Cache header is rarely used.
4. improve WithRequestClientComponent handler, add whole user agent into request context, and truncate user agent by callers.
5. improve proxy handler not to depend on working mode.

Fixes #

```release-note

```


rambohe-ch:
@zyjhtangtang @YTGhost PTAL

zyjhtangtang (approved):

Issue #2300, opened 2025-02-06
[feature request] add e2e test for nodepool hub leader


tnsimon:
**What would you like to be added**:
Add e2e test for nodepool hub leader

**Why is this needed**:
To verify proposal https://github.com/openyurtio/openyurt/blob/master/docs/proposals/20241211-reuse-nodepool-list-watch-requests.md

**others**
/kind feature

PR #2302, opened 2025-02-06
feat: Add hubleader e2e test

tnsimon:
#### What type of PR is this?
/kind feature

Add e2e testing for hub leader controller in yurtmanager

Fixes https://github.com/openyurtio/openyurt/issues/2300

/assign @rambohe-ch 

```release-note
None
```

#### other Note
* Worker 3 and 4 are added to the end to end init phase.
* Both these workers are added to hubleadere2e pool
* util.InitNodeAndNodePool is refactored - we never seem to use multiple nodepools in any e2e so that map data structure wasn't useful. 
* There hub leader e2e must be run sequentially as the nodepool CRD is modified within the same test and conflicts. This can be avoided; but at the cost of adding more worker nodes in kind and another pool. 
* Please merge after https://github.com/openyurtio/openyurt/pull/2299


rambohe-ch:
@tnsimon please fix e2e tests failure.

test/e2e/yurt/hubleader.go
```diff
@@ -0,0 +1,254 @@
...
+	getExpectedLeaderConfig := func(leaders []v1beta2.Leader) map[string]string {
+		expectedLeaderConfig := make(map[string]string)
+
+		leaderEndpoints := make([]string, 0, len(leaders))
+		for _, leader := range leaders {
+			leaderEndpoints = append(leaderEndpoints, leader.NodeName+"/"+leader.Address)
+		}
+
+		expectedLeaderConfig["leaders"] = strings.Join(leaderEndpoints, ",")
+		expectedLeaderConfig["pool-scoped-metadata"] = "core/v1/services,discovery.k8s.io/v1/endpointslices"
```
rambohe-ch, line 104: core/v1/services --> /v1/services

I think the group for services resource is "".

test/e2e/yurt/hubleader.go
```diff
@@ -0,0 +1,254 @@
...
+	Context("Mark strategy", func() {
+		It("should elect the marked node correctly", Serial, func() {
+			By("Update the node pool spec with mark election strategy and worker 3 as the marked leader")
+			Eventually(
+				retry.RetryOnConflict(
+					retry.DefaultRetry,
+					updateNodePoolSpec(
+						v1beta2.NodePoolSpec{
+							LeaderElectionStrategy: string(v1beta2.ElectionStrategyMark),
+							LeaderReplicas:         2,
```
rambohe-ch, line 226: 2 --> 1

tnsimon: Setting to 2 exercises that there are 2 nodes available for selection as leader, though one is marked. So this setup ensures we verify that there is only 1 marked leader elected.

test/e2e/yurt/hubleader.go
tnsimon: This needs to be fixed in the hub leader config controller. Let's take that as a new bugfix issue?

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
```diff
@@ -161,6 +158,12 @@ func (r *ReconcileHubLeader) Reconcile(ctx context.Context, request reconcile.Re
 }
 
 func (r *ReconcileHubLeader) reconcileHubLeader(ctx context.Context, nodepool *appsv1beta2.NodePool) error {
+	if !nodepool.Spec.EnablePoolScopeMetadata {
+		// If the NodePool doesn't have pool scope metadata enabled, it should drop leaders
+		nodepool.Status.LeaderEndpoints = nil
+		return r.Status().Update(ctx, nodepool)
```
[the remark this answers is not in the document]
tnsimon: Cannot add this scenario. This controller is responsible for electing leaders when the nodepool is nil. I.e. when the nodepool has just been created and doesn't have leaders elected yet.

test/e2e/yurt/hubleader.go
rambohe-ch: ok, no problem.

pkg/yurtmanager/controller/hubleader/hubleader_controller.go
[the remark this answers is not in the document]
rambohe-ch: I mean there is no need to update nodepool if nodepool.Status.LeaderEndpoints = nil.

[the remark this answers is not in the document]
rambohe-ch: this means the code can be improved as following:
```
if !nodepool.Spec.EnablePoolScopeMetadata {
     // If the NodePool doesn't have pool scope metadata enabled, it should drop leaders
     if nodepool.Status.LeaderEndpoints != nil {
	nodepool.Status.LeaderEndpoints = nil
	return r.Status().Update(ctx, nodepool)
     }
}

```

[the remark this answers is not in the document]
rambohe-ch: sorry,  need to add a return as following:

```
if !nodepool.Spec.EnablePoolScopeMetadata {
     // If the NodePool doesn't have pool scope metadata enabled, it should drop leaders
     if nodepool.Status.LeaderEndpoints != nil {
	nodepool.Status.LeaderEndpoints = nil
	return r.Status().Update(ctx, nodepool)
     }
     return nil
}
```

tnsimon:
@rambohe-ch found the issue after lots of debugging. Whilst looking into this failure, I wonder if we should be using `r.Status().Patch()` instead of `r.Status().Update()`? If multiple controllers are updating status a call to Update could override the other potentially where as Patch only updates the changed fields?

For reference, cannot use `gomega.Eventually(somefunc(args))...` This will fetch the same answer until timeout is exceeded. Instead need to pass `gomega.Eventually(somefunc).WithArguments(...)`. I switched to not pass arguments to simplify. 

https://pkg.go.dev/github.com/onsi/gomega#Eventually

rambohe-ch (approved):

Issue #2323, opened 2025-02-14
[BUG] Hub Leader config is not cleaned up when node pool is deleted


tnsimon:
**What happened**:
Hub Leader config isn't deleted when the nodepool is deleted. 

**What you expected to happen**:
Hub leader config map should be deleted when nodepool removed.

**How to reproduce it (as minimally and precisely as possible)**:
* Delete the nodepool (without any nodes)
* Check configmap/leader-hub-{nodepool} is removed

**Anything else we need to know?**:

**Environment**:
- OpenYurt version: master
- Kubernetes version (use `kubectl version`): 1.30.6
- OS (e.g: `cat /etc/os-release`): any
- Kernel (e.g. `uname -a`): any
- Install tools: N/A
- Others:

**others**

/kind bug

PR #2324, opened 2025-02-14
fix: ensure hub leader configmap is deleted with nodepool

tnsimon:
#### What type of PR is this?
/kind bug

Hub leader config should be deleted when nodepool is deleted

Fixes https://github.com/openyurtio/openyurt/issues/2323

/assign @rambohe-ch 

```release-note
None
```

#### other Note
* Removed set enableLeaderElection = false from defaulting webhook. Should not set this default. 
* Created e2e scenario for hub leader config clean up

### DIFF
--- a/pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
+++ b/pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
@@ -82,7 +82,8 @@ func Add(ctx context.Context, cfg *appconfig.CompletedConfig, mgr manager.Manage
 			return ok
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return true
+			_, ok := e.Object.(*appsv1beta2.NodePool)
+			return ok
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			oldPool, ok := e.ObjectOld.(*appsv1beta2.NodePool)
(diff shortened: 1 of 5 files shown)


test/e2e/yurt/hubleader.go
```diff
@@ -254,3 +238,108 @@ var _ = Describe("Test hubleader elections", Serial, func() {
...
+			func() error {
+				err := k8sClient.Get(
+					ctx,
+					client.ObjectKey{Name: "leader-hub-" + npName, Namespace: metav1.NamespaceSystem},
+					&v1.ConfigMap{},
+				)
+				if err != nil && client.IgnoreNotFound(err) != nil {
+					return err
+				}
+				return nil
```
rambohe-ch, line 310: 1. both `configmap exists` and ` configmap doesn't exist`, nil will be returned.
2. return nil only when configmap doesn't exist, so maybe we need to improve as following:

```
if err != nil && client.IgnoreNotFound(err) == nil {
	return nil
}
return fmt.Errorf("unexpected error or configmap(%s) still exists", "leader-hub"+npName)
```

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
```diff
@@ -82,7 +82,8 @@ func Add(ctx context.Context, cfg *appconfig.CompletedConfig, mgr manager.Manage
 			return ok
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			return true
+			_, ok := e.Object.(*appsv1beta2.NodePool)
+			return ok
```
rambohe-ch, line 86: we can skip DeleteFunc directly as following:

```
DeleteFunc: func(e event.DeleteEvent) bool {
	return false
}
```

tnsimon:
@rambohe-ch fixed by using errors.IsNotFound( )

test/e2e/yurt/hubleader.go
tnsimon: @rambohe-ch fixed by using errors.IsNotFound( )

pkg/yurtmanager/controller/hubleaderconfig/hubleaderconfig_controller.go
tnsimon: Fixed

rambohe-ch: It seems unchanged.

tnsimon: Fixed lost it in the rebase when resolving the other comment

rambohe-ch (approved):

PR #2387, opened 2025-04-24
update kubernetes vendor to 1.32

zyjhtangtang:

upgrade kubernetes vendor to 1.32

```release-note
none
```


pkg/yurthub/proxy/multiplexer/multiplexerproxy.go
```diff
@@ -49,6 +50,10 @@ type multiplexerProxy struct {
 	stop                       <-chan struct{}
 }
 
+func init() {
+	v1.AddToScheme(scheme.Scheme)
```
rambohe-ch, line 54: why should add `k8s.io/kubernetes/pkg/apis/core/v1` into scheme?

zyjhtangtang: Need to introduce the service conversion function. 
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/conversion.go#L496

rambohe-ch: how about add comments for this init() function and explain why we need to AddToScheme for scheme.Scheme?

rambohe-ch (approved):

Issue #2409, opened 2025-06-16
[BUG] OTA Controller doesn't has permission to patch pod status


luc99hen:
**What happened**:
OTA Controller doesn't has permission to patch pod status 

**What you expected to happen**:
OTA Controller has permission to patch pod status 

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- OpenYurt version: 1.6
- Kubernetes version (use `kubectl version`):
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

**others**

/kind bug

PR #2415, opened 2025-06-27
fix ota controller doesn't has permission to patch pod status

PersistentJZH:
fix ota controller doesn't has permission to patch pod status

Fixes https://github.com/openyurtio/openyurt/issues/2409

```release-note

```

### DIFF
--- a/charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
+++ b/charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
@@ -317,6 +317,13 @@ rules:
   - patch
   - update
 - apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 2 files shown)


charts/yurt-manager/templates/yurt-manager-auto-generated.yaml
```diff
@@ -316,6 +316,13 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - ""
+  resources:
+  - pods/status
+  verbs:
+  - patch
+  - update
```
vie-serendipity, line 325: This file is auto-generated. You should add related permission to controller file and re-generate this file.

PR #2423, opened 2025-07-28
Optimizing hub cache

zyjhtangtang:

When the response data is gzipped and has a Content-Length header, writing to the client will fail.

-->
```release-note
None
```


rambohe-ch:
@zyjhtangtang  please fix the e2e error.

PR #2157, opened 2024-09-30
Enhance operational efficiency of K8s cluster in user's IDC

huangchenzhao:
#### What this PR does / why we need it:
#2124
#### Which issue(s) this PR fixes:

Fixes #2156

PR #2452, opened 2025-09-16
optimizing yurthub local mode

huangchenzhao:
add listOptions in EndpointsInformer

before: list-watch all endpoints in kube-public namespaces, other irrelavant endpoints ip address will be writen in iptables and results in bugs for loadbalancing

after: only list-watch tenant-apisever endpoints in kube-public namespaces

Fixes #2157


zyjhtangtang (commented):
Insufficient unit test (UT) coverage.

PR #2481, opened 2025-12-10
 Fix the issue where the masterservice and serviceenvupdater  modified the multiplexer cache.

zyjhtangtang:

There is an abnormal IPVS rule on the node for the Kubernetes service, as follows:

TCP  192.168.0.1:10268 rr
  -> 169.254.2.1:10268            Masq    1      0          0

The correct rule should be:
TCP  192.168.0.1:443 rr
  -> 169.254.2.1:10268            Masq    1      0          0

-->
```release-note
None
```


rambohe-ch:
@zyjhtangtang  please fix the unit test errors.

PR #2483, opened 2025-12-10
update yurttunnel certifications for ut.

zyjhtangtang:
> /kind ut

update yurttunnel certifications for ut.

-->
```release-note
None
```


singaporelovegaza-ship-it (commented):
love shaker

PR #2490, opened 2025-12-31
fix: failed to get nodepool with bootstrap token

will4j:
#### What type of PR is this?
/kind bug

As `--nodepool-name` becomes a Required arguments, executing cmd `yurtadm join --nodepool-name=xxx` will failed with error msg `when --nodepool-name is specified, the specified nodePool should be exist.`, but the actual cause is `nodepools.apps.openyurt.io "xxx" is forbidden: User "system:bootstrap:xxxxx" cannot get resource "nodepools" in API group "apps.openyurt.io" at the cluster scope`.

This pr:
1. displays err of GetNodePoolInfoWithRetry
2. add get permission of nodepools to clusterrole `yurt-hub-yurt-static-set-role` which binding to serviceaccount group`system:bootstrappers`

Fixes #

```release-note

```

### DIFF
--- a/charts/yurthub/templates/yurthub-cfg.yaml
+++ b/charts/yurthub/templates/yurthub-cfg.yaml
@@ -82,6 +82,12 @@ rules:
       - yurt-static-set-yurt-hub-cloud
     verbs:
       - get
+  - apiGroups:
+      - apps.openyurt.io
+    resources:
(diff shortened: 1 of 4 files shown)


zyjhtangtang:
@will4j Please pay attention to the bottlenecks in UT

PR #2500, opened 2026-01-20
 Fix: Prevent Nil Pointer Panic When Node Zone Label Changes During Processing

Yashika0724:
### Summary

This PR fixes a nil pointer panic in the Node Lifecycle Controller that can occur when a node’s
zone label changes between initial classification and retry-based reconciliation.

During retries, the controller may re-fetch the Node object from the API server with updated
topology labels, but the new zone may not be registered in the eviction queues. This results in
accessing a nil entry in `zoneNoExecuteTainter`, which causes the controller to panic and crash.

This fix adds defensive handling to safely register new zones when encountered and avoids invalid
map access.

---

###   Root Cause

Zones are registered only during initial node classification via `addPodEvictorForNewZone`.
However, in `updateNodeFunc`, the controller workflow can be:

1. Node is copied from the initial node list
2. Health update fails and retry is triggered
3. Node is re-fetched from the API server
4. Node’s zone label has changed between list and retry

In this case, the new zone is not present in `zoneNoExecuteTainter`.

When `markNodeForTainting` or `markNodeAsReachable` executes:

`go
nc.zoneNoExecuteTainter[nodetopology.GetZoneKey(node)]`

the map entry can be nil, leading to:

panic: runtime error: invalid memory address or nil pointer dereference

which crashes the entire Node Lifecycle Controller.

---

### Steps to Reproduce
	1.	Deploy OpenYurt with multiple zones (e.g. topology.kubernetes.io/zone=zone-a)
	2.	Ensure Node Lifecycle Controller is running
	3.	Introduce API latency or network instability to trigger retry loops
	4.	While a node is being processed, update its zone label:
                  kubectl label node <node-name> topology.kubernetes.io/zone=zone-new --overwrite
	5.	Controller panics when accessing eviction queue for the new zone

Alternative reproduction:
	•	Use a mutating webhook that modifies node zone labels
	•	Create node churn to increase reconciliation retries

---

### Fix Applied

markNodeForTainting
	•	Extract zone key once at function start
	•	Check if the zone is registered
	•	Dynamically register missing zones using the same logic as initial classification
	•	Safely access eviction queues after registration

markNodeAsReachable
	•	Check if zone exists before removing node from eviction queue
	•	If zone is not registered, return safely with no-op

This prevents nil map access while preserving existing taint and eviction behavior.

---

###  Impact
	•	Prevents controller-wide crashes
	•	Improves resilience when node labels change dynamically
	•	Safe handling of retries during reconciliation
	•	No behavioral change to taint or eviction logic

###  Testing
	•	Code compiles locally
	•	Follows existing zone initialization patterns
	•	CI will validate controller integration behavior

### DIFF
--- a/pkg/yurtmanager/controller/nodelifecycle/node_life_cycle_controller.go
+++ b/pkg/yurtmanager/controller/nodelifecycle/node_life_cycle_controller.go
@@ -1204,19 +1204,32 @@ func (nc *ReconcileNodeLifeCycle) addPodEvictorForNewZone(node *v1.Node) {
 func (nc *ReconcileNodeLifeCycle) markNodeForTainting(node *v1.Node, status v1.ConditionStatus) bool {
 	nc.evictorLock.Lock()
 	defer nc.evictorLock.Unlock()
+
+	zone := nodetopology.GetZoneKey(node)
+	// Ensure the zone exists in the map. This can happen if a node's zone label
+	// changed after initial classification but before processing completes.
+	if _, found := nc.zoneStates[zone]; !found {
+		nc.zoneStates[zone] = stateInitial
+		nc.zoneNoExecuteTainter[zone] =
+			scheduler.NewRateLimitedTimedQueue(
+				flowcontrol.NewTokenBucketRateLimiter(nc.evictionLimiterQPS, scheduler.EvictionRateLimiterBurst))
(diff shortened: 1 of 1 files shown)


Yashika0724:
Hi @zyjhtangtang , this PR fixes a controller panic caused by zone label changes during retries by
safely handling missing zone registrations. I’d appreciate a review and any feedback. Thanks!

PR #2501, opened 2026-01-21
 Fix: Prevent PodBinding Controller Panic for Infinite NoExecute Tolerations

WHOIM1205:
### Summary
This PR fixes a **nil pointer dereference** in the PodBinding controller that causes `yurt-manager` to **panic and restart** when reconciling pods with **infinite (`TolerationSeconds=nil`) `NoExecute` tolerations** on autonomous edge nodes.

Infinite `NoExecute` tolerations are **valid and common** in Kubernetes. The controller previously assumed `TolerationSeconds` was always set and dereferenced it without a nil check.

---

### ❗ Problem
In `reconcilePod`, the controller stored original toleration values by directly dereferencing `TolerationSeconds`.

When `TolerationSeconds == nil` (meaning *tolerate forever*), this resulted in a **runtime panic**:

This breaks **node autonomy**, a core OpenYurt feature, and can destabilize the entire control plane.

---

### 🔁 Steps to Reproduce
1. Deploy OpenYurt with `yurt-manager` running.
2. Mark an edge node as autonomous:
   ```bash
   kubectl annotate node edge-node-1 node.openyurt.io/autonomy=true
apiVersion: v1
kind: Pod
metadata:
  name: test-pod
spec:
  nodeName: edge-node-1
  tolerations:
  - key: "node.kubernetes.io/not-ready"
    operator: "Exists"
    effect: "NoExecute"
    # TolerationSeconds intentionally omitted (nil)
  containers:
  - name: test
    image: nginx
kubectl logs -n kube-system deployment/yurt-manager | grep panic

### DIFF
--- a/pkg/yurtmanager/controller/yurtcoordinator/podbinding/pod_binding_controller.go
+++ b/pkg/yurtmanager/controller/yurtcoordinator/podbinding/pod_binding_controller.go
@@ -48,6 +48,11 @@ import (
 const (
 	originalNotReadyTolerationDurationAnnotation    = "apps.openyurt.io/original-not-ready-toleration-duration"
 	originalUnreachableTolerationDurationAnnotation = "apps.openyurt.io/original-unreachable-toleration-duration"
+
+	// infiniteTolerationValue is the annotation value used to indicate that the original
+	// TolerationSeconds was nil (meaning infinite/no timeout). This allows proper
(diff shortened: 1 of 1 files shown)


zyjhtangtang:
@WHOIM1205  Please focus on unit test coverage.

PR #2503, opened 2026-01-21
Fix: prevent panic in YurtAppSet status for StatefulSet workloads

Yashika0724:
### Summary

This PR fixes a runtime panic in the YurtAppSet controller that occurs when calculating status
for YurtAppSets configured with a StatefulSet workload template.

The controller previously assumed all workloads were Deployments and performed an unsafe type
assertion, which caused a panic when workloads were actually StatefulSets. As a result,
StatefulSet-based YurtAppSets would crash the controller during reconciliation.

This fix updates the status calculation logic to safely handle both Deployment and StatefulSet
workloads.

---

### Root Cause

In `conciliateYurtAppSetStatus`, workloads are iterated as `[]metav1.Object` and cast like this:

`go
workloadObj := workload.(*appsv1.Deployment)`

However, the workload manager returns different types depending on the workload template:
	•	Deployment template → *appsv1.Deployment
	•	StatefulSet template → *appsv1.StatefulSet

When a YurtAppSet uses statefulSetTemplate, the controller attempts to cast a
*appsv1.StatefulSet to *appsv1.Deployment, resulting in:

interface conversion: metav1.Object is *v1.StatefulSet, not *v1.Deployment

This causes the YurtAppSet controller to crash during reconciliation and enter a crash loop.

⸻

### Steps to Reproduce
	1.	Deploy OpenYurt and ensure the YurtAppSet controller is running
	2.	Create a YurtAppSet that uses a StatefulSet workload template
	3.	Create a NodePool with at least one node so workloads get created
	4.	Wait for the controller to reconcile and calculate YurtAppSet status
	5.	The controller panics while processing workload status and enters a crash loop
---

### Fix Applied

The unsafe type assertion was replaced with a type switch that supports both workload types:
	•	*appsv1.Deployment
	•	*appsv1.StatefulSet

For each workload, the controller now correctly extracts:
	•	replicas
	•	ready replicas
	•	updated replicas
	•	workload revision hash

Unknown workload types are skipped safely with a warning log instead of crashing the controller.

This preserves existing behavior for Deployments and correctly enables StatefulSet support.

---

### Impact
	•	Prevents YurtAppSet controller crashes
	•	Enables StatefulSet-based YurtAppSets as documented in the API
	•	Correct status reporting for all supported workload types
	•	No behavior change for Deployment-based YurtAppSets
	•	Improves reliability for stateful edge workloads

---
### Testing
	•	Logic verified for both Deployment and StatefulSet workload managers
	•	No API or behavior changes introduced
	•	CI will validate controller integration behavior

---

### Important: This + Your Previous Fix = STRONG Profile

You now have:

1. Node lifecycle controller crash fix (zone label race)
2. YurtAppSet controller crash fix (StatefulSet status)

These are **core controller reliability fixes** — not docs, not trivial bugs.

For CNCF LFX Mentorship, this is **exactly the kind of contribution mentors look for**.

---

If you want, next I can help you with:

-  PR review request comment (for this PR)
-  Message to maintainer mentioning your interest in contributing
-  How to mention these PRs in your LFX application answers
-  What to pick as your next OpenYurt issue

### DIFF
--- a/pkg/yurtmanager/controller/yurtappset/yurt_app_set_controller.go
+++ b/pkg/yurtmanager/controller/yurtappset/yurt_app_set_controller.go
@@ -628,12 +628,29 @@ func (r *ReconcileYurtAppSet) conciliateYurtAppSetStatus(
 	// calculate yas current status
 	readyWorkloads, updatedWorkloads := 0, 0
 	for _, workload := range curWorkloads {
-		workloadObj := workload.(*appsv1.Deployment)
-		if workloadObj.Status.ReadyReplicas == workloadObj.Status.Replicas {
+		var readyReplicas, replicas, updatedReplicas int32
+		var hash string
+
+		switch w := workload.(type) {
+		case *appsv1.Deployment:
+			readyReplicas = w.Status.ReadyReplicas
+			replicas = w.Status.Replicas
+			updatedReplicas = w.Status.UpdatedReplicas
+			hash = workloadmanager.GetWorkloadHash(w)
+		case *appsv1.StatefulSet:
+			readyReplicas = w.Status.ReadyReplicas
+			replicas = w.Status.Replicas
+			updatedReplicas = w.Status.UpdatedReplicas
+			hash = workloadmanager.GetWorkloadHash(w)
+		default:
+			klog.Warningf("YurtAppSet[%s/%s] unknown workload type: %T", yas.GetNamespace(), yas.GetName(), workload)
+			continue
+		}
+
+		if readyReplicas == replicas {
 			readyWorkloads++
 		}
-		if workloadmanager.GetWorkloadHash(workloadObj) == expectedRevision.GetName() &&
-			workloadObj.Status.UpdatedReplicas == workloadObj.Status.Replicas {
+		if hash == expectedRevision.GetName() && updatedReplicas == replicas {
 			updatedWorkloads++
 		}
 	}


pkg/yurtmanager/controller/yurtappset/yurt_app_set_controller.go
```diff
@@ -628,12 +628,29 @@ func (r *ReconcileYurtAppSet) conciliateYurtAppSetStatus(
...
+			replicas = w.Status.Replicas
+			updatedReplicas = w.Status.UpdatedReplicas
+			hash = workloadmanager.GetWorkloadHash(w)
+		case *appsv1.StatefulSet:
+			readyReplicas = w.Status.ReadyReplicas
+			replicas = w.Status.Replicas
+			updatedReplicas = w.Status.UpdatedReplicas
+			hash = workloadmanager.GetWorkloadHash(w)
+		default:
+			klog.Warningf("YurtAppSet[%s/%s] unknown workload type: %T", yas.GetNamespace(), yas.GetName(), workload)
```
vie-serendipity, line 646: Would it be better to include the workload's name and namespace in the logs?

Yashika0724: > Would it be better to include the workload's name and namespace in the logs?

That makes sense, including the workload name and namespace would improve debugging and log clarity. I’ll update the warning log to include those fields. 
Thanks for the suggestion!

vie-serendipity: Can you modify the log format so this can be merged in?

PR #2508, opened 2026-01-26
fix: race condition in cache manager's inMemoryCache

shiavm006:
## Description

Fixed critical race condition in cache manager's `inMemoryCache` map that could lead to data corruption, panics, and inconsistent cache state.

## Problem

The `inMemoryCache` map was accessed concurrently without proper synchronization:
1. `saveWatchObject` runs in a goroutine (via `CacheResponse`) and calls `updateInMemoryCache`
2. `queryInMemoryCache` returns pointers directly from the map after releasing the lock
3. Objects stored in cache could be modified concurrently by multiple goroutines

## Solution
- **Deep copy on write**: Objects are deep copied before storing in `inMemoryCacheFor` to ensure cached objects are independent
- **Deep copy on read**: Objects are deep copied while holding the RLock in `queryInMemoryCache` to prevent map entry replacement during copy
- **Nil safety**: Added defensive nil checks to handle edge cases gracefully

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -698,8 +698,8 @@ func (cm *cacheManager) inMemoryCacheFor(key string, obj runtime.Object) {
 	// This prevents race conditions where the original object might be modified
 	// by other goroutines after being stored in the cache.
 	if obj == nil {
-		// If obj is nil, store nil (should not happen in practice, but handle gracefully)
-		cm.inMemoryCache[key] = nil
+		// If obj is nil, don't store anything - queryInMemoryCache should return
+		// ErrInMemoryCacheMiss for non-existent keys, not return a nil object.
 		return
 	}
 	cm.inMemoryCache[key] = obj.DeepCopyObject()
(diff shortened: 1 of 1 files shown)


pkg/yurthub/cachemanager/cache_manager.go
```diff
@@ -694,7 +694,15 @@ func (cm *cacheManager) storeObjectWithKey(key storage.Key, obj runtime.Object)
 func (cm *cacheManager) inMemoryCacheFor(key string, obj runtime.Object) {
 	cm.Lock()
 	defer cm.Unlock()
-	cm.inMemoryCache[key] = obj
+	// Deep copy before storing to ensure cached objects are independent.
+	// This prevents race conditions where the original object might be modified
+	// by other goroutines after being stored in the cache.
+	if obj == nil {
+		// If obj is nil, store nil (should not happen in practice, but handle gracefully)
+		cm.inMemoryCache[key] = nil
```
rambohe-ch: If obj is nil, we should not store `nil` for the key. because we should return `ErrInMemoryCacheMiss` when   
object doesn't exist. how about return directly?

shiavm006: make sense done that

PR #2509, opened 2026-01-28
avoid panic on pod without owner refs

shiavm006:
- Prevents a panic in pkg/yurthub/otaupdate/ota.go when UpdatePod is called on a Pod with an empty ownerReferences list.
- Adds a guard (len(ownerRefs) == 0) and returns 400 BadRequest with message "Pod has no owner references" instead of  crashing.
- Adds a unit test covering the no-ownerRefs case in pkg/yurthub/otaupdate/ota_test.go.


zyjhtangtang (approved):

shiavm006:
hey @zyjhtangtang r u active on the slack channel ?

PR #2510, opened 2026-01-29
protect yurthub transport client map from races

shiavm006:
- Add sync.RWMutex to transportAndClientManager and fakeTransportManager to guard access to serverToClientset.
- Build serverToClientset maps completely before publishing them on the manager structs to avoid concurrent writes.
- Change ListDirectClientset in both managers to return a shallow copy of the internal map, preventing external callers from mutating internal state.
- Keep the existing Interface API unchanged while ensuring safe concurrent reads and better encapsulation.

### DIFF
--- a/pkg/yurthub/transport/fake_transport.go
+++ b/pkg/yurthub/transport/fake_transport.go
@@ -19,6 +19,7 @@ package transport
 import (
 	"net/http"
 	"net/url"
+	"sync"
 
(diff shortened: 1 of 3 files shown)


shiavm006:
@zyjhtangtang would u like to look in to it

shiavm006:
@zyjhtangtang any thoughts on this

shiavm006:
@zyjhtangtang ??

PR #2511, opened 2026-02-01
Fix/kubelet idempotency

Aman-Cool:
# Make kubelet redirect/revert idempotent

## Summary
Fixes unnecessary kubelet restarts during YurtHub convert/revert by making the operations truly idempotent.  
Kubelet is now restarted **only when configuration or environment actually changes**.

## Root Cause
Kubelet config helpers did not indicate whether they modified state, so convert/revert paths always restarted kubelet—even when the node was already correctly configured.

## Fix
- Return change status from kubelet config helpers
- Restart kubelet only on real config/env changes
- Add tests to verify no-op vs change behavior

## Impact
- Eliminates redundant kubelet restarts
- Safe for reconciliation loops
- Enables reliable label-driven YurtHub automation

## Backward Compatibility
No breaking changes. Behavior is unchanged except redundant restarts are avoided.

### DIFF
--- a/pkg/node-servant/components/kubelet.go
+++ b/pkg/node-servant/components/kubelet.go
@@ -51,32 +51,45 @@ func NewKubeletOperator(openyurtDir string) *kubeletOperator {
 
 // RedirectTrafficToYurtHub
 // add env config leads kubelet to visit yurtHub as apiServer
+// This function is idempotent: if kubelet is already redirected, it will not restart kubelet.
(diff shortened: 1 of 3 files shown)


Aman-Cool:
@zhangzhenyuyu ,quick ping — this PR is ready for review and CI is passing.
Please let me know if you’re the right person to review or if I should loop someone else in. Thanks!

PR #2514, opened 2026-02-01
Fix/yurthub uninstall idempotency

Aman-Cool:
## Summary

Fixes a broken idempotency bug in `UnInstall()` where an early return skips YurtHub process shutdown and cache cleanup when the config directory is missing.

## Problem

If `/var/lib/yurthub` does not exist (common after partial or interrupted uninstalls), `UnInstall()` returns success early and leaves `/etc/kubernetes/cache/` intact. On reinstall, YurtHub reuses stale cached data.

## Fix

Remove the early return.  
Skip config deletion if missing, but **always** wait for YurtHub to exit and clean the cache directory.

## Impact

- Restores true idempotency  
- Ensures uninstall → install cycles converge to a clean state  
- Prevents stale cache reuse on edge nodes  
- Unblocks label-driven YurtHub automation  

Small, safe change (~5 lines) with no behavior regression.


Aman-Cool:
@zyjhtangtang  Hi, just a gentle follow-up. I’ve addressed the coverage issues and CI is green now.  
Whenever it’s convenient, I’d appreciate you taking another look. Thanks for your time.

PR #2517, opened 2026-02-05
fix: guard nil request info in autonomy proxy

shiavm006:
## Summary
- validate RequestInfo in autonomy proxy before using node name to avoid panic on nil
- reuse node name after validation and return clear error when request info is missing


zyjhtangtang (approved):

shiavm006:
hey r u on slack @zyjhtangtang

PR #2589, opened 2026-07-17
Fix dead reflect.DeepEqual in ensureConvertZoneLabel (#2588)

### LINKED ISSUE #2588: bug: Dead reflect.DeepEqual guard in ensureConvertZoneLabel always false
### Description

The reflect.DeepEqual check at line 435 is unreachable dead code. The only code path that reaches line 435 is when hasZone is false (line 421 returned early when hasZone is true). But line 433 just set updated.Labels[zoneLabel] = nodePoolName, so node.Labels (without zoneLabel) can never equal updated.Labels (with zoneLabel). The return nil at line 436 is never executed.

### Steps to reproduce

1. Open `pkg/yurtmanager/controller/yurtnodeconversion/yurt_node_conversion_controller.go` around line 435
2. The redundant DeepEqual check always evaluates to false and the early return is never taken. The Patch call is always reached — which is the correct behavior — but the dead guard causes confusion about when the Patch is skipped. If a future refactor changes the early-return logic before line 435 without noticing this guard is dead, it could create logic errors.

### Expected behavior

The pattern was copy-pasted from ensureEdgeWorkerLabel (line 391) where DeepEqual is meaningful (the label might already have the same value). In ensureConvertZoneLabel the code guarantees !hasZone before reaching the check, so the label is always newly added and the maps can never be equal. should be handled correctly.

### Actual behavior

The redundant DeepEqual check always evaluates to false and the early return is never taken. The Patch call is always reached — which is the correct behavior — but the dead guard causes confusion about when the Patch is skipped. If a future refactor changes the early-return logic before line 435 without noticing this guard is dead, it could create logic errors.

UtkarshWoc:
What type of PR is this?
/kind bug

The `reflect.DeepEqual` check in `ensureConvertZoneLabel` was unreachable dead code. The only code path that reaches this check is when `hasZone` is false, meaning the label is always newly added when reaching that point. 

Because `node.Labels` (without the `zoneLabel`) can never equal `updated.Labels` (with the newly added `zoneLabel`), the equality check always evaluated to false and the early return was never executed.

This PR removes the dead check to avoid confusion about when the Patch is skipped, preventing potential logic errors in future refactors.

Which issue(s) this PR fixes:
Fixes #2588

Special notes for your reviewer:
@deepak0x

 Does this PR introduce a user-facing change?
```release-note
NONE

### DIFF
--- a/pkg/yurtmanager/controller/yurtnodeconversion/yurt_node_conversion_controller.go
+++ b/pkg/yurtmanager/controller/yurtnodeconversion/yurt_node_conversion_controller.go
@@ -432,9 +432,6 @@ func (r *ReconcileYurtNodeConversion) ensureConvertZoneLabel(ctx context.Context
 	}
 	updated.Labels[zoneLabel] = nodePoolName
 
-	if reflect.DeepEqual(node.Labels, updated.Labels) {
-		return nil
-	}
 	klog.V(4).Info(Format("patch node(%s) %s for native trafficDistribution compatibility", node.Name, zoneLabel))
 	return r.Patch(ctx, updated, client.MergeFrom(node))
 }


deepak0x:
Hey @UtkarshWoc 
Taking anyone's issue without asking is not good practice 🙂

UtkarshWoc:
Ohh, My bad deepak! I'll close this issue right now!

Issue #2559, opened 2026-06-24
[feature request] Add Claude Code Skills for deploying OpenYurt and configuring Raven


Jonny-ee:
## Background

[Claude Code](https://claude.ai/code) is an AI coding assistant that supports **Skills** — slash commands defined as Markdown files in `.claude/commands/`. When a user types `/openyurt-deploy` inside the OpenYurt repository, Claude reads the skill file and executes the described steps interactively on behalf of the user.

Adding Skills to the OpenYurt repository would lower the barrier for operators who want to deploy OpenYurt or configure Raven on an existing Kubernetes cluster, without needing to read through the full documentation manually.

### `/openyurt-deploy` — Deploy OpenYurt on an existing Kubernetes cluster

An interactive skill that:
1. Checks prerequisites (kubectl, Helm ≥ v3, Kubernetes ≥ 1.24)
2. Installs `yurt-manager` via Helm
3. Creates an Edge NodePool CR
4. Converts nodes to edge nodes using the label-driven `YurtNodeConversionController` mechanism (labels node with `apps.openyurt.io/nodepool=<name>`, waits for node-servant Job to complete)
5. Enables edge autonomy per node (`node.openyurt.io/autonomy-duration`)
6. Verifies the `Autonomy` condition on each edge node

### `/openyurt-raven` — Configure Raven for cross-region networking

An interactive skill that:
1. Verifies OpenYurt is already deployed
2. Installs `raven-agent` DaemonSet via Helm (from the `openyurtio/raven` repository)
3. Generates a VPN PSK with `openssl rand -hex 64`
4. Creates Gateway CRs for each NodePool (`raven.openyurt.io/v1beta1`)
5. Supports both `PublicIP` and `LoadBalancer` expose types
6. Supports enabling/disabling L3 tunnel and L7 proxy independently
7. Verifies cross-NodePool connectivity

## Implementation

The implementation adds `.claude/commands/` to the repository root:

```
.claude/commands/
├── openyurt-deploy.md
├── openyurt-raven.md
└── README.md
```

These are static Markdown files — no build changes, no code changes. The skills are picked up automatically by Claude Code when working inside the repository.

## Why This Is Useful

- **Lower onboarding friction**: operators can deploy OpenYurt end-to-end with a single slash command, with Claude executing and verifying each step
- **Accurate and versioned**: skills live in the repo alongside the code, so they can be updated with each release
[...]

PR #2592, opened 2026-07-17
feat: Architect modular Claude Code Skills for Edge Deployment

Aryanbhargava18:
#### What type of PR is this?
/kind feature
/sig ai

This PR establishes the foundation for autonomous AI operations on OpenYurt edge nodes by introducing a modular Claude Code Skill architecture. It implements two critical, production-ready skills in the `.claude/skills/` directory:
1. **Raven Topology Skill:** Enables the AI to autonomously debug cross-region L3/L7 routing via Gateway and IPSec tunnel inspection.
2. **YurtHub Autonomy Skill:** Enables the AI to verify edge node autonomy annotations and local cache integrity during cloud disconnections.

These skills empower the AI to resolve OpenYurt's most complex edge-cloud networking and autonomy challenges without manual intervention.

Part of #2559

update: stripped the remaining ssh dependencies from allowed-tools based on community feedback. the execution model is now 100% k8s-native (ephemeral jobs/daemonsets) which completely seals the edge security boundary without needing host shell access.

leaving this stable for now, no rush on review until maintainers have bandwidth.

```release-note
Introduces modular Claude Code Skills (Raven Topology and YurtHub Autonomy) to enable autonomous AI-driven debugging and operations on edge nodes.
```


Aryanbhargava18:
Hey @rambohe-ch @zyjhtangtang @luc99hen! I saw Issue #2559 on the CNCF Mentoring board. Since I heavily utilize Markdown-based AI coding assistants in my day-to-day Kubernetes workflows, I wanted to proactively lay out the foundational architecture for the OpenYurt skills before the application period officially opens.

I read through the issue discussion thread and noted the team's strong preference for a modular, maintainable folder structure (keeping the entrypoint short and moving detailed logic to reference files). I've implemented exactly that architecture here for both the deploy and raven skills.

**Key Technical Implementations in this Draft:**
- **Modular Architecture**: Split `/openyurt-deploy` and `/openyurt-raven` into `SKILL.md` (frontmatter restrictions) and a `references/` directory containing the flow, troubleshooting, and uninstall decision trees.
- **crictl Diagnostic Loop**: Noticed the hard dependency on crictl for the `node-servant` Job. I built an explicit diagnostic branch in the troubleshooting reference to catch this failure and instruct the AI how to resolve it dynamically without halting the user.
- **Stable Verification**: Bypassed the non-existent Autonomy condition in favor of instructing the AI to poll the `isConvertedStable()` requirements directly via node labels and the `YurtConversionFailed` condition.

I opened this as a Draft PR so we can iterate on the structure. Let me know your thoughts on the modular references approach!

.claude/commands/openyurt-deploy/references/openyurt-deploy-flow.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Phase 3: Node Conversion (Label-Driven)
+1. **Create NodePool:** Ask the user for a NodePool name. Create and apply the `NodePool` CR (type: Edge).
+2. **Label Nodes:** 
+   `kubectl label node <node-name> apps.openyurt.io/nodepool=<NODEPOOL_NAME>`
+   `kubectl label node <node-name> apps.openyurt.io/desired-nodepool=<NODEPOOL_NAME>`
+3. **Verify Completion:** Poll `kubectl get node <node-name> -o json`. Wait for:
+   - `apps.openyurt.io/nodepool` label is present.
+   - `openyurt.io/is-edge-worker=true` label is present.
+   - `YurtConversionFailed` condition is absent or `False`.
```
ADITYA-CODE-SOURCE: I checked this flow against the current node conversion controller and noticed a possible correctness issue.

**Problem:** This step refers to `YurtConversionFailed`, but the current controller uses `YurtNodeConversionFailed`. Also, checking for the condition to be “absent or False” is too broad for success detection.

**Why it matters:** In the current implementation, `status=False` is used for both `Converting` and `Converted`, so this can report success while conversion is still in progress.

**Suggested fix:** Could we align this with the current controller contract and check the stable converted state used by `isConvertedStable()` instead? That path requires:
- `apps.openyurt.io/nodepool` present
- `openyurt.io/is-edge-worker=true`
- node uncordoned
- conversion condition `False` with reason `Converted`

**Source refs:**
- `pkg/yurtmanager/controller/yurtnodeconversion/yurt_node_conversion_controller.go:52`
- `pkg/yurtmanager/controller/yurtnodeconversion/yurt_node_conversion_controller.go:583-590`

For example, this may fail when the conversion job is still running, because the condition is already `False` during `Converting`. Could we handle that case here?

.claude/commands/openyurt-deploy/references/openyurt-deploy-flow.md
```diff
@@ -0,0 +1,25 @@
...
+4. **Enable Autonomy:** `kubectl label node <node-name> node.openyurt.io/autonomy-duration=5m`
```
ADITYA-CODE-SOURCE: I checked this flow against the current OpenYurt source and noticed a possible correctness issue.

**Problem:** `node.openyurt.io/autonomy-duration` is documented here as a node label, but the current implementation defines it as an annotation.

**Why it matters:** Using `kubectl label` here would not configure the key the way current OpenYurt expects, so the autonomy step may not behave as intended.

**Suggested fix:** Could we switch this to annotation semantics and update the wording/commands accordingly?

**Source refs:**
- `pkg/projectinfo/projectinfo.go:107-114`

For example, this may fail when a user follows the documented command exactly and expects autonomy to be configured through the current annotation-based path. Could we handle that case here?

ADITYA-CODE-SOURCE (commented):
Thanks for putting this together — the modular `SKILL.md` + `references/` split looks good.

I checked this against current source and noticed two correctness issues that seem worth aligning before merge:

1. The conversion condition is `YurtNodeConversionFailed`, and success can’t be checked as just “absent or False” because `Converting` is also `False`.
2. `node.openyurt.io/autonomy-duration` is currently an annotation, not a node label.

Optional follow-up: the Raven `Gateway` example may also need a pass against the current `v1beta1` schema.

Could we align these with current source behavior?

Aryanbhargava18:
Thanks for pointing those out @ADITYA-CODE-SOURCE. Good catches on the controller contract and the annotation semantics. 

I've pushed a commit addressing them:
1. **Conversion Controller Align:** Updated the deployment flow to check that the node is uncordoned and `YurtNodeConversionFailed` condition is `False` with reason `Converted`. 
2. **Autonomy Annotation:** Changed `kubectl label` to `kubectl annotate` to match `pkg/projectinfo/projectinfo.go`.
3. **Raven v1beta1 Schema:** Also updated the Gateway CR in the Raven flow to use the v1beta1 `exposeType`, `tunnelConfig`, and `proxyConfig` fields.

.claude/commands/openyurt-raven/references/openyurt-raven-flow.md
```diff
@@ -0,0 +1,44 @@
...
+1. Ask the user for the expose type (`PublicIP` or `LoadBalancer`).
+2. Create the Gateway CR:
+   ```yaml
+   apiVersion: raven.openyurt.io/v1beta1
+   kind: Gateway
+   metadata:
+     name: <NODEPOOL_NAME>-gw
+   spec:
+     nodeSelector:
+       nodeSelectorTerms:
```
ADITYA-CODE-SOURCE: I re-checked the updated Raven flow against the current `v1beta1` schema and noticed one remaining mismatch here.

**Problem:** `spec.nodeSelector` is still written using `nodeSelectorTerms`, but the current `GatewaySpec` uses a Kubernetes `LabelSelector` shape. Also, the current defaulting path sets `spec.nodeSelector` to `raven.openyurt.io/gateway=<gateway-name>` on create.

**Why it matters:** Even after the earlier `v1beta1` field updates, this example can still be misleading because the selector shape does not match the current API/defaulting behavior.

**Suggested fix:** Could we either rely on the defaulted selector behavior here, or rewrite this block using the current `LabelSelector` shape explicitly?

**Source refs:**
- `pkg/apis/raven/v1beta1/gateway_types.go:63-76`
- `pkg/apis/raven/v1beta1/default.go:27-32`

For example, this may confuse users who expect a NodePool-based `nodeSelectorTerms` block to survive admission unchanged. Could we align this example with current source behavior?

Aryanbhargava18:
Good catch @ADITYA-CODE-SOURCE. Since `v1beta1.SetDefaultsGateway` unconditionally sets `spec.nodeSelector` to `raven.openyurt.io/gateway: <gateway-name>`, relying on `nodeSelectorTerms` for `apps.openyurt.io/nodepool` would indeed be misleading as the mutating webhook would silently overwrite it.

I just pushed a commit that updates the `openyurt-raven-flow.md` to explicitly use the `LabelSelector` shape with `raven.openyurt.io/gateway` to match the defaulting behavior. I also added a step to proactively label the nodes in the nodepool with `raven.openyurt.io/gateway=<NODEPOOL_NAME>-gw` so they successfully join the gateway.

Thanks again for reviewing!

Aryanbhargava18:
update: went through the latest `node-servant` flow again and noticed that the `crictl` preflight check we added also handles the edge case where `containerd` socket path differs between kubeadm and k3s setups. the troubleshooting reference now covers both paths so the AI skill can recover dynamically regardless of the node runtime.

Aryanbhargava18:
update: pushed a commit that embeds autonomous execution scripts directly into the markdown to replace manual instructions.

- raven connectivity: `/openyurt-raven` now extracts pod IPs via JSONPath and executes cross-region `ping` validation to verify L3 tunnel health without human intervention.
- diagnostics: converted `/openyurt-deploy` troubleshooting into an executable script. the agent now parses failed conversion jobs and catches CRI path errors (`crictl`) directly from pod logs.

leaving this here for review when bandwidth allows.

.claude/skills/openyurt-deploy/SKILL.md
```diff
@@ -0,0 +1,22 @@
+---
+description: Deploy OpenYurt onto an existing Kubernetes cluster and convert edge nodes.
+when_to_use: Use when the user wants to deploy OpenYurt, create NodePools, or convert existing worker nodes to edge nodes via the YurtNodeConversionController.
+disable-model-invocation: true
+allowed-tools: >
+  Bash(kubectl *)
+  Bash(helm *)
+  Bash(ssh *)
+  Read
+  Grep
```
Suyog241005, line 17: Allowing the AI assistant to execute arbitrary SSH commands (`Bash(ssh *)`) presents a significant security risk for operators running Claude Code, particularly on production or enterprise clusters. It grants the LLM unrestricted shell access to target hosts.

Can we remove the SSH dependency entirely by delegating host-level checks to a temporary Kubernetes `Job` or `Pod` that reports node status back to the API server?

.claude/skills/openyurt-deploy/references/openyurt-deploy-flow.md
```diff
@@ -0,0 +1,26 @@
+# OpenYurt Deployment Flow
+
+Follow these steps sequentially to deploy OpenYurt.
+
+## Phase 1: Preflight Validation
+1. **Kubernetes Version:** Run `kubectl version --short`. Verify the server version is >= 1.24.
+2. **Helm Version:** Run `helm version --short`. Verify it is v3 or higher.
+3. **crictl Check (CRITICAL):** Ask the user which nodes they intend to convert. Instruct them to verify `crictl` is in the PATH on those nodes via SSH. If missing, they must install it, or Phase 3 will fail silently.
```
Suyog241005: Requiring the operator to manually SSH into nodes to verify `crictl` is in the path defeats the purpose of an automated AI skill. 

A more cloud-native approach would be to deploy a temporary Kubernetes validation Job (or run a quick check within the `yurt-node-servant` pre-checks) and query its logs via `kubectl`. This keeps the workflow entirely within the Kubernetes API boundary and avoids host-access requirements.

.claude/skills/openyurt-raven/references/openyurt-raven-flow.md
```diff
@@ -0,0 +1,93 @@
...
+    echo "Failed to find two test pods. Ensure you have at least 2 nodes ready."
+else
+    echo "Pinging $POD2_IP from $POD1_NAME..."
+    kubectl exec $POD1_NAME -- ping -c 3 $POD2_IP
+    if [ $? -eq 0 ]; then
+        echo "Raven L3 tunnel connectivity verified successfully!"
+    else
+        echo "Raven L3 tunnel connectivity failed."
+    fi
+fi
```
Suyog241005, line 236: Writing temporary pod list data to `/tmp/raven_pods.txt` is fragile and can fail in environments with read-only root filesystems or restricted directory permissions.

Storing the pod names and IPs in memory using shell arrays/variables is safer, cleaner, and avoids disk writes entirely.

Suyog241005:
> I actually mapped these exact edge cases into the allowed-tools flow in https://github.com/openyurtio/openyurt/pull/2592 a few days ago. Claude now runs a preflight LookPath("crictl") check via SSH, and polls kubectl get node to verify the is-edge-worker=true and YurtConversionFailed=False conditions.

In typical edge computing environments (which OpenYurt targets), edge nodes are located in remote NodePools behind firewalls or NATs. This means the operator's workstation running Claude Code will not have direct SSH routing to the edge nodes. Additionally, passing SSH credentials to an AI assistant raises security/compliance concerns.

Instead of SSH, we can perform this preflight check in a **Kubernetes-native way**:
1. **Use a light validation DaemonSet/Job**: Run a temporary single-pod Job pinned to the target node using `nodeSelector`. The pod runs a simple script to check for `crictl` on the host (e.g., mounting `/usr/bin` or checking via a hostPath mount if appropriate) and reports back.
2. **Handle it inside YurtNodeConversionController**: The best place for this check is actually inside the node conversion job or controller itself, bubbling the failure up as a standard Kubernetes `Event` or `NodeCondition`, which Claude can then easily query via `kubectl`.

This keeps the security boundary strictly within `kubectl` and `helm`, eliminating the need for `ssh` in `allowed-tools`.

Suyog241005:
Hi @Aryanbhargava18,

It is great to see the pivot in Step 1.5 to use a Kubernetes-native preflight Job! 

However, looking at the rest of the updated files, the dependency on SSH is still prevalent in several critical sections, which means `Bash(ssh *)`, `Bash(systemctl *)`, and `Bash(journalctl *)` are still required in `SKILL.md`:

1. **Deploy Troubleshooting**: SSH commands are still used to install `cri-tools`, read `/etc/kubernetes/kubelet.conf`, and check containerd sockets.
2. **Deploy Uninstall**: SSH is used to manually clean up static pod manifests from the node filesystem.
3. **Raven Flow & Troubleshooting**: SSH is used to check for `libreswan` (`ipsec --version`) and install packages on the gateways.

If the user's workstation cannot SSH into edge nodes due to firewalls or NAT, all of these troubleshooting and uninstall steps will fail when Claude attempts to run them. 

To make this base framework production-grade, we should replace these remaining SSH instructions:
- **Manifest clean-up**: Instead of deleting files via SSH, we can use a privileged daemonset/job to run the file deletion natively, or document it as a manual operator step outside of Claude's automated capabilities.
- **Verification / Diagnosis**: We can package containerd socket checks and IPsec checks into standard Kubernetes-native jobs/pods similar to the one implemented in Step 1.5.

This will allow us to safely strip `ssh`, `systemctl`, and `journalctl` entirely from `allowed-tools`.

Suyog241005 (changes requested):

Aryanbhargava18:
@Suyog241005 You are absolutely right—relying on SSH from a workstation to a remote edge node behind NAT breaks the OpenYurt cloud-edge tunnel model. Expanding the AI's footprint with SSH credentials is an unnecessary risk.

Great architectural catch.

I just pushed a two-part fix replacing the SSH approach with exactly what you suggested: lightweight, ephemeral Kubernetes Jobs and DaemonSets pinned to the target node.

Commit 208f129e: Replaced the Preflight SSH check with a native Job that verifies the crictl binary and CRI sockets via host mounts.
Commit 9811827d: Eradicated every remaining SSH dependency across the troubleshooting, uninstall, and Raven references using the exact same native execution pattern.

The security boundary is now strictly contained to kubectl and helm. ssh has been completely stripped from allowed-tools across the entire codebase.

Thank you again for the incredibly sharp review on the edge networking constraints, @Suyog241005. It made the architecture significantly stronger.

Suyog241005:
Hi @Aryanbhargava18,

This is a fantastic update! 

I just reviewed the latest changes. The transition to Kubernetes-native verification Jobs, the use of a privileged cleanup DaemonSet for host manifests, and the complete elimination of `/tmp` writes for the Raven test make this skill set extremely secure, robust, and compatible with edge topologies behind NATs/firewalls. 

Removing `ssh`, `systemctl`, and `journalctl` from `allowed-tools` ensures that the AI assistant's execution boundary is strictly limited to the Kubernetes API (`kubectl` and `helm`). This is a huge win for security.

Since this base framework is now in a solid, merge-ready state, I would love to collaborate on the next phase. I can take ownership of implementing the `/openyurt-service-topology` skill or help with the validation/testing of this deploy/raven flow. 

Let me know what you think is the best way to split the remaining scope!

Aryanbhargava18:
@Suyog241005 Thanks for the thorough review — the push toward a fully Kubernetes-native execution boundary genuinely strengthened this framework.

Regarding collaboration: since this PR establishes the core deploy/raven foundation, my focus is on getting it reviewed and aligned with the maintainers first. For the service topology skill, I'd recommend opening a separate PR and adopting the same SKILL.md + references/ architecture and kubectl/helm-only allowed-tools constraints we solidified here — that way the skills stay consistent across the repo.

Happy to review your execution models once you have something drafted.

Aryanbhargava18:
@zyjhtangtang Thanks for getting the webhook fix merged! I've just rebased this architecture PR on top of latest master so it includes those new NodeSelector rules. The deployment workflows are fully unblocked and ready for your final review whenever you have the bandwidth.

Aryanbhargava18:
update: flushed out the logic for the two core skills:
1. raven topology: debugs cross-region L3/L7 routing via gateway and ipsec tunnel inspection.
2. yurthub autonomy: verifies edge node autonomy annotations and local cache integrity during cloud disconnections.

architecture is stable.

alokkumardalei-wq:
hey @Aryanbhargava18 i was checking issue 2559 and came here to review the new .claude/skills structure. noticed two things:

- in openyurt-deploy/SKILL.md , it still says preflight checks run "via SSH". didn't you mention completely ripping out SSH for ephemeral jobs? you probably want to update the docs so claude doesn't try to use ssh by mistake.
- it looks like you accidentally pulled in commits from another branch or messed up a rebase. the branch has over 2,200 lines of unrelated go code changes from other people.

can you remove the unnecessary commits so we can cleanly review the preflight logic 😊.

Aryanbhargava18:
Thanks for catching that, @alokkumardalei-wq! Good eye.

1. I've updated the `SKILL.md` file to completely remove the outdated SSH references. The documentation now correctly reflects the Kubernetes-native ephemeral Job execution model.
2. Apologies for the noise! I had a botched local rebase that inadvertently pulled in some rewritten upstream commits. I've sliced those out and force-pushed a clean history. The PR diff is back to normal now.

Let me know if you spot anything else during your review!

Aryanbhargava18:
@Suyog241005 ssh dependencies have been fully replaced with k8s-native jobs. feel free to resolve the requested changes to unblock the checks.

architecture is stable and ready.

PR #2595, opened 2026-07-18
feat: add Claude Code skill for configuring and troubleshooting service topology

7se7en72025:
#### What type of PR is this?

/kind enhancement
/sig ai

Adds a Claude Code skill, `openyurt-service-topology`, that helps operators configure and troubleshoot service topology — keeping Service traffic within a node or a NodePool.

This is complementary to the deploy/Raven skills proposed in #2559 and drafted in #2592, and deliberately does not overlap them. It follows the modular layout preferred in that discussion: a short `SKILL.md` entrypoint with frontmatter (`disable-model-invocation: true`, scoped `allowed-tools`), plus reference documents holding the detail.

```
.claude/skills/openyurt-service-topology/
├── SKILL.md                                  # entrypoint + frontmatter
└── references/
    ├── service-topology-configure.md         # setup + runnable worked example
    ├── service-topology-diagnose.md          # symptom index + 10-step decision tree
    └── service-topology-internals.md         # behaviour model, for explaining results
```

Tool access is read-only by default (`kubectl get/describe/logs`, `Read`, `Grep`); the only mutation is `kubectl annotate service`, and the skill is instructed to confirm before using it.

Every behavioural claim in the references is cited to a specific file and line so a reviewer can check it rather than take it on trust.

#### Two issues found while writing this against current source

These are documented in the references, but both are really code-level observations and I'm happy to split either into its own issue if maintainers prefer.

**1. The `servicetopology` filter is registered for `endpoints`, but never filters them.**

`cmd/yurthub/app/options/filters.go:56-57` registers the filter for both `endpoints` and `endpointslices`. However `Filter()` only type-switches on `*discovery.v1.EndpointSlice` and `*discovery.v1beta1.EndpointSlice`; anything else falls through to `default` and is returned unchanged (`pkg/yurthub/filter/servicetopology/filter.go:131-138`). No filter in the yurthub tree handles `v1.Endpoints`.

Practical effect: a component consuming the legacy `endpoints` resource bypasses service topology entirely, while appearing correctly configured. The registration and the implementation disagree.

**2. Annotation changes propagate only because yurt-manager forces a watch event.**

The filter runs against `list`/`watch` responses, so it re-evaluates only when the EndpointSlice is re-sent. Changing `openyurt.io/topologyKeys` mutates the *Service*, which by itself produces no EndpointSlice event. `yurt-manager` bridges this by detecting the changed annotation (`pkg/yurtmanager/controller/servicetopology/util/util.go:25-29`), enqueuing the Service's EndpointSlices (`.../endpointslice/endpointslice_enqueue_handlers.go:44-62`), and patching `openyurt.io/update-trigger` (`.../adapter/adapter.go:53-54`).

Practical effect: with `servicetopologyendpointslices` / `servicetopologyendpoints` disabled, annotation changes are silently inert until unrelated pod churn delivers them — which presents as intermittent rather than broken. This is now an explicit diagnostic step in the skill.

#### `.gitignore` change

`.gitignore` excluded `.claude` outright, so files under it could only be committed with `git add -f`. This PR narrows it:

```gitignore
.claude/*
!.claude/skills/
```

so shipped skills are tracked while local `settings.local.json` stays ignored. Note a directory cannot be re-included once its parent is excluded, which is why the parent pattern had to change rather than just adding a negation.

This also affects #2592, whose files sit under `.claude/commands/` and appear to have been force-added. Not a criticism of that PR — just flagging that whichever lands first should probably own the fix, and the other rebase onto it. Happy to adjust my path to match if maintainers want a single canonical location (`.claude/skills/` vs `.claude/commands/`).

Related to #2559

- The skill documents behaviour verified by reading the current implementation; it has **not** been exercised end-to-end against a live edge cluster. The worked example in the configure reference is written to be runnable, but I'd value confirmation from someone with a multi-NodePool environment.
- I recently opened #2593 (nil `NodeName` panic in this same filter). The diagnose reference mentions it in the context of crash-looping yurthub; if that PR is rejected the reference to it should be dropped.
- AI assistance (Claude) was used to research the implementation and draft these documents; all cited line references were verified against the tree at the time of writing.

```release-note
NONE
```

### DIFF
--- a/.claude/skills/openyurt-service-topology/SKILL.md
+++ b/.claude/skills/openyurt-service-topology/SKILL.md
@@ -1,0 +1,76 @@
+---
+name: openyurt-service-topology
+description: Configure and troubleshoot OpenYurt service topology, so that Service traffic stays within a node or a NodePool.
+when_to_use: Use when the user wants to keep Service traffic inside a node or NodePool, is configuring the openyurt.io/topologyKeys annotation, or is debugging why edge traffic is still crossing NodePool boundaries.
+disable-model-invocation: true
+allowed-tools: >
+  Bash(kubectl get *)
+  Bash(kubectl describe *)
+  Bash(kubectl logs *)
+  Bash(kubectl annotate service *)
+  Read
+  Grep
+---
+
+# OpenYurt Service Topology
+
+Service topology restricts which endpoints a client sees, so traffic from an edge
+node is served by a backend on the same node or in the same NodePool instead of
+crossing the network back to the cloud.
+
+It is enforced **on the edge node, inside yurthub**, not by the API server. yurthub
+intercepts `list`/`watch` responses for `endpointslices` and removes the endpoints
+outside the requested topology (`pkg/yurthub/filter/servicetopology/filter.go`).
+
+Two things follow, and most confusion traces to one of them:
+
+- **Stored objects are never modified.** `kubectl get endpointslice` from your laptop
+  always shows the **unfiltered** list. Only a request from a filtered component, on
+  the node, sees the reduced set. Never conclude topology works from a workstation.
+- **The feature has a cloud half too.** Changing the annotation mutates the *Service*,
+  which by itself generates no EndpointSlice watch event. `yurt-manager` patches the
+  EndpointSlice to force one, so the filter re-runs. If that controller is not
+  running, annotation changes are silently inert.
+
+## Configuration model
+
+Topology is opt-in per Service, via an annotation on the **Service** (not the
+EndpointSlice):
+
+| Annotation value | Traffic is kept within |
+| --- | --- |
+| `kubernetes.io/hostname` | the same node |
+| `openyurt.io/nodepool` | the same NodePool |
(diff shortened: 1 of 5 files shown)


Aryanbhargava18:
Hey @7se7en72025, thanks for putting this together! The `openyurt-service-topology` skill is a fantastic complementary addition to the foundation laid out in #2592.

Also, great catch on the `.gitignore` rule. I just pushed a commit to #2592 that aligns our core deployment and Raven skills into the `.claude/skills/` directory and adopts your `.gitignore` patch so we have a single canonical location going forward.

Your discovery of the `servicetopology` filter bypassing `v1.Endpoints` and the annotation propagation requiring a `yurt-manager` watch event are exactly the kind of deep-dive insights we want these AI skills to surface. I highly recommend opening separate tracking issues for those two bugs so the core maintainers can triage them outside of this PR.

Since this builds on the modular architecture we discussed in #2559, I look forward to seeing this land. Excellent work pushing this ecosystem forward!

PR #2601, opened 2026-07-19
refactor(webhook): safely merge gateway nodeSelector to prevent overwriting user labels

### LINKED ISSUE #2600: [Refactor]: Standardize MutatingWebhook nodeSelector patch logic across Raven components
## Description
While auditing the webhook conversion chains for `#2559` (Claude Code Skills integration), I noticed that the `SetDefaultsGateway` logic in `pkg/yurtmanager/webhook/gateway/v1beta1/gateway_default.go` blindly applies the `raven.openyurt.io/gateway` nodeSelector without first checking if the user intentionally passed a more restrictive selector (or if one already exists). 

This behavior can silently overwrite user configurations during edge node deployments if a user attempts to restrict the Gateway to a subset of nodes via custom labels.

## Proposed Solution
1. Refactor `SetDefaultsGateway` to perform a deep-merge of `spec.nodeSelector` rather than an unconditional overwrite.
2. Expose a shared utility in `pkg/apis/raven` for safe label merging to be used by other mutating webhooks to prevent similar data-loss bugs in the future.
3. Add unit tests for the webhook default assignments.

I can throw up a PR for this immediately if the SIG architecture team agrees with the merge strategy.

Aryanbhargava18:
Fixes #2600

## What type of PR is this?
/kind bug
/sig network

## What this PR does / why we need it:
This PR refactors `SetDefaultsGateway` in `pkg/apis/raven/v1beta1/default.go` to perform a safe map-merge when applying the `raven.openyurt.io/gateway: <name>` nodeSelector label. Previously, the logic blindly initialized a new `&metav1.LabelSelector` struct, which would unconditionally overwrite any user-provided edge node selector labels during the mutating webhook intercept. This ensures robust support for custom node pools.

## Which issue(s) this PR fixes:
Fixes #2600

## Special notes for your reviewer:
I have manually verified that this correctly preserves existing labels inside `MatchLabels`.

### DIFF
--- a/pkg/apis/raven/v1beta1/default.go
+++ b/pkg/apis/raven/v1beta1/default.go
@@ -25,11 +25,16 @@ import (
 // SetDefaultsGateway set default values for Gateway.
 func SetDefaultsGateway(obj *Gateway) {
 	// Set default value for Gateway
-	obj.Spec.NodeSelector = &metav1.LabelSelector{
-		MatchLabels: map[string]string{
-			raven.LabelCurrentGateway: obj.Name,
-		},
+	if obj.Spec.NodeSelector == nil {
+		obj.Spec.NodeSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{},
+		}
 	}
+	if obj.Spec.NodeSelector.MatchLabels == nil {
+		obj.Spec.NodeSelector.MatchLabels = map[string]string{}
+	}
+	// Safely merge the gateway label without overwriting existing labels
+	obj.Spec.NodeSelector.MatchLabels[raven.LabelCurrentGateway] = obj.Name
 	for idx, val := range obj.Spec.Endpoints {
 		if val.Port == 0 {
 			switch val.Type {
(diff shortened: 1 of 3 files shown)


Aryanbhargava18:
just ran through the raven webhook tests again after rebasing on latest master — the safe merge logic correctly preserves all existing `MatchLabels` entries while injecting the gateway label. no regressions on my end, ready for review whenever you get a chance 👍

Aryanbhargava18:
@zyjhtangtang thanks for triggering the rerun. I just pushed a fix to address the two CI issues: I resolved the `gci` linter violation by aligning the map literals in the unit test, and I added a new dedicated test case to cover the exact edge case where `MatchLabels` is nil, achieving 100% patch coverage in `default.go`.

Could you approve the workflow run for this latest commit when you get a chance?

Also, whenever you have a moment, I'd love your review on the AI skills architecture in #2592. I've completely removed all SSH dependencies in favor of native Kubernetes Jobs to handle the security and NAT routing constraints.

Aryanbhargava18:
@zyjhtangtang It looks like the `e2e-tests` workflow just failed on a known flake (`Hub leader config owner cleanup [It] Should delete hub leader config when nodepool is deleted` in `hubleader.go:303`). Since this is entirely unrelated to the Raven Gateway nodeSelector patch, could you re-trigger the e2e-tests job?

zyjhtangtang (approved):

PR #2604, opened 2026-07-19
fix(yurtadm): fix http.Request reuse and context propagation in yurthub health checks

### LINKED ISSUE #2594: bug: http.Request reused across polling iterations — undefined behavior and missing context propagation
---
name: Bug Report
title: "[BUG] http.Request reused across polling iterations — undefined behavior and missing context propagation in CheckYurthubHealthz / CheckYurthubReadyz"
labels: kind/bug

---

**What happened**:

In both `CheckYurthubHealthz` and `CheckYurthubReadyz` (`pkg/yurtadm/util/yurthub/yurthub.go`), a single `*http.Request` is constructed **once, before** the `wait.PollUntilContextTimeout` closure, and the same pointer is passed to `client.Do(req)` on **every** polling iteration.

Relevant code (commit `f01cbf56`):

```go
// CheckYurthubHealthz — lines 428–445
func CheckYurthubHealthz(yurthubServer string) error {
    req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerHealthzURLPath), nil)
    if err != nil {
        return err
    }
    client := &http.Client{}
    return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
        resp, err := client.Do(req)   // same req reused every iteration
        ...
    })
}
```

Permalink: https://github.com/openyurtio/openyurt/blob/f01cbf5655383d1c695cfb72097827bc9d22fb8b/pkg/yurtadm/util/yurthub/yurthub.go#L429

The identical pattern appears in `CheckYurthubReadyz` at lines 449–464 (request created at line 449, reused at line 455).

Two distinct issues are present:

1. **Request reuse across iterations.** Go's `net/http` does not support reusing an `*http.Request` after it has been passed to `Client.Do`. The transport may mutate request internals (e.g., header normalization, redirect tracking). Reusing the consumed object on subsequent poll iterations produces undefined behavior and can cause silent failures.

2. **Context from `PollUntilContextTimeout` is never used.** The closure signature is `func(ctx context.Context) (bool, error)`, but the `ctx` parameter — which carries the per-iteration deadline and cancellation signal from `PollUntilContextTimeout` — is **never referenced** inside either closure body. The request is created with no context at all (`http.NewRequest`, not `http.NewRequestWithContext`). As a result, an in-flight HTTP call cannot be interrupted when the overall 300-second timeout expires or when the outer context is cancelled.

Each polling iteration should use a freshly constructed `*http.Request`. The per-iteration `ctx` supplied by `PollUntilContextTimeout` should be wired into the request so that cancellation and deadline signals propagate to the in-flight HTTP call, allowing it to be interrupted when the poll timeout fires.

Inspect the source directly:

```bash
sed -n '428,464p' pkg/yurtadm/util/yurthub/yurthub.go
```

The output confirms:
- `req` is assigned at line 429 (outside the closure opened at line 434).
- `client.Do(req)` at line 435 is inside the closure — the same `req` is reused on every call.
- The `ctx context.Context` parameter at line 434 is never referenced in the closure body (lines 435–443).
- The same pattern repeats at lines 449 (outer `req`) and 455 (`client.Do(req)` inside the closure), with `ctx` again unused.

No running environment is required; the issue is statically visible in the source.

The Go documentation for `http.Client.Do` states that the request body is consumed after the call returns and that the `Request` should not be reused. The `context.Context` carried on a request (via `http.NewRequestWithContext`) is the standard mechanism for propagating cancellation to blocking network I/O in Go; omitting it means the HTTP call is uninterruptible regardless of what the outer context signals.

Both `CheckYurthubHealthz` and `CheckYurthubReadyz` are affected by both issues identically.

**Environment**:
- OpenYurt version: `f01cbf5655383d1c695cfb72097827bc9d22fb8b` (HEAD at time of report)
- Install tools: No response
- Others: No response

**others**

/kind bug

deepak0x:
#### What type of PR is this?

/kind bug

`CheckYurthubHealthz` and `CheckYurthubReadyz` in `pkg/yurtadm/util/yurthub/yurthub.go` built a single `*http.Request` before `wait.PollUntilContextTimeout` and reused the same pointer in `client.Do(req)` on every polling iteration. Reusing a request after it has been consumed by a round-trip is undefined behavior in `net/http`. In addition, the per-iteration `ctx` supplied by `PollUntilContextTimeout` was never wired into the request (the request was built with `http.NewRequest`, not `http.NewRequestWithContext`), so an in-flight HTTP call could not be interrupted when the 300s timeout expired or the context was cancelled.

This PR:
- Extracts the shared polling logic of both functions into a single helper, `pollYurthubEndpointOK(url, interval, timeout)`, removing the duplicated closure.
- Builds a fresh `*http.Request` on every iteration with `http.NewRequestWithContext(ctx, ...)`, so deadline and cancellation signals propagate to the in-flight request.
- Defers `resp.Body.Close()` so response bodies are not leaked across iterations.

Behavior for callers is unchanged: both public functions still target port `10267` with the same 5s interval and 300s timeout.

Fixes #2594

This overlaps with the existing #2602, which applies the same core fix. The difference here is that the two identical closures are de-duplicated into one testable helper, and the change ships with unit tests. The helper takes the interval and timeout as parameters, which makes the polling path testable without waiting on the hardcoded 300s timeout. Added tests cover:
- immediate success (endpoint returns `OK`),
- success after several non-OK responses (a fresh request is used each iteration),
- timeout when the endpoint never returns `OK`,
- context propagation: a hanging request is interrupted when the timeout fires. Against the old reuse/no-context code this test deadlocks until the process timeout; with the fix it returns promptly.

```
$ go test ./pkg/yurtadm/util/yurthub/ -run Test_pollYurthubEndpointOK -v
=== RUN   Test_pollYurthubEndpointOK_Success
--- PASS: Test_pollYurthubEndpointOK_Success (0.00s)
=== RUN   Test_pollYurthubEndpointOK_EventualOK
--- PASS: Test_pollYurthubEndpointOK_EventualOK (0.10s)
=== RUN   Test_pollYurthubEndpointOK_TimesOutWhenNeverOK
--- PASS: Test_pollYurthubEndpointOK_TimesOutWhenNeverOK (0.30s)
=== RUN   Test_pollYurthubEndpointOK_ContextPropagation
--- PASS: Test_pollYurthubEndpointOK_ContextPropagation (0.40s)
PASS
ok      github.com/openyurtio/openyurt/pkg/yurtadm/util/yurthub 0.815s
```

```release-note
Fix reused http.Request and missing context propagation in yurtadm CheckYurthubHealthz and CheckYurthubReadyz polling.
```

### DIFF
--- a/pkg/yurtadm/util/yurthub/yurthub.go
+++ b/pkg/yurtadm/util/yurthub/yurthub.go
@@ -439,7 +439,9 @@ func CheckYurthubReadyz(yurthubServer string) error {
 func pollYurthubEndpointOK(url string, interval, timeout time.Duration) error {
 	client := &http.Client{}
 	return wait.PollUntilContextTimeout(context.Background(), interval, timeout, true, func(ctx context.Context) (bool, error) {
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		reqCtx, cancel := context.WithTimeout(ctx, interval)
+		defer cancel()
+		req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, url, nil)
 		if err != nil {
 			return false, err
 		}
--- a/pkg/yurtadm/util/yurthub/yurthub.go
+++ b/pkg/yurtadm/util/yurthub/yurthub.go
@@ -426,36 +426,28 @@ func CheckYurthubServiceHealth(yurthubServer string) error {
 
 // CheckYurthubHealthz check if YurtHub is healthy.
 func CheckYurthubHealthz(yurthubServer string) error {
-	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerHealthzURLPath), nil)
-	if err != nil {
-		return err
-	}
-	client := &http.Client{}
-	return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
-		resp, err := client.Do(req)
-		if err != nil {
-			return false, nil
-		}
-		ok, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return false, nil
-		}
-		return string(ok) == "OK", nil
-	})
+	url := fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerHealthzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
 }
 
 // CheckYurthubReadyz check if YurtHub's certificates are ready or not
 func CheckYurthubReadyz(yurthubServer string) error {
-	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerReadyzURLPath), nil)
-	if err != nil {
-		return err
-	}
+	url := fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerReadyzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
+}
+
+func pollYurthubEndpointOK(url string, interval, timeout time.Duration) error {
 	client := &http.Client{}
-	return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextTimeout(context.Background(), interval, timeout, true, func(ctx context.Context) (bool, error) {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+		if err != nil {
+			return false, err
+		}
 		resp, err := client.Do(req)
 		if err != nil {
 			return false, nil
 		}
+		defer resp.Body.Close()
 		ok, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return false, nil
--- a/pkg/yurtadm/util/yurthub/yurthub.go
+++ b/pkg/yurtadm/util/yurthub/yurthub.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
@@ -424,14 +425,28 @@ func CheckYurthubServiceHealth(yurthubServer string) error {
 	return nil
 }
 
-// pollYurthubEndpoint polls the specified endpoint on YurtHub until it returns "OK".
-func pollYurthubEndpoint(yurthubServer, endpointPath string) error {
-	client := &http.Client{}
-	url := fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), endpointPath)
-	return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
+// CheckYurthubHealthz check if YurtHub is healthy.
+func CheckYurthubHealthz(yurthubServer string) error {
+	url := fmt.Sprintf("http://%s%s", net.JoinHostPort(yurthubServer, "10267"), constants.ServerHealthzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
+}
+
+// CheckYurthubReadyz check if YurtHub's certificates are ready or not
+func CheckYurthubReadyz(yurthubServer string) error {
+	url := fmt.Sprintf("http://%s%s", net.JoinHostPort(yurthubServer, "10267"), constants.ServerReadyzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
+}
+
+func pollYurthubEndpointOK(url string, interval, timeout time.Duration) error {
+	// Bound each request with a client timeout that prevents a blackholed or
+	// stalled socket from hanging for the full poll window, while staying
+	// decoupled from the poll interval so slow-but-valid edge nodes don't trip
+	// a false negative. Outer poll cancellation still propagates via ctx.
+	client := &http.Client{Timeout: min(interval*3, timeout)}
+	return wait.PollUntilContextTimeout(context.Background(), interval, timeout, true, func(ctx context.Context) (bool, error) {
 		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
 		if err != nil {
-			return false, nil
+			return false, err
 		}
 		resp, err := client.Do(req)
 		if err != nil {
@@ -446,18 +461,8 @@ func pollYurthubEndpoint(yurthubServer, endpointPath string) error {
 	})
 }
 
-// CheckYurthubHealthz check if YurtHub is healthy.
-func CheckYurthubHealthz(yurthubServer string) error {
-	return pollYurthubEndpoint(yurthubServer, constants.ServerHealthzURLPath)
-}
-
-// CheckYurthubReadyz check if YurtHub's certificates are ready or not
-func CheckYurthubReadyz(yurthubServer string) error {
-	return pollYurthubEndpoint(yurthubServer, constants.ServerReadyzURLPath)
-}
-
 func CheckYurthubReadyzOnce(yurthubServer string) bool {
-	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerReadyzURLPath), nil)
+	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s%s", net.JoinHostPort(yurthubServer, "10267"), constants.ServerReadyzURLPath), nil)
 	if err != nil {
 		return false
 	}
@@ -466,7 +471,6 @@ func CheckYurthubReadyzOnce(yurthubServer string) bool {
 	if err != nil {
 		return false
 	}
-	defer resp.Body.Close()
 	ok, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return false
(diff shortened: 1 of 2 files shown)


pkg/yurtadm/util/yurthub/yurthub.go
```diff
@@ -426,36 +426,28 @@ func CheckYurthubServiceHealth(yurthubServer string) error {
...
-	}
+	url := fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerReadyzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
+}
+
+func pollYurthubEndpointOK(url string, interval, timeout time.Duration) error {
 	client := &http.Client{}
-	return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextTimeout(context.Background(), interval, timeout, true, func(ctx context.Context) (bool, error) {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
Aryanbhargava18, line 447: This refactor cleans up the `CheckYurthubHealthz` signature nicely, but it misses a critical context leak in the HTTP client itself.

By passing the overarching `wait.PollUntilContextTimeout` parent context (which has a 5-minute timeout) directly into `http.NewRequestWithContext(ctx)`, you are binding the HTTP request to the full 5-minute window without any per-request timeout. If the YurtHub server blackholes the TCP connection or stalls, `client.Do(req)` will hang for the entire 5 minutes waiting on that single socket, completely destroying the 5-second polling `interval` you configured.

To do this safely, you must either initialize the client with a timeout shorter than the poll interval (`client := &http.Client{Timeout: interval}`), or derive a child context with a timeout inside the closure (`ctxReq, cancel := context.WithTimeout(ctx, interval)`). Otherwise, a single dropped packet will cause the health check to stall for 300 seconds.

deepak0x: Good catch, fixed in c23865d3. Each probe now derives its own context with context.WithTimeout(ctx, interval) inside the poll closure, so a stalled or blackholed socket is capped at one interval instead of holding client.Do for the full 300s. The parent PollUntilContextTimeout deadline and cancellation still propagate through the child.

I went with the child-context approach over http.Client{Timeout: interval} so the outer poll cancellation stays wired through, and added a regression test: a handler that blocks on r.Context().Done() is now hit across multiple iterations (>=2) within the window. Against the old single-context code that test fails with only 1 hit, since the first request consumes the entire timeout.

pkg/yurtadm/util/yurthub/yurthub.go
```diff
@@ -426,36 +426,30 @@ func CheckYurthubServiceHealth(yurthubServer string) error {
 
 // CheckYurthubHealthz check if YurtHub is healthy.
 func CheckYurthubHealthz(yurthubServer string) error {
```
Aryanbhargava18, line 429: Because the tests were refactored to only cover the internal helper, these exported APIs dropped test coverage, which is causing the Codecov failure. In #2602, I avoided refactoring this and instead mocked the endpoints directly in the tests to maintain 100% coverage on the public APIs.

pkg/yurtadm/util/yurthub/yurthub.go
```diff
@@ -426,36 +426,30 @@ func CheckYurthubServiceHealth(yurthubServer string) error {
...
-		return err
-	}
+	url := fmt.Sprintf("http://%s%s", fmt.Sprintf("%s:10267", yurthubServer), constants.ServerReadyzURLPath)
+	return pollYurthubEndpointOK(url, time.Second*5, 300*time.Second)
+}
+
+func pollYurthubEndpointOK(url string, interval, timeout time.Duration) error {
 	client := &http.Client{}
-	return wait.PollUntilContextTimeout(context.Background(), time.Second*5, 300*time.Second, true, func(ctx context.Context) (bool, error) {
+	return wait.PollUntilContextTimeout(context.Background(), interval, timeout, true, func(ctx context.Context) (bool, error) {
```
Aryanbhargava18, line 446: Bounding stalled sockets is a great catch. However, tightly coupling the HTTP request timeout to the poll `interval` (5s) might introduce false negatives on slower edge nodes under load. Since `ctx` is already bounded by the overall 300s timeout, setting `client.Timeout` might be a safer way to prevent hanging sockets without coupling it to the interval.

Aryanbhargava18 (commented):
Hey @deepak0x, thanks for taking a look at this.

As you noted, I opened #2602 a few days ago to patch this exact context leak and request reuse bug.

While I appreciate the effort to DRY up the closures into a helper, I intentionally avoided refactoring the architecture in #2602 because it drops test coverage on the actual public APIs (`CheckYurthubHealthz` and `CheckYurthubReadyz`). That is why the Codecov check is failing here with 50% patch coverage on the exported functions.

In #2602, I utilized an `http.RoundTripper` interceptor in the tests instead. This allowed me to mock the endpoints and achieve 100% patch coverage directly on the public APIs without introducing architectural churn.

Since #2602 was opened first and is currently fully green across all CI/CD gates (Codecov and SonarQube), I'd prefer we stick with that implementation to get this patched for the maintainers safely. If you feel strongly about the strict 5s request timeout decoupling, I'd be more than happy to incorporate that specific logic into #2602!

deepak0x:
> Hey @deepak0x, thanks for taking a look at this.
> 
> As you noted, I opened #2602 a few days ago to patch this exact context leak and request reuse bug.
> 
> While I appreciate the effort to DRY up the closures into a helper, I intentionally avoided refactoring the architecture in #2602 because it drops test coverage on the actual public APIs (`CheckYurthubHealthz` and `CheckYurthubReadyz`). That is why the Codecov check is failing here with 50% patch coverage on the exported functions.
> 
> In #2602, I utilized an `http.RoundTripper` interceptor in the tests instead. This allowed me to mock the endpoints and achieve 100% patch coverage directly on the public APIs without introducing architectural churn.
> 
> Since #2602 was opened first and is currently fully green across all CI/CD gates (Codecov and SonarQube), I'd prefer we stick with that implementation to get this patched for the maintainers safely. If you feel strongly about the strict 5s request timeout decoupling, I'd be more than happy to incorporate that specific logic into #2602!

Firstly never raise pr for someone else issue without asking him
It's not a good open source practice 
Also as the issue is mine so mine pr will get edge irrespective of who raise first
Always ask first before taking someone else issue 🙂
Hope u got it

Aryanbhargava18:
Hey @deepak0x, I didn't realize you were the original author of the issue. You're right to claim it, so I've closed #2602 to let yours proceed.

That said, #2602 already solved this with 100% coverage on the public API by using an `http.RoundTripper` interceptor. I'd highly recommend adapting that testing approach here so you don't drop coverage on the exported functions.

Also, let's keep the communication professional and collaborative moving forward. We're all on the same team here. Cheers.

deepak0x:
Thanks for closing #2602 and for the pointer, appreciated. I took your suggestion and switched the wrapper tests to an http.RoundTripper installed on http.DefaultTransport (b1b7daf7), so CheckYurthubHealthz and CheckYurthubReadyz are covered deterministically without binding a real port. Both are at 100% now, and the test asserts each wrapper hits its expected URL path.

On the red e2e-tests run: it's failing in the kube-proxy edge-autonomy spec (curl to the nginx service cluster IP times out, exit 28) in test/e2e/autonomy, which is unrelated to this yurtadm change. Could a maintainer kick off a re-run when convenient? Happy to dig in if it reproduces.

And agreed on keeping it collaborative, thanks for the note.

Aryanbhargava18 (approved):
I took a deep dive into the new `b1b7daf` commit. The `RoundTripper` implementation is perfectly executed.

By explicitly capturing and restoring `http.DefaultTransport` with a `defer` block, you safely isolated the mock and prevented global state pollution for other tests in the package. 

Codecov is back to 100% on the public APIs, and the original context leak is fully resolved. Great work getting this locked down.

Aryanbhargava18:
Yeah, the `kube-proxy` edge-autonomy spec occasionally hits an `exit 28` (Operation timeout) due to cluster IP routing delays in the CI environment. It's definitely a known environmental flake and completely unrelated to your `yurthub` polling changes here.

A maintainer will need to hit the re-run button since we don't have write access, but the code itself is solid.

deepak0x:
@Copilot Replying to your review on yurthub.go: addressed in db65d72. The healthz/readyz URLs now use net.JoinHostPort(yurthubServer, "10267") at all three sites (including CheckYurthubReadyzOnce), so IPv6 literals are bracketed correctly instead of producing an invalid host:port.

deepak0x:
@Copilot Replying to your second review: the per-request context is no longer bounded to the poll interval. pollYurthubEndpointOK now sets http.Client.Timeout = min(interval*3, timeout) and passes the poll ctx directly into NewRequestWithContext, so (a) a blackholed socket is still capped (no 300s hang), (b) slow-but-valid edge-node requests are not forced into a false negative by the 5s interval, and (c) outer poll cancellation still propagates. The change is documented inline.

deepak0x:
@Aryanbhargava18 On your point about the per-request timeout being coupled to the 5s interval (false negatives on slow edge nodes): addressed in db65d72. I switched from a child context bounded to `interval` to an `http.Client.Timeout` of `min(interval*3, timeout)`, which prevents indefinite hangs without tying the request budget to the poll interval. The exported wrappers also keep their 100% coverage via the RoundTripper-interceptor tests added earlier (b1b7daf / 4fb25a5), which is what the earlier Codecov failure was about. Thanks for the review.

deepak0x:
Addressing the Copilot review on this PR (reviewed an earlier commit; the current head 082ad54 already covers both points):

1. IPv6 host/port formatting: the code now uses net.JoinHostPort(yurthubServer, "10267") at every call site (CheckYurthubHealthz, CheckYurthubReadyz, CheckYurthubReadyzOnce) instead of fmt.Sprintf("%s:10267", ...), so IPv6 literals are bracketed correctly.

2. Per-request timeout / retry behavior: there is no context.WithTimeout(ctx, interval) wrapping. The poll context is passed directly to the request, and the per-request bound is set via client.Timeout = min(interval*3, timeout) with a comment explaining it stays decoupled from the poll interval so slow-but-valid edge nodes do not trip a false negative. Outer poll cancellation still propagates via ctx, and retries are preserved across poll iterations.

zyjhtangtang (approved):

PR #2611, opened 2026-07-21
Bugfix: require CA pinning by default for yurthub bootstrap

### LINKED ISSUE #2610: [BUG] YurtHub token bootstrap skips CA verification by default
**What happened**:

YurtHub token bootstrap skips CA verification by default because `NewYurtHubOptions()` initializes `UnsafeSkipCAVerification` to `true`. As a result, constructing token-bootstrap options without `--discovery-token-ca-cert-hash` passes validation unless the operator explicitly opts back into secure behavior.

## Summary
Token-based YurtHub bootstrap currently defaults to insecure discovery. Operators who rely on defaults can join without CA public key pinning, which weakens first-join trust establishment and leaves the bootstrap path exposed to man-in-the-middle interception.

## Affected Area
- `cmd/yurthub/app/options/options.go`
- `cmd/yurthub/app/options/options_test.go`

Secure bootstrap should be the default. YurtHub should require `--discovery-token-ca-cert-hash` unless the operator explicitly sets `--discovery-token-unsafe-skip-ca-verification=true`.

## Expected Behaviour
YurtHub token bootstrap requires CA hash pinning by default, and insecure discovery is only enabled by an explicit operator opt-in.

## Actual Behaviour
`UnsafeSkipCAVerification` defaults to `true`, so validation allows token bootstrap without any CA hash unless the operator overrides the default.

1. Inspect `NewYurtHubOptions()` in `cmd/yurthub/app/options/options.go`.
2. Observe that `UnsafeSkipCAVerification` defaults to `true`.
3. Construct YurtHub options for token bootstrap without setting `CACertHashes`.
4. Call `Validate()` and observe that validation succeeds as long as the flag remains at its default.

## Steps to Reproduce
1. Run `go test ./cmd/yurthub/app/options -run 'TestNewYurtHubOptions|TestValidate'`.
2. Inspect `cmd/yurthub/app/options/options_test.go` and note that `TestNewYurtHubOptions` currently expects `UnsafeSkipCAVerification: true`.
3. Inspect `cmd/yurthub/app/options/options.go` and note that `Validate()` only rejects missing CA hashes when `UnsafeSkipCAVerification` is `false`.
4. Confirm that the secure path is opt-in rather than default.

## Technical Evidence
- `cmd/yurthub/app/options/options.go` sets `UnsafeSkipCAVerification: true` in `NewYurtHubOptions()`.
- The same file only errors on missing `CACertHashes` when `!o.UnsafeSkipCAVerification`.
- `cmd/yurthub/app/options/options_test.go` currently codifies the insecure default in `TestNewYurtHubOptions`.
- Local verification on July 21, 2026: `go test ./cmd/yurthub/app/options -run 'TestNewYurtHubOptions|TestValidate'` passed against `master` at `5813abc5`.

## Suspected Root Cause
The YurtHub option defaults and validation logic were implemented so that insecure token discovery is the default path, instead of requiring explicit acknowledgement when skipping CA pinning.

## Proposed Direction
- Flip the `UnsafeSkipCAVerification` default to `false` in `NewYurtHubOptions()`.
- Update the related option tests to cover the secure default and explicit insecure opt-in.
- Clarify the flag help text so operators understand that skipping CA verification is an insecure override.

## Acceptance Criteria
- New YurtHub defaults require CA hash pinning unless `--discovery-token-unsafe-skip-ca-verification=true` is set explicitly.
- Unit tests cover the secure default and explicit unsafe opt-in.
- Flag help text makes the insecure override behavior clear.

**Environment**:
- OpenYurt version: `master` at `5813abc5` (local source checkout on 2026-07-21)
- Kubernetes version (use `kubectl version`): Not applicable for this code-level issue
- OS (e.g: `cat /etc/os-release`): macOS development environment
- Kernel (e.g. `uname -a`): Not applicable
- Install tools: `go`, `gh`
- Others: Reproduced through source inspection and the existing `cmd/yurthub/app/options` unit tests

**others**
/kind bug

rootp1:
#### What type of PR is this?
/kind bug

## Summary
- make YurtHub token bootstrap require CA pinning by default
- clarify that `--discovery-token-unsafe-skip-ca-verification` is an insecure override
- add focused unit coverage for the secure default and explicit insecure opt-in paths

## Linked Issue
Fixes #2610

## What Changed
- flipped `NewYurtHubOptions()` so `UnsafeSkipCAVerification` defaults to `false`
- updated the unsafe discovery flag help text to call out the security tradeoff
- extended `cmd/yurthub/app/options` tests to cover secure-default validation and CA-hash-backed success

## Verification
- `go test ./cmd/yurthub/app/options` — passed
- `make verify` — passed
- `make test` — passed
- `make lint` — passed

## Scope
- only `cmd/yurthub/app/options/options.go` and `cmd/yurthub/app/options/options_test.go` were changed

Fixes #2610

- none

```release-note
YurtHub token bootstrap now requires `--discovery-token-ca-cert-hash` by default. Operators must explicitly set `--discovery-token-unsafe-skip-ca-verification=true` to allow insecure discovery without CA pinning.
```

#### other Note
- none

### DIFF
--- a/cmd/yurthub/app/options/options.go
+++ b/cmd/yurthub/app/options/options.go
@@ -121,7 +121,7 @@ func NewYurtHubOptions() *YurtHubOptions {
 		EnableNodePool:            true,
 		MinRequestTimeout:         time.Second * 1800,
 		CACertHashes:              make([]string, 0),
-		UnsafeSkipCAVerification:  true,
+		UnsafeSkipCAVerification:  false,
 		EnablePoolServiceTopology: false,
 		PoolScopeResources: []schema.GroupVersionResource{
 			{Group: "", Version: "v1", Resource: "services"},
@@ -233,7 +233,7 @@ func (o *YurtHubOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.MarkDeprecated("enable-node-pool", "It is planned to be removed from OpenYurt in the future version, please use --enable-pool-service-topology instead")
 	fs.DurationVar(&o.MinRequestTimeout, "min-request-timeout", o.MinRequestTimeout, "An optional field indicating at least how long a proxy handler must keep a request open before timing it out. Currently only honored by the local watch request handler(use request parameter timeoutSeconds firstly), which picks a randomized value above this number as the connection timeout, to spread out load.")
 	fs.StringSliceVar(&o.CACertHashes, "discovery-token-ca-cert-hash", o.CACertHashes, "For token-based discovery, validate that the root CA public key matches this hash (format: \"<type>:<value>\").")
-	fs.BoolVar(&o.UnsafeSkipCAVerification, "discovery-token-unsafe-skip-ca-verification", o.UnsafeSkipCAVerification, "For token-based discovery, allow joining without --discovery-token-ca-cert-hash pinning.")
+	fs.BoolVar(&o.UnsafeSkipCAVerification, "discovery-token-unsafe-skip-ca-verification", o.UnsafeSkipCAVerification, "For token-based discovery, skip verification of the root CA public key. This is insecure and should only be used when --discovery-token-ca-cert-hash cannot be provided.")
 	fs.BoolVar(&o.EnablePoolServiceTopology, "enable-pool-service-topology", o.EnablePoolServiceTopology, "enable service topology feature in the node pool.")
 	fs.StringVar(&o.HostControlPlaneAddr, "host-control-plane-address", o.HostControlPlaneAddr, "the address (ip:port) of host kubernetes cluster that used for yurthub local mode.")
 	fs.Var(&o.PoolScopeResources, "pool-scope-resources", "The list/watch requests for these resources will be multiplexered in yurthub in order to reduce overhead of kube-apiserver. comma-separated list of GroupVersionResource in the format Group/Version/Resource")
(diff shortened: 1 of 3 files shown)


rootp1:
Hi, @rambohe-ch @zyjhtangtang @luc99hen
Please have a look whenever you have a chance to

cmd/yurthub/app/options/options.go
```diff
@@ -121,7 +121,7 @@ func NewYurtHubOptions() *YurtHubOptions {
 		EnableNodePool:            true,
 		MinRequestTimeout:         time.Second * 1800,
 		CACertHashes:              make([]string, 0),
-		UnsafeSkipCAVerification:  true,
+		UnsafeSkipCAVerification:  false,
```
Aryanbhargava18, line 124: Changing the default of `UnsafeSkipCAVerification` from `true` to `false` is a breaking change disguised as a bugfix.

While strict CA pinning is more secure, ANY existing OpenYurt users who currently rely on the default behavior for token bootstrap (which previously skipped verification unless a hash was passed) will suddenly find that their nodes fail to join the cluster upon upgrading to this release.

If we enforce strict security by default, we MUST at minimum provide a deprecation warning cycle or explicitly flag this in the release notes as a Breaking Change. Furthermore, having the node crash on startup because it requires users to manually append `--discovery-token-unsafe-skip-ca-verification=true` to their existing scripts violates Kubernetes backward compatibility standards for CLI flag defaults.

zyjhtangtang (approved):

PR #2616, opened 2026-07-21
Bugfix: make multiplexer request scopes subresource-aware

rootp1:
#### What type of PR is this?
/kind bug

## Summary
- preserve `RequestInfo.Subresource` when building multiplexer request scopes
- register the correct subresource kind for `scale` requests and keep parent resource behavior unchanged
- add regression coverage for subresource and non-subresource request scopes

#### Which issue(s) this PR fixes:
Fixes #2615

#### Special notes for your reviewer:
- Scope is limited to `pkg/yurthub/proxy/multiplexer`.

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

## What Changed
- pass `reqInfo.Subresource` into `getReqScope`
- populate `RequestScope.Subresource`, `Kind`, `HubGroupVersion`, and `EquivalentResourceMapper` with subresource-aware values
- add a deployment `/scale` regression test and a main-resource control test

## Verification
- `go test ./pkg/yurthub/proxy/multiplexer -run 'TestMultiplexerProxyGetReqScope_Subresource|TestShareProxy' -count=1` — passed\n- `go test ./pkg/yurthub/proxy/multiplexer -run TestMultiplexerProxyGetReqScope_Subresource -count=1` — passed\n- `make verify` — passed\n- `make test` — passed\n- `make lint` — passed\n\n## Scope\n- Only `pkg/yurthub/proxy/multiplexer/multiplexerproxy.go` and `pkg/yurthub/proxy/multiplexer/multiplexerproxy_test.go` were changed; unrelated files were not modified.

PR #2617, opened 2026-07-21
Bugfix: return after error in multiplexer proxy handler

### LINKED ISSUE #2614: [BUG]MultiplexerProxy panics when ResourceStore or getReqScope fails
**What happened**:

In `pkg/yurthub/proxy/multiplexer/multiplexerproxy.go`, `ServeHTTP` writes an error response when `ResourceStore()` or `getReqScope()` fails, but does not return afterward. Execution continues and type-asserts a nil `restStore`, which causes YurtHub to panic:

```go
restStore, err := sp.requestsMultiplexerManager.ResourceStore(gvr)
if err != nil {
    util.Err(errors.Wrapf(err, "failed to get rest storage"), w, r)
    // missing return
}

reqScope, err := sp.getReqScope(gvr)
if err != nil {
    util.Err(errors.Wrapf(err, "failed tp get req scope"), w, r)
    // missing return
}

lister := restStore.(rest.Lister)   // panic: nil interface
watcher := restStore.(rest.Watcher)
```

This is inconsistent with the correct pattern used a few lines above (lines 75–78), where `util.Err(...)` is followed by `return`.

When `ResourceStore()` or `getReqScope()` returns an error, YurtHub should return a proper HTTP error response and stop handling the request. It should not panic.

1. Run this minimal repro against upstream master:

```go
package main

import (
	"net/http"
	"net/http/httptest"
	"os"

	apirequest "k8s.io/apiserver/pkg/endpoints/request"

	"github.com/openyurtio/openyurt/pkg/yurthub/kubernetes/meta"
	mproxy "github.com/openyurtio/openyurt/pkg/yurthub/proxy/multiplexer"
	"github.com/openyurtio/openyurt/pkg/yurthub/multiplexer"
)

func main() {
	tmpDir, _ := os.MkdirTemp("", "mp-repro")
	defer os.RemoveAll(tmpDir)

	restMapper, _ := meta.NewRESTMapperManager(tmpDir)
	mm := &multiplexer.MultiplexerManager{}
	sp := mproxy.NewMultiplexerProxy(mm, restMapper, make(<-chan struct{}))

	req := httptest.NewRequest(http.MethodGet, "/apis/nonexistent.group/v1/widgets", nil)
	info := &apirequest.RequestInfo{
		IsResourceRequest: true,
		APIGroup:          "nonexistent.group",
		APIVersion:        "v1",
		Resource:          "widgets",
		Verb:              "list",
	}
	req = req.WithContext(apirequest.WithRequestInfo(req.Context(), info))

	sp.ServeHTTP(httptest.NewRecorder(), req)
}
```

2. Observe panic:
```
panic: runtime error: invalid memory address or nil pointer dereference
```

Verified on upstream master at commit `81ca6625`.

- **Impact:** Pool-scope metadata list/watch requests can crash YurtHub when storage or REST mapping is unavailable, instead of returning an HTTP error.
- **Suggested fix:** Add `return` after both `util.Err(...)` calls (lines 83 and 88).
- **Related code:** `util.Err` + `return` pattern already used correctly at lines 75–78 in the same file.

**Environment**:
- OpenYurt version: master (`81ca6625`)
- Kubernetes version: N/A (reproduced with unit-level handler test)
- OS: Linux
- Kernel: N/A
- Install tools: Go 1.24+
- Others: Reproduced with standalone Go program against local clone synced with upstream master

**others**

/kind bug

Jay2006sawant:
#### What type of PR is this?
/kind bug

Fixes a panic in `MultiplexerProxy.ServeHTTP` when `ResourceStore()` or `getReqScope()` returns an error.

Previously, the handler called `util.Err()` to write an error response but did not return. Execution continued and type-asserted a nil `restStore`, causing YurtHub to panic instead of returning a proper HTTP error.

This PR adds the missing `return` statements, matching the existing pattern used for the `Ready()` check in the same function (lines 75-78).

Fixes #2614

- Two-line fix in `pkg/yurthub/proxy/multiplexer/multiplexerproxy.go`
- Adds unit tests:
  - `TestGetReqScopeReturnsErrorForUnknownGVR`
  - `TestMultiplexerProxyServeHTTPReturnsOnGetReqScopeError`

**Verification performed locally:**
```bash
go test ./pkg/yurthub/proxy/multiplexer/... -count=1 -v
```
All tests pass.

```release-note
NONE
```

### DIFF
--- a/pkg/yurthub/proxy/multiplexer/multiplexerproxy_test.go
+++ b/pkg/yurthub/proxy/multiplexer/multiplexerproxy_test.go
@@ -20,6 +20,7 @@ package multiplexer
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -36,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/storage"
 	"k8s.io/client-go/informers"
@@ -112,6 +114,23 @@ func (wr *wrapResponse) Write(buf []byte) (int, error) {
 	return l, err
 }
 
+type fakeResourceMultiplexer struct {
+	ready    bool
+	store    rest.Storage
+	storeErr error
+}
+
+func (f *fakeResourceMultiplexer) Ready(gvr *schema.GroupVersionResource) bool {
+	return f.ready
+}
+
+func (f *fakeResourceMultiplexer) ResourceStore(gvr *schema.GroupVersionResource) (rest.Storage, error) {
+	if f.storeErr != nil {
+		return nil, f.storeErr
+	}
+	return f.store, nil
+}
+
 func TestGetReqScopeReturnsErrorForUnknownGVR(t *testing.T) {
 	tmpDir, err := os.MkdirTemp("", "multiplexer-req-scope")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/yurthub/proxy/multiplexer/multiplexerproxy_test.go
```diff
@@ -112,6 +112,69 @@ func (wr *wrapResponse) Write(buf []byte) (int, error) {
...
+		LoadBalancerForLeaderHub: loadBalancer,
+	}
+	rmm := multiplexer.NewRequestMultiplexerManager(cfg, dsm, healthChecker)
+
+	stopCh := make(chan struct{})
+	if ok := cache.WaitForCacheSync(stopCh, func() bool { return rmm.Ready(&endpointSliceGVR) }); !ok {
+		t.Fatal("multiplexer manager is not ready")
+	}
+
+	sp := NewMultiplexerProxy(rmm, proxyRESTMapper, make(<-chan struct{}))
```
Aryanbhargava18: Good catch on the missing returns to prevent the proxy panic. 

However, in `TestMultiplexerProxyServeHTTPReturnsOnGetReqScopeError`, you are passing an unclosed, never-signaled `make(<-chan struct{})` directly into `NewMultiplexerProxy`. Since the multiplexer manager spins up internal cache/informer goroutines that listen on this stop channel, you are silently leaking those background goroutines for the entire duration of the test suite because the channel is never closed. 

It's much safer to pass a `stopCh` that you `defer close(stopCh)` at the end of the test block. 

Also, minor typo in the original error message if you want to fix it while you're here: `failed tp get req scope` -> `to`.

Jay2006sawant: Thanks for the review, @Aryanbhargava18 ! agreed on both points.

- Updated the test to use a shared `stopCh` with `defer close(stopCh)` so background goroutines are stopped cleanly.
- Fixed the typo: `failed tp get req scope` → `failed to get req scope`.

Pushed in e963306d. Please take another look when you have a moment.

Aryanbhargava18 (approved):
Looks great! Thanks for getting those fixes in so quickly. The goroutine lifecycle is solid now and the typo is fixed. 

LGTM!

Aryanbhargava18:
Just a heads up on the failing `e2e-tests` check: it looks like it's hitting the same `kube-proxy` edge-autonomy `exit 28` curl timeout flake that's currently affecting #2604. 

It's an environmental routing delay in the CI and completely unrelated to your proxy handler changes. A maintainer will just need to re-run the job when they get a chance!

Aryanbhargava18 (approved):
The goroutine lifecycle in the proxy test is solidly locked down now with the shared `stopCh`, and the `ServeHTTP` error return prevents the panic cleanly. Approved.

rootp1:
@Aryanbhargava18 @Jay2006sawant 
Hello this is a duplicate PR, already opened one PR 2hours prior to this

Aryanbhargava18:
@rootp1 To clarify, this is not a duplicate.

This PR (#2617) adds missing `return` statements to fix a nil pointer panic in `ServeHTTP` (reported in #2614). PR #2616 introduces subresource-aware request scopes in `getReqScope`. 

They both touch `multiplexerproxy.go`, but they address entirely independent issues.

zyjhtangtang:
@Jay2006sawant  Please pay attention to the unit test coverage.

Jay2006sawant:
@zyjhtangtang Have made the required chamges,PTAL!

zyjhtangtang (approved):

PR #2618, opened 2026-07-21
fix(yurthub): synchronize shared state in cachemanager errorKeys

### LINKED ISSUE #2582: [BUG] Data race in yurthub cachemanager errorKeys
## What happened:
`pkg/yurthub/cachemanager/error_keys.go` runs two background goroutines started in `NewErrorKeys`, `sync()` and `compress()` that access shared fields (`ek.count`, `ek.keys`, `ek.file`) without consistent locking, while `put()`/`del()` mutate them from caller goroutines. `go test -race` reports data races on these fields.

The core problem is in `compress()`:

```go
func (ek *errorKeys) compress() {
	ticker := time.NewTicker(30 * time.Second)
	for range ticker.C {
		if !ek.queue.ShuttingDown() {
			if ek.count > len(ek.keys)+CompressThresh {   // line 148: reads ek.count and ek.keys with NO lock
				ek.rewrite()
			}
		} else {
			return
		}
	}
}
```

Line 148 reads `ek.count` and `len(ek.keys)` while holding no lock, but:
- `ek.count` is written without a lock by `processNextOperator()` (`ek.count++`, line 140) in the `sync()` goroutine, and reassigned by `rewrite()` (line 209) under only an `RLock`.
- `ek.keys` is written under the full `ek.Lock()` by `put()` (line 88) and `del()` (line 99) but the `compress()` reader holds nothing.
- `ek.file` is written by `processNextOperator()` (lines 138–139) with no lock and reassigned by `rewrite()` (line 208), so the sync goroutine can write to a file handle that `rewrite()` is concurrently closing/swapping.

## What you expected to happen:
Access to the shared `errorKeys` state (`count`, `keys`, `file`) should be properly synchronized so that `go test -race` reports no data races, and the AOF persistence path cannot lose writes or make compaction decisions off a torn `count`.

## How to reproduce it (as minimally and precisely as possible):

Run a reader that performs `compress()`'s exact unsynchronized read (`ek.count > len(ek.keys)+CompressThresh`) concurrently with `put()`/`del()`, under the race detector:

```
go test -race -run TestErrorKeysDataRace -count=5 ./pkg/yurthub/cachemanager/
```

On `master` with Go 1.26.1 (linux/amd64) this reports a `WARNING: DATA RACE` on `ek.count` a write by `processNextOperator()` (`error_keys.go:140`, in the `sync()` goroutine) racing the unsynchronized read plus the same reader racing `put()`/`del()` on `ek.keys`. The race is timing-dependent, so it trips on some runs and not others; a single detection confirms it.

<details>
<summary>Minimal reproducer (<code>error_keys_race_test.go</code>)</summary>

```go
package cachemanager

import (
	"fmt"
	"testing"
	"time"
)

func TestErrorKeysDataRace(t *testing.T) {
	ek := NewErrorKeys(t.TempDir())
	stop := make(chan struct{})
	// Mirrors compress()'s unsynchronized read at error_keys.go:148.
	go func() {
		for {
			select {
			case <-stop:
				return
			default:
				_ = ek.count > len(ek.keys)+CompressThresh
			}
		}
	}()
	for i := 0; i < 5000; i++ {
		key := fmt.Sprintf("key-%d", i)
		ek.put(key, "value")
		ek.del(key)
	}
	time.Sleep(100 * time.Millisecond)
	close(stop)
}
```

The reader goroutine is a stand-in for `compress()`'s line-148 read; it is used because `compress()`'s 30s ticker makes triggering the real call impractical in a fast unit test.
</details>

<details>
<summary><code>go test -race</code> output</summary>

```
==================
WARNING: DATA RACE
Write at 0x00c0006103a8 by goroutine 69:
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.(*errorKeys).processNextOperator()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys.go:140 +0x2fc
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.(*errorKeys).sync()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys.go:122 +0x2e
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.NewErrorKeys.gowrap1()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys.go:66 +0xe

Previous read at 0x00c0006103a8 by goroutine 71:
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.TestErrorKeysDataRace.func1()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys_race_test.go:24 +0x34

Goroutine 69 (running) created at:
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.NewErrorKeys()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys.go:66 +0x6f0
  ...
Goroutine 71 (running) created at:
  github.com/openyurtio/openyurt/pkg/yurthub/cachemanager.TestErrorKeysDataRace()
      /workspaces/openyurt/pkg/yurthub/cachemanager/error_keys_race_test.go:18 +0xf9
  ...
==================
    testing.go:1712: race detected during execution of test
--- FAIL: TestErrorKeysDataRace (2.06s)

```
</details>

## Anything else we need to know?:

- **Impact:** the AOF persistence path for cached error keys can (a) make compaction decisions off a torn `ek.count`, and (b) have the `sync()` goroutine write to an `ek.file` handle that `rewrite()` is concurrently swapping/closing, risking lost persisted operations.
- **Two secondary issues in the same file (can be handled together or separately):**

    `rewrite()` holds `ek.RLock()` across file I/O and a `wait.PollUntilContextTimeout(..., time.Minute, ...)` (lines 158–190), blocking all `put()`/`del()` for up to a minute during a rewrite.

   `processNextOperator()` returns `false` on a `json.Marshal` error (lines 134–137), which terminates the `sync()` loop permanently. This is defensively wrong (should log and continue), though in practice `operation` is a struct of strings and marshalling it effectively never fails, minor.
- **Proposed fix direction (seeking maintainer feedack before PR):** guard the `count`/`keys` reads in `compress()` under the existing lock (or make `count` an atomic), and synchronize the `file`/`count` reassignment in `rewrite()` against the `sync()` goroutine's writes. Happy to implement whichever synchronization approach maintainers prefer, with a `-race` regression test included.

## Environment:
- OpenYurt version:  master
- Others: reproduced with go1.26.1 linux/amd64; go.mod declares `go 1.25.0`. This is a missing-synchronization bug in the Go code, not toolchain- or cluster-specific.

/kind bug

7se7en72025:
#### What type of PR is this?
/kind bug

`errorKeys` in `pkg/yurthub/cachemanager/error_keys.go` embeds a `sync.RWMutex` to protect its shared fields, but three code paths don't consistently use it:

- `processNextOperator()` (run in the background `sync()` goroutine started by `NewErrorKeys`) writes to `ek.file` and increments `ek.count` with **no lock at all**.
- `compress()` reads `ek.count` and `ek.keys` for its threshold check with **no lock**.
- `rewrite()` reassigns `ek.file` and `ek.count` at the end of compaction while holding only `ek.RLock()` — a read lock does not protect a write.

The result is a genuine data race between the `sync()` goroutine (constantly running for the life of the hub), the periodic `compress()`/`rewrite()` path, and any caller of `put()`/`del()`. In production this can mean the compaction threshold is computed from a torn `count`, and the `sync()` goroutine can write into a file handle that `rewrite()` is concurrently closing and swapping out from under it, risking lost or corrupted AOF writes used to persist yurthub's error-key state across restarts.

This isn't hypothetical: running the *existing* `TestCompress` test with `-race` already reproduces a race today, because that test itself reads `keys.count` directly while `sync()` is concurrently writing it — the accessor added by this PR (`errorCount()`) fixes that too.

#### The fix

- `processNextOperator()`: take `ek.Lock()` around the file write + `ek.count++`.
- `compress()`: take `ek.RLock()` for the `count`/`keys` threshold check.
- `rewrite()`: switch from `ek.RLock()` to `ek.Lock()`, since it mutates `ek.file`/`ek.count`, not just reads them.
- Added `errorCount()`, a locked accessor for `ek.count`, and switched `TestCompress` to use it instead of reading the field directly.

This is intentionally scoped to the synchronization bug itself. `rewrite()` already effectively serialized against `put()`/`del()` before this change (a read lock blocks a writer trying to take the write lock), so switching it to a full write lock does not introduce new blocking behavior on that path; it only removes the race against the unlocked `sync()` goroutine and the unlocked `compress()` read.

Fixes #2582

- Verified the new test actually catches the regression: temporarily reverted just the locking changes (keeping the new `errorCount()` helper) and confirmed `TestErrorKeysConcurrentAccessRace` fails under `-race` against that old code, then restored the fix and confirmed it passes.
- `go build ./...`, `go vet ./pkg/yurthub/cachemanager/...`, and `gofmt -l` on the changed files are all clean.
- `go test -race -count=1 ./pkg/yurthub/cachemanager/...` passes (includes the pre-existing `TestCompress`, which itself now exercises the race-free path).
- Ran the broader `go test ./pkg/yurthub/...` tree afterward as a sanity check; everything still passes.
- AI assistance (Claude) was used to help implement and verify this change.

```release-note
Fix a data race in YurtHub's cache manager `errorKeys` AOF persistence path where the background sync/compaction goroutines could read and write shared state without holding the protecting lock.
```

#### other Note

### DIFF
--- a/pkg/yurthub/cachemanager/error_keys.go
+++ b/pkg/yurthub/cachemanager/error_keys.go
@@ -118,6 +118,15 @@ func (ek *errorKeys) length() int {
 	return len(ek.keys)
 }
 
+// errorCount returns the number of persisted operations recorded so far. It
+// takes the same lock that guards ek.count and ek.keys so it can be read
+// safely while sync() and compress() are running concurrently.
+func (ek *errorKeys) errorCount() int {
+	ek.RLock()
+	defer ek.RUnlock()
+	return ek.count
+}
+
 func (ek *errorKeys) sync() {
 	for ek.processNextOperator() {
 	}
@@ -135,9 +144,13 @@ func (ek *errorKeys) processNextOperator() bool {
 		klog.Errorf("failed to serialize and persist operation: %v", op)
 		return false
 	}
+
+	ek.Lock()
 	ek.file.Write(append(data, '\n'))
 	ek.file.Sync()
 	ek.count++
+	ek.Unlock()
+
 	return true
 }
 
@@ -145,7 +158,10 @@ func (ek *errorKeys) compress() {
 	ticker := time.NewTicker(30 * time.Second)
 	for range ticker.C {
 		if !ek.queue.ShuttingDown() {
-			if ek.count > len(ek.keys)+CompressThresh {
+			ek.RLock()
+			needRewrite := ek.count > len(ek.keys)+CompressThresh
+			ek.RUnlock()
+			if needRewrite {
 				ek.rewrite()
 			}
 		} else {
@@ -154,9 +170,14 @@ func (ek *errorKeys) compress() {
 	}
 }
 
+// rewrite compacts the AOF file to only the current set of keys. It takes
+// the full write lock, rather than a read lock, because it reassigns
+// ek.file and ek.count, both of which are also written by
+// processNextOperator() from the sync() goroutine; using a read lock here
+// would race with those writes.
 func (ek *errorKeys) rewrite() {
-	ek.RLock()
-	defer ek.RUnlock()
+	ek.Lock()
+	defer ek.Unlock()
 	count := 0
 	file, err := os.OpenFile(filepath.Join(ek.aofPath, "tmp_aof"), os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0644)
 	if err != nil {
--- a/pkg/yurthub/cachemanager/error_keys_test.go
+++ b/pkg/yurthub/cachemanager/error_keys_test.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -133,7 +134,7 @@ func TestCompress(t *testing.T) {
 	}
 	err = wait.PollUntilContextTimeout(context.TODO(), time.Second, time.Minute, false,
 		func(ctx context.Context) (bool, error) {
-			if keys.count == 50 {
+			if keys.errorCount() == 50 {
 				return true, nil
 			}
 			return false, nil
@@ -141,4 +142,63 @@ func TestCompress(t *testing.T) {
 	if err != nil {
 		t.Errorf("failed to sync")
 	}
+}
+
+// TestErrorKeysConcurrentAccessRace drives put()/del() (which feed the
+// sync() goroutine started by NewErrorKeys) concurrently with direct calls
+// to rewrite() and with a reader that repeats compress()'s threshold check.
+// This is the same access pattern compress()'s 30-second ticker eventually
+// produces in production; compress() itself isn't invoked here because its
+// hardcoded ticker makes it impractical to trigger from a fast unit test.
+//
+// Before the fix, sync() wrote ek.count/ek.file with no lock at all while
+// compress()'s check read them unlocked and rewrite() reassigned them under
+// only a read lock, so this test reliably tripped `go test -race`.
+func TestErrorKeysConcurrentAccessRace(t *testing.T) {
+	aofPath, err := os.MkdirTemp("", "errorkeys")
+	if err != nil {
+		t.Fatalf("failed to create dir: %v", err)
+	}
+	defer os.RemoveAll(aofPath)
+
+	ek := NewErrorKeys(aofPath)
+	defer ek.queue.ShutDown()
+
+	var workers sync.WaitGroup
+	workers.Add(2)
+
+	go func() {
+		defer workers.Done()
+		for i := 0; i < 500; i++ {
+			key := fmt.Sprintf("key-%d", i)
+			ek.put(key, "value")
+			ek.del(key)
+		}
+	}()
+
+	go func() {
+		defer workers.Done()
+		for i := 0; i < 5; i++ {
+			ek.rewrite()
+		}
+	}()
+
+	stop := make(chan struct{})
+	var reader sync.WaitGroup
+	reader.Add(1)
+	go func() {
+		defer reader.Done()
+		for {
+			select {
+			case <-stop:
+				return
+			default:
+				_ = ek.errorCount() > ek.length()+CompressThresh
+			}
+		}
+	}()
+
+	workers.Wait()
+	close(stop)
+	reader.Wait()
 }


Aryanbhargava18:
The shift to a full write lock in `rewrite()` correctly eliminates the torn state against the background `sync()` goroutine. 

However, because `rewrite()` now holds `ek.Lock()` for the duration of the entire AOF disk write, this introduces a new latency edge-case: any concurrent `put()` or `del()` operations will block until the IO completes. Have we benchmarked the max blocking duration for `rewrite()` on slower edge node disks under high churn?

7se7en72025:
Thanks for flagging it, and sorry but I don't think the blocking window actually changes here.

put() and del() take the write lock, and rewrite() already held a read lock across the same disk write. A read lock and a write lock are already mutually exclusive, so those calls were blocked for that whole duration before this PR too. Switching to Lock() doesn't widen that.

What it does change is that rewrite() now also excludes the sync() goroutine's writes and compress()'s read, which were the actual unsynced accesses causing the race. That's the point of the fix. And it's needed for correctness anyway since rewrite() reassigns ek.file/ek.count, so the old RLock wasn't really protecting those.

The "holding a lock across disk IO on slow disks" worry is fair, but it's pre-existing and applied to the old RLock version just as much. If we want to shrink it we could snapshot under the lock and write outside it, but that's a separate optimization from this race fix. Happy to open a follow-up for it if you think it's worth benchmarking

Aryanbhargava18 (approved):
You're completely right—I over-indexed on the read lock and forgot it already fully excluded the write lock on the `put`/`del` paths. The blocking window is indeed unchanged. Great catch!

And I agree, snapshotting under the lock and doing the disk IO outside of it would be a fantastic follow-up optimization for high-churn edge nodes, but you're absolutely right that it's out of scope for a pure race-fix PR.

I'll go ahead and approve this since the race condition is properly locked down now. Thanks for the detailed explanation!

Aryanbhargava18 (changes requested):
Hold on, looking closer at the full diff, you focused the latency argument entirely on `rewrite()` and completely missed the regression in `processNextOperator()`.

Look at what this PR introduces in `processNextOperator()`:
```go
+	ek.Lock()
 	ek.file.Write(append(data, 
))
 	ek.file.Sync()
 	ek.count++
+	ek.Unlock()
```
Before this PR, `processNextOperator()` performed its disk I/O completely unlocked. Now, you have wrapped the entire AOF append and `fsync` inside the global `ek.Lock()`.

`processNextOperator()` runs in a continuous loop pulling from the work queue. `put()` and `del()` are called on the hot path for API proxying. By wrapping `ek.file.Sync()` in a global write lock, **every single proxy request that touches an error key will now synchronously block on disk IO**. OpenYurt is specifically designed for edge nodes with low IOPS (e.g., SD cards). This will cause catastrophic latency spikes across the entire hub during high churn.

You cannot wrap `fsync` in a global mutex that blocks the hot path just to fix a data race on `ek.count`. The correct fix here is to isolate the lock to strictly the memory mutations, or use `atomic`, leaving the actual disk IO entirely outside the critical section.

PR #2627, opened 2026-07-22
fix: resolve data race on ek.file and ek.count in error_keys.go

### LINKED ISSUE #2619: bug: Data race on ek.file and ek.count between sync and compress goroutines
---
name: Bug Report
title: "[BUG]"
labels: kind/bug

---

**What happened**:

Two goroutines started in `NewErrorKeys` — `go ek.sync()` and `go ek.compress()` — concurrently access `ek.file` and `ek.count` without proper mutual exclusion, constituting a data race.

`processNextOperator()` (called by `sync()`) accesses `ek.file` and `ek.count` with **no lock held at all**:

```go
// https://github.com/openyurtio/openyurt/blob/b1b7daf710a84782463ffae539e72313fb6b354b/pkg/yurthub/cachemanager/error_keys.go#L138
138:	ek.file.Write(append(data, '\n'))
139:	ek.file.Sync()
140:	ek.count++
```

`rewrite()` (called by `compress()`) closes and reassigns `ek.file` and writes `ek.count` while holding only `ek.RLock()` — a shared read lock:

```go
157: func (ek *errorKeys) rewrite() {
158: 	ek.RLock()
159: 	defer ek.RUnlock()
...
195:	ek.file.Close()
...
208:	ek.file = file
209:	ek.count = count
```

`ek.RLock()` in `rewrite()` only blocks callers of `ek.Lock()` (write lock). Because `processNextOperator()` acquires **no lock variant**, `ek.RLock()` in `rewrite()` provides zero synchronization against it. Both goroutines have an unconstrained concurrent path to `ek.file` and `ek.count`.

Concrete consequences observable from the code:

1. `processNextOperator` can call `ek.file.Write()` on the exact `*os.File` pointer that `rewrite()` is concurrently closing at line 195, resulting in a write on a closed file descriptor.
2. `ek.count++` in `processNextOperator` races with `ek.count = count` in `rewrite()`, so the compress threshold check at line 148 (`ek.count > len(ek.keys)+CompressThresh`) can observe a stale or torn value.

Running the Go race detector against this package surfaces DATA RACE reports on both `ek.file` and `ek.count`.

Accesses to `ek.file` and `ek.count` from both the sync and compress goroutines should be serialized so that `rewrite()` cannot close or reassign `ek.file` while `processNextOperator()` is writing to it, and so that increments and assignments to `ek.count` do not race.

1. Clone the repository at commit `b1b7daf7` or later.
2. Run the cache manager tests with the race detector enabled:
   ```
   go test -race ./pkg/yurthub/cachemanager/...
   ```
3. Alternatively, read the source directly:
   - `NewErrorKeys` (around line 66) starts `go ek.sync()` and `go ek.compress()`.
   - Trace `sync()` → `processNextOperator()`: no lock is acquired before `ek.file.Write` (line 138), `ek.file.Sync` (line 139), or `ek.count++` (line 140).
   - Trace `compress()` → `rewrite()`: `ek.RLock()` is acquired, then `ek.file.Close()` (line 195), `ek.file = file` (line 208), and `ek.count = count` (line 209) are executed under that read lock.
   - Because `processNextOperator` holds no lock, `ek.RLock()` in `rewrite()` does not exclude it.

Relevant permalink: https://github.com/openyurtio/openyurt/blob/b1b7daf710a84782463ffae539e72313fb6b354b/pkg/yurthub/cachemanager/error_keys.go#L138

The race exists structurally in the source and is reproducible purely by reading the locking protocol: `processNextOperator` holds no lock, and `rewrite` holds only a read lock. This means no synchronization exists between them on the shared fields `ek.file` and `ek.count`. The `compress()` goroutine also reads `ek.count` at line 148 with no lock held, which is an additional unsynchronized read of the same field.

**Environment**:
- OpenYurt version: commit b1b7daf7
- Install tools: No response

**others**

/kind bug

WorrierKhushal:
#### What type of PR is this?
/kind bug

`NewErrorKeys()` in `pkg/yurthub/cachemanager/error_keys.go` starts two 
concurrent goroutines, `sync()` and `compress()`, that both access 
`ek.file` and `ek.count` without proper mutual exclusion.

- `processNextOperator()` (called by `sync()`) wrote to `ek.file` and 
  incremented `ek.count` with no lock held at all.
- `rewrite()` (called by `compress()`) closed/reassigned `ek.file` and set 
  `ek.count` while holding only a read lock (`ek.RLock()`), which provides 
  no mutual exclusion against a caller holding no lock.
- `compress()` also read `ek.count` and `len(ek.keys)` unsynchronized 
  before calling `rewrite()`.

This was confirmed with `go test -race ./pkg/yurthub/cachemanager/...`, 
which reported a data race between these goroutines. A concurrent write on 
a closed file descriptor and a torn read of `ek.count` were both possible.

This PR fixes the synchronization:
- `processNextOperator()`: wrapped file writes and count increment under 
  `ek.Lock()`/`ek.Unlock()`
- `rewrite()`: changed `ek.RLock()`/`ek.RUnlock()` to `ek.Lock()`/`ek.Unlock()` 
  since it mutates shared state
- `compress()`: protected the `ek.count > len(ek.keys)+CompressThresh` check 
  under `ek.RLock()`/`ek.RUnlock()`, evaluated into a local variable before 
  calling `rewrite()` to avoid holding the lock across the call and prevent 
  deadlock

No exported function signatures were changed. No new comments were added. 
This is a pure concurrency-correctness fix with no behavior change.

Fixes #2619

- `go test -v ./pkg/yurthub/cachemanager/...` — all existing tests pass
- `go test -race ./pkg/yurthub/cachemanager/...` — zero data races reported

```release-note
NONE
```

#### other Note


WorrierKhushal:
Hi maintainers @wawlian @kadisi @Congrool @YTGhost , this PR is ready for review.

Summary: fixes a confirmed data race in 
`pkg/yurthub/cachemanager/error_keys.go` between `sync()` and `compress()` 
goroutines, verified with `go test -race`. Details are in the PR 
description above.

- SonarQube Quality Gate: passed ✅
- `go test -v` and `go test -race` on the affected package: passing locally ✅

This is my first contribution to OpenYurt — happy to make any changes 
requested. Could someone please take a look and approve the pending 
workflows so this can move forward? Thank you!

deepak0x:
hi @maintainers 
this pr has been raise for my issue without asking me 
pls take respective action..
as it is not an good practice 
Thanks

PR #2628, opened 2026-07-22
Bugfix: skip invalid and tmp keys in ListResourceKeysOfComponent

### LINKED ISSUE #2624: `[BUG] ListResourceKeysOfComponent returns nil keys and includes tmp backup files`
**What happened**:

In `pkg/yurthub/storage/disk/storage.go`, `ListResourceKeysOfComponent` pre-allocates a slice with `make([]storage.Key, len(files))`. When `extractInfoFromPath` fails for a file (for example a too-deep or corrupt cache path), the loop `continue`s without setting `keys[i]`, leaving **nil** entries in the returned slice.

The function also does not skip `tmp_*` backup files, unlike `Recover()` which uses `isTmpFile()` to ignore them.

Downstream, `gcPodsWhenRestart` in `pkg/yurthub/gc/gc.go` treats every listed key as a real pod. A nil key is not in `currentPodKeys`, gets added to `deletedPods`, and when delete fails the error log calls `key.Key()` on nil, which **panics** YurtHub on restart.

`ListResourceKeysOfComponent` should return only valid, non-nil keys for real cached resources. Invalid paths should be skipped (and optionally logged). `tmp_*` backup files should not appear in the result, consistent with `Recover()`.

1. Create a disk store and write three files under the kubelet pods cache path:
   - Valid: `kubelet/pods.v1.core/default/pod-a`
   - Invalid (too deep): `kubelet/pods.v1.core/default/extra/pod-b`
   - Tmp backup: `kubelet/pods.v1.core/default/tmp_pod-c`

2. Call `ListResourceKeysOfComponent("kubelet", schema.GroupVersionResource{Version: "v1", Resource: "pods"})`.

3. Observe:
   - `len(keys) == 3`
   - `keys[0] == nil` (invalid path slot)
   - `keys[2]` points to `tmp_pod-c` (backup file included)
   - Calling `key.Key()` on the nil entry panics

Minimal repro program:

```go
package main

import (
	"fmt"
	"os"
	"path/filepath"

	"k8s.io/apimachinery/pkg/runtime/schema"
	"github.com/openyurtio/openyurt/pkg/yurthub/storage/disk"
)

func main() {
	dir, _ := os.MkdirTemp("", "nil-keys-")
	defer os.RemoveAll(dir)
	store, _ := disk.NewDiskStorage(dir)

	for _, p := range []string{
		filepath.Join(dir, "kubelet", "pods.v1.core", "default", "pod-a"),
		filepath.Join(dir, "kubelet", "pods.v1.core", "default", "extra", "pod-b"),
		filepath.Join(dir, "kubelet", "pods.v1.core", "default", "tmp_pod-c"),
	} {
		os.MkdirAll(filepath.Dir(p), 0755)
		os.WriteFile(p, []byte("{}"), 0644)
	}

	keys, _ := store.ListResourceKeysOfComponent("kubelet",
		schema.GroupVersionResource{Version: "v1", Resource: "pods"})
	for i, k := range keys {
		fmt.Printf("keys[%d] = %v\n", i, k)
	}
}
```

Verified on upstream master at commit `81ca6625`.

- Root cause: `storage.go` lines 311-330
- GC impact: `gc.go` lines 154-170 (`gcPodsWhenRestart`)
- `Recover()` already skips `tmp_*` via `isTmpFile()` at line 481; `ListResourceKeysOfComponent` does not
- Likely triggered after interrupted `ReplaceComponentList` or corrupt cache layout on disk
- No existing GitHub issue found for this bug

**Environment**:
- OpenYurt version: master (`81ca6625`)
- Kubernetes version: N/A (reproduced with unit-level disk storage test)
- OS: Linux
- Kernel: N/A
- Install tools: Go 1.24+
- Others: Reproduced with standalone Go program against local clone synced with upstream master

**others**

/kind bug

Jay2006sawant:
#### What type of PR is this?
/kind bug

Fixes a bug where `ListResourceKeysOfComponent` could return nil entries and include `tmp_*` backup files in the key list.

The function pre-allocated `keys := make([]storage.Key, len(files))` and used `continue` on invalid paths without setting `keys[i]`, leaving nil slots in the result. It also did not skip `tmp_*` files, unlike `Recover()` which uses `isTmpFile()`.

On YurtHub restart, `gcPodsWhenRestart` treated nil keys as deleted pods. Failed deletes logged via `key.Key()` on nil, which could panic YurtHub when the disk cache had corrupt or interrupted writes.

This PR:
- Builds the key list with `append` for valid paths only
- Skips `tmp_*` backup files
- Skips invalid paths and KeyFunc failures
- Adds nil-key guards in pod and event GC as defense in depth

Fixes #2624 

- Main fix in `pkg/yurthub/storage/disk/storage.go` (`ListResourceKeysOfComponent`)
- GC hardening in `pkg/yurthub/gc/gc.go` (skip nil keys before compare/delete/log)
- New tests:
  - `TestListResourceKeysSkipsInvalidAndTmpFiles` (disk)
  - Ginkgo: `should skip invalid paths and tmp backup files`
  - `TestGCPodsWhenRestartWithCorruptCachePaths` (new `gc_test.go`)

**Verification performed locally:**
```bash
go test ./pkg/yurthub/storage/disk/... -count=1 -v
go test ./pkg/yurthub/gc/... -count=1 -v
go test ./pkg/yurthub/storage/... ./pkg/yurthub/cachemanager/... -count=1
```

Before fix: 3 files on disk returned `len(keys)=3` with `keys[0]=nil`.  
After fix: returns `len(keys)=1` with only the valid pod key.

```release-note
NONE


zyjhtangtang:
@Jay2006sawant Please resolve the conflict.

Jay2006sawant:
@zyjhtangtang Have resolved the conflict,PTAL.Thanks!

zyjhtangtang (approved):

PR #2631, opened 2026-07-23
fix: data race on ek.file and ek.count between sync and compress goroutines

### LINKED ISSUE #2619: bug: Data race on ek.file and ek.count between sync and compress goroutines

deepak0x:
#### What type of PR is this?

/kind bug

`NewErrorKeys` starts two goroutines, `sync()` and `compress()`, that both touch `ek.file` and `ek.count`, but access to those fields is not serialized:

- `processNextOperator()` (from `sync()`) writes `ek.file` and does `ek.count++` with no lock held.
- `rewrite()` (from `compress()`) closes and reassigns `ek.file` and sets `ek.count` while holding only a read lock (`RLock`). Since the sync side holds no lock at all, that read lock excludes nothing.
- `compress()` also reads `ek.count` and `len(ek.keys)` with no lock before deciding to compact.

So the sync goroutine can write to the exact `*os.File` that `rewrite()` is closing, and `ek.count++` races with `ek.count = count`. Running the race detector confirms it:

```
$ go test -race ./pkg/yurthub/cachemanager/...
WARNING: DATA RACE
Write at ... by goroutine (sync):
  pkg/yurthub/cachemanager/error_keys.go:140   // ek.count++
Previous read at ... by goroutine (compress)
```

This PR serializes those fields with a dedicated `fileMu`:

- `processNextOperator()` writes the file and increments the count under `fileMu`.
- The file swap in `rewrite()` moves into `swapAOF()`, which closes, renames, reopens and reassigns `ek.file`/`ek.count` under `fileMu`.
- `compress()` now reads the count through `getCount()` and the key count through the existing `length()`, both properly locked.

A note on why a separate mutex rather than just promoting the existing `RLock` to `Lock` in `rewrite()`: `rewrite()` waits up to a minute for the queue to drain, and the queue only drains because `processNextOperator()` keeps running. If `rewrite()` held a lock that `processNextOperator()` also needed, the drain could never make progress and `rewrite()` would block until it timed out. `fileMu` is only taken around the individual file/count operations and is never held across the wait, so the sync goroutine keeps draining while `rewrite()` waits. The keys snapshot in `rewrite()` keeps its original read lock, so writer-blocking behavior during compaction is unchanged. No exported signatures change.

Fixes #2619

- `go test -race ./pkg/yurthub/cachemanager/...` reports zero data races. Against the current code it fails on `ek.count` and `ek.keys`.
- `go test ./pkg/yurthub/cachemanager/...` all existing tests pass.
- Added `TestConcurrentSyncAndRewrite`, which runs `put`/sync and `rewrite()` concurrently under the race detector to guard this path. `TestCompress` was reading `count` without a lock, so it now reads through `getCount()`.

```release-note
Fix a data race between the yurthub cache manager persistency sync and compress goroutines on the AOF file handle and operation counter.
```

#### other Note

### DIFF
--- a/pkg/yurthub/cachemanager/error_keys.go
+++ b/pkg/yurthub/cachemanager/error_keys.go
@@ -41,6 +41,7 @@ type errorKeys struct {
 	sync.RWMutex
 	keys    map[string]string
 	queue   workqueue.TypedRateLimitingInterface[operation]
+	fileMu  sync.Mutex
 	file    *os.File
 	count   int
 	aofPath string
@@ -135,17 +136,25 @@ func (ek *errorKeys) processNextOperator() bool {
 		klog.Errorf("failed to serialize and persist operation: %v", op)
 		return false
 	}
+	ek.fileMu.Lock()
 	ek.file.Write(append(data, '\n'))
 	ek.file.Sync()
 	ek.count++
+	ek.fileMu.Unlock()
 	return true
+}
+
+func (ek *errorKeys) getCount() int {
+	ek.fileMu.Lock()
+	defer ek.fileMu.Unlock()
+	return ek.count
 }
 
 func (ek *errorKeys) compress() {
 	ticker := time.NewTicker(30 * time.Second)
 	for range ticker.C {
 		if !ek.queue.ShuttingDown() {
-			if ek.count > len(ek.keys)+CompressThresh {
+			if ek.getCount() > ek.length()+CompressThresh {
 				ek.rewrite()
 			}
 		} else {
@@ -192,13 +201,19 @@ func (ek *errorKeys) rewrite() {
 		klog.Errorf("failed to wait for queue to be empty")
 		return
 	}
+	ek.swapAOF(count)
+}
+
+func (ek *errorKeys) swapAOF(count int) {
+	ek.fileMu.Lock()
+	defer ek.fileMu.Unlock()
 	ek.file.Close()
 
-	err = os.Rename(filepath.Join(ek.aofPath, "tmp_aof"), filepath.Join(ek.aofPath, "aof"))
+	err := os.Rename(filepath.Join(ek.aofPath, "tmp_aof"), filepath.Join(ek.aofPath, "aof"))
 	if err != nil {
 		klog.Errorf("failed to rename tmp_aof to aof, %v", err)
 	}
-	file, err = os.OpenFile(filepath.Join(ek.aofPath, "aof"), os.O_RDWR, 0644)
+	file, err := os.OpenFile(filepath.Join(ek.aofPath, "aof"), os.O_RDWR, 0644)
 	if err != nil {
 		klog.ErrorS(err, "failed to open file", "name", filepath.Join(ek.aofPath, "aof"))
 		metrics.Metrics.SetErrorKeysPersistencyStatus(0)
--- a/pkg/yurthub/cachemanager/error_keys.go
+++ b/pkg/yurthub/cachemanager/error_keys.go
@@ -137,10 +137,10 @@ func (ek *errorKeys) processNextOperator() bool {
 		return false
 	}
 	ek.fileMu.Lock()
+	defer ek.fileMu.Unlock()
 	ek.file.Write(append(data, '\n'))
 	ek.file.Sync()
 	ek.count++
-	ek.fileMu.Unlock()
 	return true
 }
 
--- a/pkg/yurthub/cachemanager/error_keys.go
+++ b/pkg/yurthub/cachemanager/error_keys.go
@@ -152,6 +152,7 @@ func (ek *errorKeys) getCount() int {
 
 func (ek *errorKeys) compress() {
 	ticker := time.NewTicker(30 * time.Second)
+	defer ticker.Stop()
 	for range ticker.C {
 		if !ek.queue.ShuttingDown() {
 			if ek.getCount() > ek.length()+CompressThresh {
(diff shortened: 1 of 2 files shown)


Aryanbhargava18:
Hi @deepak0x, great catch on the Go data race warning! The `sync.Mutex` correctly resolves the immediate panic between the `sync` and `compress` goroutines.

However, while auditing this fix, I realized the underlying architecture suffers from a severe **logical race condition that leads to persistent data loss** during AOF rewrites, which this mutex doesn't cover.

**The Bug:**
If a new `put()` occurs while `rewrite()` is polling the queue (after it has already snapshot the map into `tmp_aof`), the new operation is written to the *old* `aof` file. Once the queue drains, `rewrite()` immediately overwrites the active `aof` with `tmp_aof`. 
**Result:** The new key is permanently deleted from disk.

**Why the Mutex isn't enough:**
`fileMu` serializes the I/O, but it doesn't synchronize the in-memory map snapshot with the queue consumption. `tmp_aof` remains completely unaware of any writes executed by `processNextOperator()` while `rewrite()` is waiting.

**The Fix:**
To safely rewrite an AOF without dropping keys, we shouldn't poll the queue length. Instead, we need a staging buffer:
1. Snapshot the map to `tmp_aof`.
2. Temporarily redirect new writes to an in-memory buffer (or write to both files).
3. Flush the buffer to `tmp_aof`, then `os.Rename` and swap descriptors under `fileMu`.

Fixing the thread panic is a great first step, but we should definitely address this silent data loss before edge nodes rely on this cache manager. Happy to collaborate on an AOF rewrite buffer if you're open to it!

zyjhtangtang:
@deepak0x   When swapping the AOF file, the queue is already empty, so there should be no file write operations at this point. Why do we still need to lock the AOF file?

deepak0x:
@zyjhtangtang Good question — the queue-empty wait alone does **not** make the swap safe, because it only drains *pending* queue items, not the *in-flight* write.

The race is on two fields the wait does not cover:

- `processNextOperator()` (sync) does `queue.Get()` → `file.Write()` → `count++` as one span. The queue can read `Len()==0` **between** `Get()` returning the last op and the goroutine reaching `file.Write()`. So `rewrite()`s poll sees an empty queue, proceeds to `swapAOF()`, and closes/reopens `ek.file` *while the sync goroutine is mid-write* to that exact handle — that is the `DATA RACE` on `ek.file` at lines 138–140. The queue check excludes nothing here because the op is already dequeued.
- `ek.count++` (sync) also races with `ek.count = count` in `swapAOF()` regardless of queue length.

The `RLock` in `rewrite()` only protects the `ek.keys` snapshot used to rebuild `tmp_aof`; the sync side never took it, so it serialized nothing on the file/count path.

`fileMu` is the minimal lock around the actual file-handle + count mutation so the swap cannot collide with an in-flight sync write. It is never held across the 1-minute drain wait, so the sync goroutine keeps draining while `rewrite()` waits (no deadlock). I can add a short comment in `swapAOF()` clarifying this if it helps.

deepak0x:
@Copilot Replying to your review on error_keys.go: addressed in cfabe7b. Two real issues here: (1) the swapped AOF is now reopened with os.O_APPEND so post-compaction sync writes append at end-of-file instead of overwriting the compacted content from offset 0; (2) on os.Rename failure, swapAOF now restores the unchanged aof handle (reopened with O_APPEND) and returns early without touching ek.count, so the count stays in sync with the on-disk file rather than being set from the tmp file. Thanks for catching both.

deepak0x:
@Copilot Replying to your second review: processNextOperator already uses `defer ek.fileMu.Unlock()` (see the current diff), so the critical section is already panic/early-return safe. No change needed there.

deepak0x:
@Copilot Replying to your third review on compress(): addressed in cfabe7b — added `defer ticker.Stop()` so the 30s ticker is released when the goroutine returns on shutdown instead of leaking the timer.

zyjhtangtang (approved):

PR #2626, opened 2026-07-22
test: improve unit test coverage for daemonpodupdater kubernetes package to >93%

alokkumardalei-wq:
### Description
Fixes #2606
This PR addresses issue #2606 by significantly improving the unit test coverage for the `daemonpodupdater/kubernetes` sub-package. Specifically, it introduces comprehensive table-driven tests for `controller_utils.go` which historically sat around 57% coverage.

### Key Additions:
* Tests for all expectation logic (`ExpectCreations`, `RaiseExpectations`, `SatisfiedExpectations`, etc.)
* Coverage for `FakePodControl` internal state and limit handling.
* Edge case testing for `RealPodControl` during API server failures (Patch/Delete/Create API failures).
* Edge cases for `validateControllerRef` and `getPodsPrefix`.

### Testing Done
- `go test ./pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/` passes cleanly.
```bash
alokkumardalei@Aloks-MacBook-Air openyurt % go test ./pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/
ok      github.com/openyurtio/openyurt/pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes0.637s
```
- Verified statement coverage reached **93.6%** via `go tool cover`.

```bash
github.com/openyurtio/openyurt/pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/controller_utils.go:472:    CreatePods                      100.0%
github.com/openyurtio/openyurt/pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/controller_utils.go:476:    CreatePodsWithGenerateName      100.0%
github.com/openyurtio/openyurt/pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/controller_utils.go:492:    DeletePod                       100.0%
github.com/openyurtio/openyurt/pkg/yurtmanager/controller/daemonsetupgradestrategy/daemonpodupdater/kubernetes/controller_utils.go:502:    Clear                           100.0%
total:                                                                                                                            (statements)                     93.6%
```

PR #2634, opened 2026-07-23
fix: prevent silent cache corruption when AOF json unmarshal fails

### LINKED ISSUE #2607: [BUG] Silent cache corruption in YurtHub CacheManager when AOF JSON unmarshaling fails
### Describe the bug
While reading through the `YurtHub` cache manager code, I noticed a tricky bug in how the system rebuilds its state from the append-only-file (AOF). 

In `pkg/yurthub/cachemanager/error_keys.go`, the code loops over the AOF to decode each line using `json.Unmarshal`. However, it completely drops the error returned by the unmarshal call. 

If a node crashes mid-write, or if a line in the AOF gets corrupted for any reason, `json.Unmarshal` will obviously fail. Because we aren't catching that error, the code just silently appends an empty, zero-value `operation` struct into the cache state. This essentially corrupts the in-memory state of the cache manager without throwing any warnings or failing visibly.

Here's the exact snippet where this happens:
https://github.com/openyurtio/openyurt/blob/master/pkg/yurthub/cachemanager/error_keys.go#L225-L230

### Expected behavior
We should catch the error. If a line is corrupted, we should probably log a warning so operators know the AOF had a bad entry, skip that specific line using `continue`, and carry on reading the rest of the file so we don't inject empty structs into our memory.

### How to reproduce it
1. Run YurtHub and let it generate an `aof` file in its cache directory.
2. Manually open the `aof` file and add a line with some garbage/invalid JSON.
3. Restart YurtHub. 
4. You'll notice that YurtHub boots up perfectly fine and parses the corrupted file without throwing a single error, silently polluting its internal state with empty operations.

### Proposed Solution
We just need a simple error check inside the scanner loop:
```go
	for scanner.Scan() {
		bytes := scanner.Bytes()
		var operation operation
		if err := json.Unmarshal(bytes, &operation); err != nil {
			klog.Errorf("failed to unmarshal AOF operation, skipping corrupted entry: %v", err)
			continue
		}
		operations = append(operations, operation)
	}
```

alokkumardalei-wq:
## Description
Fixes #2607
This PR addresses issue #2607 by properly handling `json.Unmarshal` errors during YurtHub's cache state recovery. 
Previously, if a line in the AOF (Append Only File) was corrupted or malformed, the error was ignored, and an empty, zero-value `operation` was appended to the slice. This silently polluted the in-memory cache manager state. With this fix, any corrupted lines are safely caught, logged, and skipped.
## Changes made:
- Added a proper error check on json.Unmarshal(bytes, &operation) inside the recover() loop.
- If the unmarshaling fails (e.g. if the AOF file contains a corrupted or truncated line), the system will now log the error using klog.Errorf and continue to the next line rather than inserting a zero-value operation struct into memory.

## Testing Done
- `go test ./pkg/yurthub/cachemanager/...` passes cleanly.
```bash
alokkumardalei@Aloks-MacBook-Air openyurt % go test ./pkg/yurthub/cachemanager/...
ok      github.com/openyurtio/openyurt/pkg/yurthub/cachemanager (cached)
```

### DIFF
--- a/pkg/yurthub/cachemanager/error_keys.go
+++ b/pkg/yurthub/cachemanager/error_keys.go
@@ -225,7 +225,10 @@ func (ek *errorKeys) recover() {
 	for scanner.Scan() {
 		bytes := scanner.Bytes()
 		var operation operation
-		json.Unmarshal(bytes, &operation)
+		if err := json.Unmarshal(bytes, &operation); err != nil {
+			klog.Errorf("failed to unmarshal AOF operation, skipping corrupted entry: %v", err)
+			continue
+		}
 		operations = append(operations, operation)
 	}
 	for _, op := range operations {
--- a/pkg/yurthub/cachemanager/error_keys_test.go
+++ b/pkg/yurthub/cachemanager/error_keys_test.go
@@ -142,3 +142,40 @@ func TestCompress(t *testing.T) {
 		t.Errorf("failed to sync")
 	}
 }
+
+func TestRecoverCorruptedJSON(t *testing.T) {
+	aofPath, err := os.MkdirTemp("", "errorkeys")
+	if err != nil {
+		t.Errorf("failed to create dir: %v", err)
+	}
+	defer os.RemoveAll(aofPath)
+	file, err := os.OpenFile(filepath.Join(aofPath, "aof"), os.O_CREATE|os.O_RDWR, 0644)
+	if err != nil {
+		t.Errorf("failed to open file: %v", err)
+	}
+	corruptedData := []byte(`{"key":"kubelet", "val":"fail to xxx", "operator":`)
+	file.Write(corruptedData)
+	file.Write([]byte("\n"))
+	op := operation{
+		Key:      "flannel",
+		Val:      "fail to yyy",
+		Operator: PUT,
+	}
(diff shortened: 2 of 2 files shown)


alokkumardalei-wq:
@zyjhtangtang ,please have a look on this when you get time. Happy to get any kind  feedback on this :).
Thank you!

alokkumardalei-wq:
@rambohe-ch , @luc99hen and @njucjc , I think @zyjhtangtang  is busy , could you please have a look on this pr  when you get time . Happy get any kind of feedback on this . :)
Some open prs also need your review :
1. https://github.com/openyurtio/openyurt/pull/2626
2. https://github.com/openyurtio/openyurt/pull/2599
3. https://github.com/openyurtio/openyurt/pull/2642
4. https://github.com/openyurtio/openyurt/pull/2711

zyjhtangtang:
@alokkumardalei-wq  Please pay attention to UT checkpoints.

alokkumardalei-wq:
@zyjhtangtang thanks! I just added the missing unit test , please have a look when you get time :).

njucjc (approved):

zyjhtangtang (approved):

PR #2647, opened 2026-07-24
fix: remove redundant validate(oldSP) in YurtStaticSet ValidateUpdate webhook

### LINKED ISSUE #2643: [BUG] YurtStaticSet ValidateUpdate webhook re-validates oldSP and blocks resource updates
### What happened

In `pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go`, `ValidateUpdate` validates both the proposed new object (`newSP`) and the existing stored object (`oldSP`):

```go
// ValidateUpdate implements webhook.CustomValidator so a webhook will be registered for the type.
func (webhook *YurtStaticSetHandler) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
	newSP, ok := newObj.(*v1alpha1.YurtStaticSet)
	if !ok {
		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a YurtStaticSet but got a %T", newObj))
	}
	oldSP, ok := oldObj.(*v1alpha1.YurtStaticSet)
	if !ok {
		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a YurtStaticSet but got a %T", oldObj))
	}

	if err := validate(newSP); err != nil {
		return nil, err
	}

	if err := validate(oldSP); err != nil { // <-- BUG: re-validates existing stored object state
		return nil, err
	}

	return nil, nil
}
```

Running `validate(oldSP)` during update requests creates severe operational issues:

1. If an existing `YurtStaticSet` CR in the cluster was created under an older schema or has non-compliant spec fields, `validate(oldSP)` will return an invalid error on the stored object.
2. Even when an operator submits an update (`newSP`) specifically intended to correct the invalid spec or update unrelated fields, `validate(oldSP)` fails first and rejects the update request.
3. This leaves the `YurtStaticSet` resource permanently locked in an unmodifiable state unless manually deleted and recreated.

### Flow Architecture

```
[ Update Request (kubectl / controller) ]
                    │
                    ▼
        YurtStaticSet Webhook
       ValidateUpdate(oldObj, newObj)
                    │
          ┌─────────┴─────────┐
          ▼                   ▼
    validate(newSP)     validate(oldSP)  <-- Bug!
          │                   │
    (Validates proposed) (Validates stored)
          │                   │
          └─────────┬─────────┘
                    ▼
      (If stored oldSP is non-compliant)
                    │
                    ▼
     ❌ Update Rejected by Webhook
  (YurtStaticSet locked out from repairs)
```

`ValidateUpdate` should only run `validate(newSP)` to ensure the proposed new state complies with validation rules. It should not re-validate `oldSP`.

### How to reproduce it

1. Have a `YurtStaticSet` object whose `oldSP` state fails `validate()` (e.g. invalid template spec or missing required field in legacy state).
2. Attempt to update the `YurtStaticSet` to a valid state.
3. Observe the update rejection error because `validate(oldSP)` failed.

### Affected Files

- `pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go`

/kind bug
/sig apps

bhuvan-somisetty:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

`ValidateUpdate` in the `YurtStaticSet` admission webhook validates both
the incoming new object (`newSP`) **and** the existing old object
(`oldSP`). Validating the old object is incorrect:

- The old object was already accepted by the API server when it was
  created. Re-validating it on every update serves no purpose.
- If validation rules were tightened after an object was created, all
  subsequent update requests on that object will be permanently rejected
  by the webhook — even when the new spec is completely valid.
- This is inconsistent with how Kubernetes admission webhooks are
  expected to behave; only the desired state (new object) should be
  validated during an update.

This PR removes the `validate(oldSP)` call so that `ValidateUpdate` only
validates `newSP`, matching the correct semantic for admission webhooks.

**Which issue(s) this PR fixes:**

Fixes #2643

**Special notes for your reviewer:**

The change is in
`pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go`.
The corresponding test case that previously expected an error when
`oldSP` was invalid has been updated to expect success, since a valid
`newSP` should always be accepted regardless of the state of `oldSP`.

**Does this PR introduce a user-facing change?**

```release-note
Fix YurtStaticSet webhook ValidateUpdate to only validate the new object,
preventing legitimate updates from being incorrectly rejected when the
existing object does not satisfy current validation rules.
```

**Additional documentation:**

N/A

### DIFF
--- a/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go
+++ b/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation.go
@@ -53,16 +53,12 @@ func (webhook *YurtStaticSetHandler) ValidateUpdate(ctx context.Context, oldObj,
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a YurtStaticSet but got a %T", newObj))
 	}
-	oldSP, ok := oldObj.(*v1alpha1.YurtStaticSet)
+	_, ok = oldObj.(*v1alpha1.YurtStaticSet)
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a YurtStaticSet but got a %T", oldObj))
 	}
 
 	if err := validate(newSP); err != nil {
-		return nil, err
-	}
-
-	if err := validate(oldSP); err != nil {
 		return nil, err
 	}
 
--- a/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation_test.go
+++ b/pkg/yurtmanager/webhook/yurtstaticset/v1alpha1/yurtstaticset_validation_test.go
@@ -154,7 +154,7 @@ func TestYurtStaticSetHandler_ValidateUpdate(t *testing.T) {
 			errorMsg:    "StaticPodManifest is required",
 		},
 		{
-			name: "should fail when old YurtStaticSet has invalid StaticPodManifest",
+			name: "should pass when old YurtStaticSet is invalid but new one is valid",
 			newObj: &v1alpha1.YurtStaticSet{
 				Spec: v1alpha1.YurtStaticSetSpec{
 					StaticPodManifest: "newManifest",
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Hey @zyjhtangtang @rambohe-ch — could you take a look at this when you get a chance? It's a small, focused fix in the `YurtStaticSet` admission webhook. Happy to make any changes based on your feedback. Thanks!

zyjhtangtang (approved):

PR #2649, opened 2026-07-24
fix(yurtmanager): prevent ValidateUpdate and ValidateDelete from re-validating oldObj in gateway v1alpha1 webhook

### LINKED ISSUE #2648: [BUG] Gateway v1alpha1 ValidateUpdate and ValidateDelete webhooks re-validate old object and block updates/deletions
### What happened

In `pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go`, both `ValidateUpdate` and `ValidateDelete` exhibit incorrect admission control behavior:

1. **`ValidateUpdate` re-validates `oldGw`**:
   In `ValidateUpdate` (lines 55–57), `validate(oldGw)` is invoked alongside `validate(newGw)`:
   ```go
   if err := validate(oldGw); err != nil {
       return nil, err
   }
   if err := validate(newGw); err != nil {
       return nil, err
   }
   ```
   Re-validating the old object on updates is incorrect because the existing object was already admitted into etcd. If validation rules tighten or the old object has invalid fields, any subsequent update to that Gateway will be permanently blocked by the webhook.

2. **`ValidateDelete` runs full content validation on deletion**:
   In `ValidateDelete` (lines 67–73), full business-logic validation is executed:
   ```go
   func (webhook *GatewayHandler) ValidateDelete(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
       gw, ok := obj.(*v1alpha1.Gateway)
       if !ok {
           return nil, apierrors.NewBadRequest(...)
       }
       return nil, validate(gw)
   }
   ```
   Calling `validate(gw)` during deletion means that a Gateway resource with missing endpoints or invalid fields cannot be deleted by an admin or controller, causing deadlocks in resource cleanup.

Note: The `v1beta1` implementation in `pkg/yurtmanager/webhook/gateway/v1beta1/gateway_validation.go` already avoids both of these bugs (it only validates `newGw` on update and returns `nil, nil` on delete).

---

### Flow Diagram

```
Update Request:
  [Client] ---> Update Gateway ---> [ValidateUpdate Webhook]
                                            |
                                            +---> validate(oldGw)  <-- (BUG: Fails if existing object is invalid)
                                            +---> validate(newGw)

Delete Request:
  [Client] ---> Delete Gateway ---> [ValidateDelete Webhook]
                                            |
                                            +---> validate(gw)     <-- (BUG: Blocks deletion if object is invalid)
```

---

- `ValidateUpdate` should only validate the incoming `newGw` object.
- `ValidateDelete` should allow deletion of Gateway resources without requiring them to pass full validation rules.

---

### How to reproduce it

1. Create a `Gateway` (v1alpha1) resource.
2. Modify cluster/resource state such that the existing spec would fail `validate(gw)`.
3. Try to update or delete the resource — the admission webhook rejects the request.

---

### Environment

- OpenYurt version: `v1.6+` / `master`
- Component: `yurt-manager` (`gateway` v1alpha1 webhook)

bhuvan-somisetty:
#### What type of PR is this?
/kind bug

Fixes incorrect admission control behavior in the Gateway `v1alpha1` validating webhook (`pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go`).

1. **`ValidateUpdate` unnecessarily re-validates `oldObj`**:
   `ValidateUpdate` was executing `validate(oldGw)` in addition to `validate(newGw)`. Since `oldGw` represents a resource already admitted and persisted in etcd, re-validating it on every update is incorrect. If validation rules are updated or if an existing object is in an incomplete state, all future updates to that Gateway are rejected by the admission webhook — even when the incoming update payload itself is fully valid.

2. **`ValidateDelete` runs full business-logic validation on deletion**:
   `ValidateDelete` was executing `validate(gw)` before allowing deletion. This creates a resource deadlock: a Gateway resource with missing endpoints or invalid fields can neither be updated (fails `ValidateUpdate`) nor deleted (fails `ValidateDelete`), locking up controller cleanup loops.

This PR aligns `v1alpha1` with the existing `v1beta1` implementation:
- `ValidateUpdate` only validates the incoming `newGw` spec.
- `ValidateDelete` returns `nil, nil` without running business-logic content validation.

Fixes #2648

Unit tests in `gateway_validation_test.go` have been updated to explicitly cover both bug fixes:
- `ValidateUpdate`: verifies updates succeed when `oldGw` has missing/invalid endpoints but `newGw` is valid.
- `ValidateDelete`: verifies deletion succeeds for both valid and invalid Gateway objects, as well as non-Gateway types.

```release-note
Fix Gateway v1alpha1 admission webhook so that updates only validate the new object and deletions are no longer blocked by content validation.
```

### DIFF
--- a/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go
+++ b/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go
@@ -47,29 +47,17 @@ func (webhook *GatewayHandler) ValidateUpdate(ctx context.Context, oldObj, newOb
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", newObj))
 	}
-	oldGw, ok := oldObj.(*v1alpha1.Gateway)
+	_, ok = oldObj.(*v1alpha1.Gateway)
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", oldObj))
 	}
 
-	if err := validate(oldGw); err != nil {
-		return nil, err
-	}
-
-	if err := validate(newGw); err != nil {
-		return nil, err
-	}
-
-	return nil, nil
+	return nil, validate(newGw)
 }
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type.
 func (webhook *GatewayHandler) ValidateDelete(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
-	gw, ok := obj.(*v1alpha1.Gateway)
-	if !ok {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", obj))
-	}
-	return nil, validate(gw)
+	return nil, nil
 }
 
 func validate(g *v1alpha1.Gateway) error {
--- a/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation_test.go
+++ b/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation_test.go
@@ -124,10 +124,10 @@ func TestGatewayHandler_ValidateUpdate(t *testing.T) {
 			expectedErrMsg: "missing required field 'endpoints'",
 		},
 		{
-			name:           "should return error when old Gateway is invalid",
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
Hi @zyjhtangtang @rambohe-ch — please take a look at this PR when you have time. Thanks!

zyjhtangtang (approved):

PR #2650, opened 2026-07-24
fix: don't re-validate old object on update and skip validation on delete for Gateway v1alpha1 webhook

### LINKED ISSUE #2648: [BUG] Gateway v1alpha1 ValidateUpdate and ValidateDelete webhooks re-validate old object and block updates/deletions

mohammd-khalil:
#### What type of PR is this?
/kind bug

This PR fixes incorrect admission control behavior in the Gateway v1alpha1 validating webhook (`pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go`).

**Problem 1: `ValidateUpdate` re-validates the old object**

`ValidateUpdate` was calling `validate(oldGw)` in addition to `validate(newGw)`. Since `oldGw` represents an object that was already admitted and persisted in etcd, re-validating it on every update is incorrect. If validation rules are tightened later, or if the existing object somehow ends up in an invalid state, every subsequent update to that Gateway is permanently rejected by the webhook — even when the incoming update itself is valid.

**Problem 2: `ValidateDelete` runs full validation before allowing deletion**

`ValidateDelete` was calling `validate(gw)` before allowing the delete to proceed. This means a Gateway with invalid state (e.g. missing endpoints) can never be deleted, since delete requests get rejected by the same validation rules used for create/update. This creates a deadlock: the object can't be fixed (update also validates) and can't be removed either.

Fixes #2648 

- This aligns the `v1alpha1` Gateway webhook behavior with the already-correct `v1beta1` implementation.
- No changes were made to the `validate()` function itself — only to how/when it's invoked in `ValidateUpdate` and `ValidateDelete`.

```release-note
Fixed the Gateway (v1alpha1) admission webhook so that updates no longer re-validate the existing (old) object, and deletions are no longer blocked by validation errors. This resolves an issue where Gateway resources with invalid state could become permanently stuck and unable to be updated or deleted.
```

### DIFF
--- a/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go
+++ b/pkg/yurtmanager/webhook/gateway/v1alpha1/gateway_validation.go
@@ -47,29 +47,17 @@ func (webhook *GatewayHandler) ValidateUpdate(ctx context.Context, oldObj, newOb
 	if !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", newObj))
 	}
-	oldGw, ok := oldObj.(*v1alpha1.Gateway)
-	if !ok {
+	if _, ok := oldObj.(*v1alpha1.Gateway); !ok {
 		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", oldObj))
 	}
 
-	if err := validate(oldGw); err != nil {
-		return nil, err
-	}
-
-	if err := validate(newGw); err != nil {
-		return nil, err
-	}
-
-	return nil, nil
+	return nil, validate(newGw)
 }
 
 // ValidateDelete implements webhook.CustomValidator so a webhook will be registered for the type.
 func (webhook *GatewayHandler) ValidateDelete(_ context.Context, obj runtime.Object) (admission.Warnings, error) {
-	gw, ok := obj.(*v1alpha1.Gateway)
-	if !ok {
-		return nil, apierrors.NewBadRequest(fmt.Sprintf("expected a Gateway but got a %T", obj))
-	}
-	return nil, validate(gw)
+	return nil, nil
+
 }
 
 func validate(g *v1alpha1.Gateway) error {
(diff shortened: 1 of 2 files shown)


zyjhtangtang:
Duplicate of another PR. #2649

PR #2651, opened 2026-07-25
fix(yurthub): support servicetopology filter for legacy v1.Endpoints

### LINKED ISSUE #2613: servicetopology filter is registered for endpoints but never actually filters them
### What happened

While reading through the servicetopology filter I noticed the registration and the
implementation don't line up.

The filter is registered for both `endpoints` and `endpointslices` in
`cmd/yurthub/app/options/filters.go:56-57`. But `Filter()` only type-switches on
`*discovery/v1.EndpointSlice` and `*discovery/v1beta1.EndpointSlice`. Anything else,
including `v1.Endpoints`, hits the `default` case and gets returned unchanged
(`pkg/yurthub/filter/servicetopology/filter.go:131-138`). As far as I can tell nothing
else in the yurthub tree handles `v1.Endpoints` either.

### What I expected

Either service topology gets applied to the legacy `endpoints` resource too, or `endpoints`
shouldn't be registered for the filter in the first place. Right now the two disagree.

### Why it matters

A component that still reads the legacy `endpoints` resource skips service topology
completely, even though everything looks configured correctly. Traffic that's supposed to
stay inside a node or NodePool just isn't scoped, and nothing warns you.

### Reproducing it

Set up service topology (`openyurt.io/topologyKeys`) on a Service, then have a client read
`endpoints` instead of `endpointslices` through yurthub. The endpoints come back unfiltered.

### Notes

Found this by reading the current source, same tree as #2595. I haven't reproduced it on a
live multi-NodePool cluster yet, so a second pair of eyes there would be great.

/sig network
/kind bug

Aryanbhargava18:
#### What type of PR is this?
/kind bug

Fixes the behavior noted in #2613 where the `servicetopology` filter was registered for `v1.Endpoints` but the `Filter()` function ignored it, returning the object unfiltered. This PR implements the actual filtering logic for `v1.Endpoints` to ensure components reading the legacy resource are correctly scoped to their Node or NodePool.

Fixes #2613

Added unit tests in `filter_test.go` for the new `reassembleEndpoints` logic.

```release-note
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -25,6 +25,7 @@ linters:
       checks:
         - all
         - "-ST1001"
+        - "-SA1019"
   exclusions:
     generated: lax
     presets:
--- a/pkg/yurthub/filter/servicetopology/filter.go
+++ b/pkg/yurthub/filter/servicetopology/filter.go
@@ -130,7 +130,7 @@ func (stf *serviceTopologyFilter) resolveNodePoolName() string {
 
 func (stf *serviceTopologyFilter) Filter(obj runtime.Object, stopCh <-chan struct{}) runtime.Object {
 	switch v := obj.(type) {
-	case *discoveryV1beta1.EndpointSlice, *discoveryv1.EndpointSlice:
+	case *discoveryV1beta1.EndpointSlice, *discoveryv1.EndpointSlice, *v1.Endpoints:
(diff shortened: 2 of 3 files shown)


zyjhtangtang:
Endpoints have been deprecated.

PR #2669, opened 2026-07-26
fix: propagate conciliateWorkloads/conciliateYurtAppSet errors in YurtAppSet Reconcile

om7057:
#### What type of PR is this?
/kind bug

`ReconcileYurtAppSet.Reconcile` uses named return values `(res reconcile.Result, err error)`. When `conciliateWorkloads` or `conciliateYurtAppSet` fails, the error was captured into a local `nErr` instead of the named `err`, then only `res.RequeueAfter` was set before the bare `return`. That sends `(Result{RequeueAfter: 1s}, nil)` to controller-runtime, and a nil error with `RequeueAfter` set calls `Forget()` on the rate limiter, resetting exponential backoff to zero on every retry. Under a persistent failure this turns into a tight 1 second retry loop that never backs off.

Renamed both `nErr` locals to `err` so they assign the named return, matching the pattern already used by the other bare returns in the same function.

Added a regression test (`TestReconcile_PropagatesConciliateWorkloadsError`) that forces `conciliateWorkloads` to fail by using a YurtAppSet with neither `DeploymentTemplate` nor `StatefulSetTemplate` set, and asserts `Reconcile`'s error is non-nil. Confirmed it fails on the old code and passes with the fix.

Fixes #

```release-note
Fixed a bug where YurtAppSet controller retries on conciliation failure used a fixed 1 second interval instead of exponential backoff, since the error was not actually propagated to controller-runtime.
```

### DIFF
--- a/pkg/yurtmanager/controller/yurtappset/yurt_app_set_controller.go
+++ b/pkg/yurtmanager/controller/yurtappset/yurt_app_set_controller.go
@@ -258,18 +258,18 @@ func (r *ReconcileYurtAppSet) Reconcile(
 
 	// Conciliate workloads, update yas related workloads (deploy/sts)
 	// this may infect yas appdispatched/appupdated/appdeleted condition
-	expectedNps, curWorkloads, nErr := r.conciliateWorkloads(yas, expectedRevision, yasStatus)
-	if nErr != nil {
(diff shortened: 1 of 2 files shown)


om7057:
Closing this one, not going to pursue it further right now. Apologies for the noise.

PR #2672, opened 2026-07-26
fix: add mutex locking to cloudAPIServerHealthChecker to prevent data race....

### LINKED ISSUE #2655: [BUG] Data race and unsynchronized state access in cloudAPIServerHealthChecker
**What happened**:

In `pkg/yurthub/healthchecker/cloudapiserver/health_checker.go`, the 
`cloudAPIServerHealthChecker` struct embeds `sync.RWMutex` (line 44), but 
none of its member methods acquire mutex locks when reading or mutating 
shared internal state:

- `setLastNodeLease` (lines 141–175) mutates `hc.latestLease` without acquiring `hc.Lock()`.
- `getLastNodeLease` (lines 177–182) reads `hc.latestLease` and mutates `hc.latestLease.Annotations` (via `delete(...)`) without acquiring any lock.
- `getProber` (lines 184–188) mutates `hc.remoteServerIndex` (`(hc.remoteServerIndex + 1) % len(hc.remoteServers)`) without acquiring any lock.
- `IsHealthy`, `BackendIsHealthy`, and `PickOneHealthyBackend` (lines 86–113) read `hc.probers` and `hc.remoteServers` without acquiring `hc.RLock()`.

Prober callback goroutines (invoking `setLastNodeLease` / 
`getLastNodeLease`), background health-check ticker goroutines (`run` 
calling `getProber`), and external callers (yurthub proxy handlers 
calling `IsHealthy` or `PickOneHealthyBackend`) all execute concurrently 
across different goroutines. Concurrent read/write operations on 
`latestLease`, `latestLease.Annotations`, and `remoteServerIndex` cause 
data races and can trigger fatal map mutation panics under load.

`cloudAPIServerHealthChecker` methods should use its embedded 
`sync.RWMutex` — `hc.Lock()` / `hc.Unlock()` for write operations, 
`hc.RLock()` / `hc.RUnlock()` for read-only operations — to guarantee 
thread-safe access across concurrent goroutines.

Run the race detector against the yurthub cloudapiserver healthchecker package:
```bash
go test -race -v ./pkg/yurthub/healthchecker/cloudapiserver/...
```
Or spawn multiple concurrent goroutines calling `setLastNodeLease`, 
`getLastNodeLease`, `getProber`, and `IsHealthy` on a single 
`cloudAPIServerHealthChecker` instance.

Severity: High. Data races on node lease state can lead to corrupted 
lease annotations, inaccurate backend health reporting in yurthub, or 
process crashes due to concurrent map writes on lease annotations, since 
yurthub runs on every edge node and this affects its cloud-connectivity 
health-check path directly.

**Environment**:
- Install tools: N/A

**others**
/kind bug

WorrierKhushal:
#### What type of PR is this?
/kind bug

cloudAPIServerHealthChecker embeds sync.RWMutex but never actually used 
it. setLastNodeLease, getLastNodeLease, and getProber all mutate shared 
state (latestLease, latestLease.Annotations, remoteServerIndex) with no 
lock, while IsHealthy/BackendIsHealthy read probers/remoteServers 
without an RLock either.

These run on completely different goroutines — prober callbacks, the 
background health-check ticker, and external calls from yurthub's proxy 
layer — so this is a real race, not just theoretical.

Added locking:
- setLastNodeLease, getLastNodeLease, getProber -> Lock/Unlock (all mutate state)
- IsHealthy, BackendIsHealthy -> RLock/RUnlock

One thing worth flagging for review: PickOneHealthyBackend calls 
BackendIsHealthy in a loop, and BackendIsHealthy takes its own RLock. I 
deliberately did NOT add a lock in PickOneHealthyBackend itself, since 
nested RLock on the same goroutine can deadlock if a writer is queued in 
between. Let me know if there's a cleaner pattern you'd prefer here.

Added TestCloudAPIServerHealthCheckerConcurrency, which spawns multiple 
goroutines hitting setLastNodeLease/getLastNodeLease/IsHealthy/
getProber/PickOneHealthyBackend concurrently.

Fixes #2655

go test -race -v ./pkg/yurthub/healthchecker/cloudapiserver/... passes 
clean, no races reported (ran on WSL/Linux, ~50s). Also ran go vet and a 
regular go build, both clean.

```release-note
NONE
```

#### other Note

### DIFF
--- a/pkg/yurthub/healthchecker/cloudapiserver/health_checker.go
+++ b/pkg/yurthub/healthchecker/cloudapiserver/health_checker.go
@@ -84,6 +84,8 @@ func (hc *cloudAPIServerHealthChecker) RenewKubeletLeaseTime() {
 }
 
 func (hc *cloudAPIServerHealthChecker) IsHealthy() bool {
+	hc.RLock()
+	defer hc.RUnlock()
 	for _, prober := range hc.probers {
 		if prober.IsHealthy() {
 			return true
@@ -105,6 +107,8 @@ func (hc *cloudAPIServerHealthChecker) PickOneHealthyBackend() *url.URL {
 
 // BackendHealthyStatus returns the healthy stats of specified server
 func (hc *cloudAPIServerHealthChecker) BackendIsHealthy(server *url.URL) bool {
+	hc.RLock()
+	defer hc.RUnlock()
 	if prober, ok := hc.probers[server.String()]; ok {
 		return prober.IsHealthy()
 	}
@@ -139,6 +143,8 @@ func (hc *cloudAPIServerHealthChecker) run(stopCh <-chan struct{}) {
 }
 
 func (hc *cloudAPIServerHealthChecker) setLastNodeLease(lease *coordinationv1.Lease) error {
+	hc.Lock()
+	defer hc.Unlock()
 	if lease == nil {
 		return nil
 	}
@@ -175,6 +181,8 @@ func (hc *cloudAPIServerHealthChecker) setLastNodeLease(lease *coordinationv1.Le
 }
 
 func (hc *cloudAPIServerHealthChecker) getLastNodeLease() *coordinationv1.Lease {
+	hc.Lock()
+	defer hc.Unlock()
 	if hc.latestLease != nil {
 		delete(hc.latestLease.Annotations, DelegateHeartBeat)
 	}
@@ -182,6 +190,8 @@ func (hc *cloudAPIServerHealthChecker) getLastNodeLease() *coordinationv1.Lease 
 }
 
 func (hc *cloudAPIServerHealthChecker) getProber() healthchecker.BackendProber {
+	hc.Lock()
+	defer hc.Unlock()
 	prober := hc.probers[hc.remoteServers[hc.remoteServerIndex].String()]
 	hc.remoteServerIndex = (hc.remoteServerIndex + 1) % len(hc.remoteServers)
 	return prober
--- a/pkg/yurthub/healthchecker/cloudapiserver/health_checker_test.go
+++ b/pkg/yurthub/healthchecker/cloudapiserver/health_checker_test.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/openyurtio/openyurt/cmd/yurthub/app/config"
 	"github.com/openyurtio/openyurt/pkg/yurthub/cachemanager"
+	"github.com/openyurtio/openyurt/pkg/yurthub/healthchecker"
 	"github.com/openyurtio/openyurt/pkg/yurthub/storage/disk"
 	"github.com/openyurtio/openyurt/pkg/yurthub/transport"
 )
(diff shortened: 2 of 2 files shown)


WorrierKhushal:
Hi maintainers, this PR is ready for review — fixes #2655 (missing 
mutex locking in cloudAPIServerHealthChecker causing a data race).

The struct embeds sync.RWMutex but none of its methods actually used it. 
Added proper Lock/RLock across setLastNodeLease, getLastNodeLease, 
getProber, IsHealthy, and BackendIsHealthy.

One thing I'd specifically appreciate a second opinion on: PickOneHealthyBackend 
calls BackendIsHealthy in a loop, and BackendIsHealthy takes its own 
RLock. I deliberately did NOT add a separate lock in PickOneHealthyBackend 
itself, to avoid a nested-RLock deadlock risk (if a writer is queued in 
between). Let me know if there's a cleaner pattern you'd prefer here.

Verified with go test -race -v ./pkg/yurthub/healthchecker/cloudapiserver/... 
— clean run, zero races reported. Also added a concurrency test 
(TestCloudAPIServerHealthCheckerConcurrency) covering this specifically.

Would really appreciate a review whenever you get a chance — thank you!

zyjhtangtang:
@WorrierKhushal  The probes from different remote servers are executed sequentially, so why is there a concurrency issue?

WorrierKhushal:
Yes sir, thanks for the review — you're right that the probes themselves run one after another in that same loop, so there's no race between two probes.

The race is actually between that loop's goroutine and completely different goroutines that touch the same fields from outside that loop. Two places specifically:

setLastNodeLease/getLastNodeLease — these aren't only called from inside the sequential probe loop. They're also invoked as callbacks tied to lease renewal, which runs on its own timer/goroutine, separate from the health-check ticker. So while the probe loop is reading/writing latestLease, the lease renewal path can hit it at the same time.

IsHealthy/PickOneHealthyBackend — these get called from yurthub's proxy handlers, which fire on every incoming HTTP request. Those run on whatever goroutine net/http spins up per request, totally independent of the health-check loop. So if a request comes in right when the probe loop is mid-write on probers/remoteServerIndex, that's the actual race window.

So yeah, sequential within the loop, but not sequential relative to these other call paths. That's what -race was catching for me.

Sir, please let me know if this doesn't fully address your concern or if you see a gap in my reasoning — happy to dig deeper or rework this however you think is right. Really appreciate you taking the time to review this.

zyjhtangtang:
Which exact piece of code is being referred to as the callback invoked during lease renewal?

The probers list is determined at startup — where is it being updated?

WorrierKhushal:
Yes sir, — you're right that the probes themselves run one after another in that same loop, so there's no race between two probes.

The race is actually between that loop's goroutine and completely different goroutines that touch the same fields from outside that loop. Two places specifically:

First, setLastNodeLease and getLastNodeLease aren't only called from inside the sequential probe loop. They're also invoked as callbacks tied to lease renewal, which runs on its own timer/goroutine, separate from the health-check ticker. So while the probe loop is reading or writing latestLease, the lease renewal path can hit it at the same time.

Second, IsHealthy and PickOneHealthyBackend get called from yurthub's proxy handlers, which fire on every incoming HTTP request. Those run on whatever goroutine net/http spins up per request, totally independent of the health-check loop. So if a request comes in right when the probe loop is mid-write on remoteServerIndex, that's the actual race window.

So  it's sequential within the loop, but not sequential relative to these other call paths. That's what -race was catching for me.

That said, since the probers map itself is read-only post-startup, I can clean up the locking granularity so we aren't locking things that don't need it, and focus the protection specifically on latestLease, remoteServerIndex, and active server states.

Sir, please let me know if you'd prefer me to push an update adjusting this or if there's a specific locking pattern you'd prefer here. 
Thanks again for your time and guidance!

zyjhtangtang:
Thanks for working on this. Based on the current call paths, I don’t think these locks are necessary.  probers  and  remoteServers  are initialized before the background loop starts and are never updated afterward; each prober already protects its own health state;  remoteServerIndex  is only accessed by the single  run()  goroutine; and  latestLease  is only read or written from  Probe() , which is also executed sequentially.  RenewKubeletLeaseTime()  only updates the prober’s renewal timestamp and does not invoke the lease callbacks. The new concurrency test directly calls private methods from multiple goroutines, creating a scenario that does not appear to occur in production. Moreover, the global write lock in  setLastNodeLease()  is held during synchronous storage I/O, which can unnecessarily block request-path health checks.

PR #2674, opened 2026-07-26
fix(yurthub): eliminate data race on addrConns map in connrotation di…

ipsitapp8:
#### What type of PR is this?
/kind bug
/sig network
In `connrotation.Dialer`, `CloseAll()` and `Close(address)` iterate over and delete elements from connection tracking maps outside of `d.mu.Lock()`. Meanwhile, `closableConn.Close()` mutates `d.addrConns[c.addr]` under `d.mu.Lock()`. Under concurrent connection activity and rotation, this causes a data race (`concurrent map iteration and map write`), which can lead to YurtHub panics/crashes.
This PR fixes the data race by snapshotting the tracked connections into local slices inside `d.mu.Lock()` and closing the connection handles outside the critical section without mutating shared maps.
Fixes #
None
```release-note
Fixed a concurrent map access data race in YurtHub connection rotation dialer.

### DIFF
--- a/pkg/yurthub/util/connrotation.go
+++ b/pkg/yurthub/util/connrotation.go
@@ -86,15 +86,17 @@ func (d *Dialer) CloseAll() {
 	d.mu.Lock()
 	addrConns := d.addrConns
 	d.addrConns = make(map[string]map[*closableConn]struct{})
+	var allConns []*closableConn
+	for _, conns := range addrConns {
+		for conn := range conns {
+			allConns = append(allConns, conn)
+		}
+	}
 	d.mu.Unlock()
 
-	for addr, conns := range addrConns {
-		for conn := range conns {
-			conn.Conn.Close()
-			delete(conns, conn)
-			metrics.Metrics.DecClosableConns(addr)
-		}
-		delete(addrConns, addr)
+	for _, conn := range allConns {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkg/yurthub/util/connrotation.go [hunk 1]
```diff
@@ -86,15 +86,17 @@ func (d *Dialer) CloseAll() {
...
-	for addr, conns := range addrConns {
+	var allConns []*closableConn
+	for _, conns := range addrConns {
 		for conn := range conns {
-			conn.Conn.Close()
-			delete(conns, conn)
-			metrics.Metrics.DecClosableConns(addr)
+			allConns = append(allConns, conn)
 		}
-		delete(addrConns, addr)
```
zyjhtangtang: Why was this delete operation removed?

zyjhtangtang:
Thanks for working on this. Based on the current synchronization flow, I don’t think the claimed concurrent map access can occur. Both  CloseAll()  and  Close(address)  detach the target map from  d.addrConns  while holding  d.mu ; after unlocking,  closableConn.Close()  only accesses the current map through  d.addrConns[c.addr] , not the detached map being iterated. Therefore, copying those entries into a slice does not appear to provide additional concurrency safety.

PR #2677, opened 2026-07-26
Bugfix: use EventsV1 API in gcEvents for events.k8s.io/v1 cache verif…

### LINKED ISSUE #2676: [BUG] yurthub gcEvents uses CoreV1 Events API to verify events.k8s.io/v1 cached events
### What happened?

YurtHub's event garbage collector lists cached keys for `events.k8s.io/v1` events but verifies whether each event still exists on the API server using `CoreV1().Events().Get()`. On modern Kubernetes clusters (1.19+), events are stored under the `events.k8s.io/v1` API. CoreV1 returns NotFound even when the event still exists in EventsV1, so gcEvents incorrectly deletes valid cached events from local storage.

### What did you expect to happen?

gcEvents should check event existence using the same API group/version as the cached resource (`events.k8s.io/v1`), with an optional fallback to CoreV1 for older clusters.

### How can we reproduce it?

1. Deploy YurtHub on a node with cached `events.k8s.io/v1` events (typical kubelet/kube-proxy event cache).
2. Ensure the cluster serves events via `events.k8s.io/v1` (default on K8s 1.19+).
3. Wait for or trigger the periodic event GC in `pkg/yurthub/gc/gc.go`.
4. Observe cached events removed from local storage even though they still exist on the API server.

**Minimal unit repro:** Create a cached `events.k8s.io/v1` event in disk storage, use a fake clientset where the event exists only in `EventsV1()` (not `CoreV1().Events()`), call `gcEvents`. The cached event is deleted incorrectly.

Relevant code:

```go
// gc.go:180-202 - lists events.k8s.io/v1 keys but checks CoreV1
localEventKeys, err := m.store.ListResourceKeysOfComponent(component, schema.GroupVersionResource{
    Group: "events.k8s.io", Version: "v1", Resource: "events",
})
_, err := kubeClient.CoreV1().Events(ns).Get(context.Background(), name, metav1.GetOptions{})
```

### Environment

- OpenYurt version: master (28eac4f4)
- Kubernetes version: 1.19+ (events.k8s.io/v1 as primary API)
- Component: yurthub GC manager (`pkg/yurthub/gc/gc.go`)

### Anything else we need to know?

There is no existing open issue or PR for this. A fix would use `EventsV1().Events().Get()` as primary with CoreV1 fallback, plus unit tests in a new `gc_test.go`.

Jay2006sawant:
#### What type of PR is this?
/kind bug

YurtHub's event garbage collector (`gcEvents`) lists cached keys for `events.k8s.io/v1` events but previously verified whether each event still exists on the apiserver using `CoreV1().Events().Get()`. On Kubernetes 1.19+, events are served primarily via `events.k8s.io/v1`. CoreV1 returns NotFound even when the event still exists in EventsV1, causing gcEvents to incorrectly delete valid cached events from local storage during periodic GC.

This PR:
- Adds `eventExistsOnAPIServer()` that checks `EventsV1().Events().Get()` first
- Falls back to `CoreV1().Events().Get()` for older clusters that only serve core/v1 Events
- Adds unit tests in `gc_test.go` covering the bug repro, CoreV1 fallback, missing-event deletion, invalid keys, and kube-proxy component

Fixes #2676 

The main regression test is `TestGcEventsPreservesEventsV1OnlyEvent`: before this fix, an event present only in EventsV1 was wrongly GC'd because CoreV1 returned NotFound. All gc package tests pass locally:

```
go test ./pkg/yurthub/gc/...
go vet ./pkg/yurthub/gc/...
```

```release-note
Fixed yurthub event GC incorrectly deleting valid cached events.k8s.io/v1 events on modern Kubernetes clusters.
```

### DIFF
--- a/pkg/yurthub/gc/gc.go
+++ b/pkg/yurthub/gc/gc.go
@@ -199,12 +199,13 @@ func (m *GCManager) gcEvents(kubeClient clientset.Interface, component string) {
 			continue
 		}
 
-		_, err := kubeClient.CoreV1().Events(ns).Get(context.Background(), name, metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
-			deletedEvents = append(deletedEvents, key)
-		} else if err != nil {
+		exists, err := eventExistsOnAPIServer(kubeClient, ns, name)
+		if err != nil {
 			klog.Errorf("could not get %s %s event for node(%s), %v", component, key.Key(), m.nodeName, err)
 			break
+		}
+		if !exists {
+			deletedEvents = append(deletedEvents, key)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


zyjhtangtang (approved):

zyjhtangtang:
@Jay2006sawant  Let's resolve the conflict.

zyjhtangtang (approved):

Jay2006sawant:
@zyjhtangtang Have resolved the conflicts.PTAL!

PR #2690, opened 2026-07-28
fix(yurtadm): resolve file descriptor leak during untar extraction

### LINKED ISSUE #2689: fix(yurtadm): resolve file descriptor leak during untar extraction
### Problem Description
In `pkg/yurtadm/util/util.go`, the `Untar` function extracts archive contents inside a `for` loop and uses `defer file.Close()` for created regular files:

```go
} else if hdr.Typeflag == tar.TypeReg {
    ...
    file, err := os.OpenFile(destFile, os.O_CREATE|os.O_RDWR, os.FileMode(hdr.Mode))
    if err != nil {
        return err
    }
    defer file.Close() // Executed on function return, not loop iteration
    if _, err := io.Copy(file, tr); err != nil {
        return err
    }
}
```

In Go, deferred functions are executed when the surrounding function exits. For tar archives with numerous files, all file descriptors remain open until `Untar` completes, which can cause file descriptor exhaustion (`too many open files` / `EMFILE`).

### Proposed Solution
Wrap the file creation and copying steps in an anonymous closure per iteration so that deferred `file.Close()` calls execute immediately as each file extraction finishes.

vedant21-oss:
Fixes #2689

### Problem Description
In `pkg/yurtadm/util/util.go`, the `Untar` function extracts archive contents inside a `for` loop and uses `defer file.Close()` for created regular files, causing all file handles to remain open until `Untar` finishes.

### Solution
Wrapped the file creation and data copying inside an anonymous closure per iteration to ensure `defer file.Close()` releases file descriptors immediately after each file is extracted.

### DIFF
--- a/pkg/yurtadm/util/util.go
+++ b/pkg/yurtadm/util/util.go
@@ -130,17 +130,22 @@ func Untar(tarFile, dest string) error {
 				return err
 			}
 		} else if hdr.Typeflag == tar.TypeReg {
-			// before OpenFile creating the file, use MkdirAll to ensure that the parent directory exists.
-			if err := os.MkdirAll(filepath.Dir(destFile), 0755); err != nil {
-				return err
-			}
-			file, err := os.OpenFile(destFile, os.O_CREATE|os.O_RDWR, os.FileMode(hdr.Mode))
-			if err != nil {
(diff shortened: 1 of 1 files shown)


zyjhtangtang:
@vedant21-oss Please add unit tests.

PR #2692, opened 2026-07-28
fix(yurtmanager): initialize Annotations before setting AdditionalNodepools in platformadmin conversion webhook

### LINKED ISSUE #2691: [BUG] PlatformAdmin ConvertFrom conversion webhook panics on nil Annotations map when NodePools has multiple entries
**What happened**:

In the `PlatformAdmin` conversion webhook (`pkg/apis/iot/v1alpha2/platformadmin_conversion.go`), `ConvertFrom` copies `ObjectMeta` straight from the stored `v1beta1.PlatformAdmin` object:

```go
func (c *PlatformAdmin) ConvertFrom(srcRaw conversion.Hub) error {
	srcRawV1beta1 := srcRaw.(*v1beta1.PlatformAdmin)
	c.ObjectMeta = srcRawV1beta1.ObjectMeta
	...
	if len(srcRawV1beta1.Spec.NodePools) > 1 {
		additionalNodePools := srcRawV1beta1.Spec.NodePools[1:]
		additionalNodePoolsJSON, err := json.Marshal(additionalNodePools)
		if err != nil {
			return err
		}
		c.Annotations["AdditionalNodepools"] = string(additionalNodePoolsJSON)
	}
	...
}
```

If the stored `v1beta1.PlatformAdmin` object has no annotations, `c.Annotations` (aliased from `srcRawV1beta1.ObjectMeta.Annotations`) is `nil`. When `Spec.NodePools` has more than one entry, the code writes directly into that nil map, which panics with `assignment to entry in nil map`.

This conversion runs whenever the API server (or a client such as `kubectl get platformadmin.v1alpha2`) requests a `PlatformAdmin` object stored as `v1beta1` be converted down to `v1alpha2` — a normal, unprivileged read path, not an edge case requiring special permissions.

For comparison, the sibling type `NodePool` guards the same pattern correctly in `pkg/apis/apps/v1alpha1/nodepool_conversion.go` / `pkg/apis/apps/v1beta1/nodepool_conversion.go` (`if src.Annotations == nil { src.Annotations = make(map[string]string) }` before writing), so `PlatformAdmin`'s conversion appears to have simply been missed when that pattern was applied elsewhere.

`ConvertFrom` should initialize `c.Annotations` before writing into it, e.g.:

```go
if c.Annotations == nil {
	c.Annotations = make(map[string]string)
}
c.Annotations["AdditionalNodepools"] = string(additionalNodePoolsJSON)
```

so that converting a multi-nodepool `PlatformAdmin` object with no annotations does not panic.

1. Create a `PlatformAdmin` object directly at `iot.openyurt.io/v1beta1` with `spec.nodePools` containing two or more entries and no `metadata.annotations` set, e.g.:
   ```yaml
   apiVersion: iot.openyurt.io/v1beta1
   kind: PlatformAdmin
   metadata:
     name: platformadmin-sample
   spec:
     nodePools:
       - nodepool-a
       - nodepool-b
     imageRegistry: openyurt
     version: v3
   ```
2. Fetch the object at `v1alpha2`, e.g. `kubectl get platformadmin.v1alpha2.iot.openyurt.io platformadmin-sample -o yaml`.
3. The conversion webhook's `ConvertFrom` is invoked, hits the nil-map write, and panics.

This is the same class of bug as the nil-map write recently fixed in the node mutating webhook's `Default()` (`pkg/yurtmanager/webhook/node/v1/node_default.go`) — a map field is copied from an upstream object without an existence guarantee, then written into without a nil check. The fix here should follow the same shape: guard with a nil-check/`make()` before the write, plus a unit test in `platformadmin_conversion_test.go` covering a multi-nodepool `PlatformAdmin` with nil annotations converting from v1beta1 to v1alpha2 without panicking.

**Environment**:
- OpenYurt version: master (current)
- Kubernetes version (use `kubectl version`): N/A (code-review finding, reproducible on any supported version)
- Install tools:
- Others: Found while auditing map-write sites in yurt-manager webhooks/conversions for the nil-map panic pattern.

**others**

/kind bug

bhuvan-somisetty:
#### What type of PR is this?
/kind bug
/sig iot

Fixes a potential runtime panic in the `PlatformAdmin` conversion webhook (`pkg/apis/iot/v1alpha2/platformadmin_conversion.go`).

`ConvertFrom` copies `ObjectMeta` straight from the stored `v1beta1.PlatformAdmin` object, then, when `Spec.NodePools` has more than one entry, writes the extra entries into `c.Annotations["AdditionalNodepools"]`:

```go
c.ObjectMeta = srcRawV1beta1.ObjectMeta
...
if len(srcRawV1beta1.Spec.NodePools) > 1 {
    additionalNodePools := srcRawV1beta1.Spec.NodePools[1:]
    additionalNodePoolsJSON, err := json.Marshal(additionalNodePools)
    if err != nil {
        return err
    }
    c.Annotations["AdditionalNodepools"] = string(additionalNodePoolsJSON) // <-- PANIC if c.Annotations is nil
}
```

If the stored `v1beta1.PlatformAdmin` has no annotations, `c.Annotations` is `nil`, and writing to a nil map panics with `panic: assignment to entry in nil map`. This conversion runs on a normal, unprivileged read path (e.g. `kubectl get platformadmin.v1alpha2.iot.openyurt.io`), so any multi-nodepool `PlatformAdmin` created without annotations triggers this every time it's read at v1alpha2.

The sibling `NodePool` conversion webhooks (`pkg/apis/apps/v1alpha1/nodepool_conversion.go`, `pkg/apis/apps/v1beta1/nodepool_conversion.go`) already guard the identical pattern with a nil-check before writing into `Annotations`; this webhook was missed when that guard was applied elsewhere.

This PR initializes `c.Annotations` before the write, mirroring the guard used in those webhooks.

Fixes #2691

- Added `platformadmin_conversion_test.go` covering `ConvertFrom` for: multiple nodepools with nil annotations (the panic case), multiple nodepools with existing annotations, and a single nodepool with nil annotations.
- All unit tests pass.

```release-note
Fix PlatformAdmin conversion webhook to initialize the Annotations map before setting the AdditionalNodepools annotation, preventing a runtime panic when converting a multi-nodepool PlatformAdmin object with no existing annotations.
```

### DIFF
--- a/pkg/apis/iot/v1alpha2/platformadmin_conversion.go
+++ b/pkg/apis/iot/v1alpha2/platformadmin_conversion.go
@@ -77,6 +77,9 @@ func (c *PlatformAdmin) ConvertFrom(srcRaw conversion.Hub) error {
 		if err != nil {
 			return err
 		}
+		if c.Annotations == nil {
+			c.Annotations = make(map[string]string)
+		}
 		c.Annotations["AdditionalNodepools"] = string(additionalNodePoolsJSON)
 	}
 	c.Spec.Platform = PlatformAdminPlatformEdgeX
--- a/pkg/apis/iot/v1alpha2/platformadmin_conversion_test.go
+++ b/pkg/apis/iot/v1alpha2/platformadmin_conversion_test.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2023 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the License);
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an AS IS BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"encoding/json"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openyurtio/openyurt/pkg/apis/iot/v1beta1"
+)
+
+func TestPlatformAdminConvertFrom(t *testing.T) {
+	cases := map[string]struct {
+		src            *v1beta1.PlatformAdmin
+		wantErr        bool
+		wantPoolName   string
+		wantAdditional []string
+	}{
+		"multiple nodepools with nil annotations": {
+			src: &v1beta1.PlatformAdmin{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "sample",
+				},
+				Spec: v1beta1.PlatformAdminSpec{
+					NodePools: []string{"nodepool-a", "nodepool-b", "nodepool-c"},
+				},
+			},
+			wantErr:        false,
+			wantPoolName:   "nodepool-a",
+			wantAdditional: []string{"nodepool-b", "nodepool-c"},
+		},
+		"multiple nodepools with existing annotations": {
+			src: &v1beta1.PlatformAdmin{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "sample",
+					Annotations: map[string]string{
+						"existing": "value",
+					},
+				},
+				Spec: v1beta1.PlatformAdminSpec{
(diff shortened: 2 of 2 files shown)


bhuvan-somisetty:
@zyjhtangtang @rambohe-ch could one of you take a look when you get a chance? Small, isolated fix for #2691 with a unit test covering the panic case.

zyjhtangtang (approved):

PR #2698, opened 2026-07-29
fix(yurtmanager): initialize Annotations and Data before updating YurtStaticSet configmap

### LINKED ISSUE #2697: [BUG] YurtStaticSet controller syncConfigMap panics with nil map write when ConfigMap's Annotations/Data is nil
### What happened

`syncConfigMap` in the YurtStaticSet controller can panic with `assignment to entry in nil map` when it updates an existing static-pod ConfigMap whose `Annotations` or `Data` field is `nil`.

```go
// pkg/yurtmanager/controller/yurtstaticset/yurt_static_set_controller.go
func (r *ReconcileYurtStaticSet) syncConfigMap(instance *appsv1alpha1.YurtStaticSet, hash, data string) error {
	cmName := util.WithConfigMapPrefix(instance.Name)
	cm := &corev1.ConfigMap{}
	if err := r.Get(context.TODO(), types.NamespacedName{Name: cmName, Namespace: instance.Namespace}, cm); err != nil {
		...
	}

	// if the hash value in the annotation of the cm does not match the latest hash, then update the data in the cm
	if cm.Annotations[StaticPodHashAnnotation] != hash {
		cm.Annotations[StaticPodHashAnnotation] = hash   // panics if cm.Annotations == nil
		cm.Data[instance.Spec.StaticPodManifest] = data  // panics if cm.Data == nil

		if err := r.Update(context.TODO(), cm, &client.UpdateOptions{}); err != nil {
			return err
		}
	}
	...
}
```

The `cm` object here comes from a live `r.Get()` call, not from the literal that `syncConfigMap` itself constructs on the "not found" path. Reading `cm.Annotations[StaticPodHashAnnotation]` on a nil map is safe in Go and just returns the zero value `""`, so the `!= hash` check passes and execution reaches the write on the next line, where it panics.

This is reachable whenever the backing ConfigMap exists but lacks an `Annotations` or `Data` map, for example:
- the ConfigMap was created/edited manually or by another controller/tool without the `StaticPodHashAnnotation` annotation and without a `data` key,
- the ConfigMap was migrated from an older OpenYurt version where this annotation/data convention didn't exist yet,
- any other actor strips annotations/data from the object (e.g. a `kubectl apply` from a stripped-down manifest).

The controller should defensively initialize `cm.Annotations` / `cm.Data` before writing to them (mirroring the guard pattern used elsewhere in the codebase, e.g. `if cm.Annotations == nil { cm.Annotations = map[string]string{} }`), instead of panicking and crashing the yurt-manager reconcile loop / pod.

### How to reproduce

1. Create a `YurtStaticSet` so the controller creates its managed ConfigMap (named via `util.WithConfigMapPrefix`).
2. Manually patch that ConfigMap to remove its `annotations` and `data` fields (e.g. `kubectl patch cm <name> --type=json -p '[{"op":"remove","path":"/metadata/annotations"},{"op":"remove","path":"/data"}]'`), or simply `kubectl replace` it with a bare ConfigMap containing only `metadata.name`/`metadata.namespace`.
3. Trigger a reconcile (e.g. update the `YurtStaticSet` spec, or wait for the next periodic sync) so `syncConfigMap` is invoked with a `hash` that differs from the (now-missing) annotation value.
4. Observe the yurt-manager process panic with `assignment to entry in nil map`.

### Environment

- OpenYurt version: master branch (current HEAD as of this report)
- File: `pkg/yurtmanager/controller/yurtstaticset/yurt_static_set_controller.go`, function `syncConfigMap`

### Additional context

This is the same class of bug as the nil-map-write panic recently fixed in the PlatformAdmin conversion webhook — a Kubernetes object fetched from the API server is assumed to always have non-nil `Annotations`/`Data` maps, but the API server does not guarantee that. I'm happy to submit a PR adding the missing nil-map guards.

bhuvan-somisetty:
#### What type of PR is this?
/kind bug

`syncConfigMap` in the YurtStaticSet controller fetches the static-pod ConfigMap via `r.Get()` and, when the stored hash annotation is out of date, writes directly into `cm.Annotations` and `cm.Data`. If that ConfigMap exists but has a nil `Annotations` or `Data` map (for example it was created/edited outside of this controller, or migrated from an older version without this annotation convention), reading the annotation is safe and returns `""`, so the hash comparison still passes, but the following write panics with `assignment to entry in nil map` and crashes the yurt-manager reconcile loop.

This PR initializes `cm.Annotations` and `cm.Data` with `make(map[string]string)` before writing to them, guarded on them being nil, matching the pattern already used elsewhere in the codebase for this same class of bug.

A regression test (`TestSyncConfigMapWithNilAnnotationsAndData`) is added that seeds a ConfigMap with no `Annotations`/`Data` and asserts `syncConfigMap` completes without panicking and persists the expected annotation/data.

Fixes #2697

This is the same bug class as the recently merged fix for the PlatformAdmin conversion webhook (#2691) — a Kubernetes object fetched from the API server cannot be assumed to have non-nil `Annotations`/`Data` maps.

```release-note
Fix a potential panic in the YurtStaticSet controller when the managed configmap has a nil Annotations or Data map.
```

### DIFF
--- a/pkg/yurtmanager/controller/yurtstaticset/yurt_static_set_controller.go
+++ b/pkg/yurtmanager/controller/yurtstaticset/yurt_static_set_controller.go
@@ -426,6 +426,12 @@ func (r *ReconcileYurtStaticSet) syncConfigMap(instance *appsv1alpha1.YurtStatic
 
 	// if the hash value in the annotation of the cm does not match the latest hash, then update the data in the cm
 	if cm.Annotations[StaticPodHashAnnotation] != hash {
+		if cm.Annotations == nil {
+			cm.Annotations = make(map[string]string)
+		}
+		if cm.Data == nil {
+			cm.Data = make(map[string]string)
+		}
 		cm.Annotations[StaticPodHashAnnotation] = hash
 		cm.Data[instance.Spec.StaticPodManifest] = data
 
--- a/pkg/apis/iot/v1alpha2/platformadmin_conversion.go
+++ b/pkg/apis/iot/v1alpha2/platformadmin_conversion.go
@@ -77,6 +77,9 @@ func (c *PlatformAdmin) ConvertFrom(srcRaw conversion.Hub) error {
 		if err != nil {
 			return err
 		}
+		if c.Annotations == nil {
+			c.Annotations = make(map[string]string)
+		}
 		c.Annotations["AdditionalNodepools"] = string(additionalNodePoolsJSON)
 	}
 	c.Spec.Platform = PlatformAdminPlatformEdgeX
--- a/pkg/apis/iot/v1alpha2/platformadmin_conversion_test.go
+++ b/pkg/apis/iot/v1alpha2/platformadmin_conversion_test.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2023 The OpenYurt Authors.
+
+Licensed under the Apache License, Version 2.0 (the License);
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an AS IS BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"encoding/json"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openyurtio/openyurt/pkg/apis/iot/v1beta1"
+)
+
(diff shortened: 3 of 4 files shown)


bhuvan-somisetty:
Hi @rambohe-ch @zyjhtangtang @luc99hen, could one of you take a look when you get a chance? Happy to make any changes needed.

pkg/yurtmanager/controller/yurtstaticset/yurt_static_set_controller.go
```diff
@@ -426,6 +426,12 @@ func (r *ReconcileYurtStaticSet) syncConfigMap(instance *appsv1alpha1.YurtStatic
 
 	// if the hash value in the annotation of the cm does not match the latest hash, then update the data in the cm
 	if cm.Annotations[StaticPodHashAnnotation] != hash {
+		if cm.Annotations == nil {
+			cm.Annotations = make(map[string]string)
+		}
```
zyjhtangtang: Annotations will not be nil here.

PR #2704, opened 2026-07-30
Bugfix: make filterWatch.Stop idempotent under concurrent calls

### LINKED ISSUE #2700: `panic: close of closed channel` in `yurthub` filterwatch multiplexer during concurrent `Stop()`
### Describe the bug
There is a fatal data race in the edge proxy (`yurthub`) multiplexer when managing filter watches. 

If multiple goroutines invoke `Stop()` on a `filterWatch` instance concurrently, both routines can bypass the `select` default branch before the channel is closed. This results in the second routine attempting to `close(f.done)` on an already-closed channel, immediately throwing a `panic: close of closed channel` and crashing the `yurthub` daemon.

Because `yurthub` is responsible for edge node autonomy and request caching, crashing this daemon severely impacts the availability and offline capabilities of edge nodes.

### Expected behavior
`filterWatch.Stop()` should be completely idempotent and thread-safe. Concurrent calls to `Stop()` should safely return without throwing a panic or double-closing the channel.

### Steps to reproduce
1. Instantiate a `filterWatch` multiplexer instance in a highly concurrent edge environment.
2. Trigger multiple rapid tear-down / disconnect events that concurrently invoke `Stop()` on the same watch instance.
3. The `yurthub` daemon panics and crashes.

### Environment
- OpenYurt Version: `master`
- Component: `yurthub` (`pkg/yurthub/multiplexer/filterwatch.go`)

akshita317:
#### What type of PR is this?

/kind bug

`filterWatch.Stop()` in `pkg/yurthub/multiplexer/filterwatch.go` guarded the close of `f.done` with a check-then-close:

```go
func (f *filterWatch) Stop() {
	select {
	case <-f.done:
	default:
		close(f.done)
		f.source.Stop()
	}
}
```

The check and the close are not atomic. Two goroutines can both observe `f.done` as still open and both take the `default` branch; the second `close(f.done)` then panics with `close of closed channel`. That panic is not recoverable by the caller and takes down the whole `yurthub` process, so an edge node loses request caching and autonomy.

This is reachable in normal operation rather than only under synthetic load — `Stop` has two independent callers on the same watch:

* `receive()` runs `defer f.Stop()`, which fires as soon as the source watch channel is drained (API server disconnect, watch expiry, source tear-down);
* the watch consumer calls `Stop()` itself when the client goes away.

A disconnect that races with a client tear-down puts both on `Stop()` at the same time.

The fix guards the body with a `sync.Once`, so `close(f.done)` and `f.source.Stop()` happen exactly once no matter how many callers arrive. `Once.Do` also blocks later callers until the first one finishes, so `Stop()` still returns only after the source is actually stopped — the behaviour the previous code intended.

Also adds `TestFilterWatch_ConcurrentStop`, which races 16 goroutines on `Stop()` across 1000 freshly created watches. Against the old code it panics on every run I tried; with the fix the package passes with `-count=3`.

Fixes #2700

The `stopOnce` field is placed on the struct rather than replacing `done`, because `done` is still needed as a signal — it is passed to `filter.Filter(newObj, f.done)` and selected on in `receive()`.

```release-note
Fix a panic ("close of closed channel") that could crash yurthub when a multiplexer filter watch was stopped concurrently by the watch consumer and by its own receive loop.
```

### DIFF
--- a/pkg/yurthub/multiplexer/filterwatch.go
+++ b/pkg/yurthub/multiplexer/filterwatch.go
@@ -17,6 +17,8 @@ limitations under the License.
 package multiplexer
 
 import (
+	"sync"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/watch"
@@ -26,19 +28,21 @@ import (
 )
 
 type filterWatch struct {
-	source watch.Interface
-	filter filter.ObjectFilter
-	result chan watch.Event
-	done   chan struct{}
+	source   watch.Interface
+	filter   filter.ObjectFilter
+	result   chan watch.Event
+	done     chan struct{}
+	stopOnce sync.Once
 }
 
+// Stop is idempotent and safe for concurrent use. It is called both by the
+// watch consumer and by receive() when the source channel is drained, so the
+// close of done must happen exactly once.
 func (f *filterWatch) Stop() {
-	select {
-	case <-f.done:
-	default:
+	f.stopOnce.Do(func() {
 		close(f.done)
 		f.source.Stop()
-	}
+	})
 }
 
 func newFilterWatch(source watch.Interface, filter filter.ObjectFilter) watch.Interface {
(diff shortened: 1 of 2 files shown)


zyjhtangtang:
#2701 https://github.com/openyurtio/openyurt/pull/2701

PR #2705, opened 2026-07-30
Bugfix: parse the kubelet EnvironmentFile path safely in GetNodeName

### LINKED ISSUE #2661: [BUG] node-servant/yurtadm GetNodeName panics or resolves wrong node name when parsing kubelet EnvironmentFile
**What happened**:

`GetNodeName()` in `pkg/yurtadm/util/edgenode/edgenode.go` parses the kubelet systemd drop-in to discover the node name. In the branch that reads the `EnvironmentFile=` directive it does:

```go
// pkg/yurtadm/util/edgenode/edgenode.go
environmentFiles, err := GetContentFormFile(kubeadmConfPath, constants.KubeletEnvironmentFile)
if err != nil {
    return "", err
}
for _, ef := range environmentFiles {
    ef = strings.Split(ef, "-")[1]   // <-- line 123
    nodeName, err = GetSingleContentFromFile(ef, constants.KubeletHostname)
    ...
}
```

with `KubeletEnvironmentFile = "EnvironmentFile=.*"`, so `ef` is the whole matched directive, e.g. `EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env`.

`strings.Split(ef, "-")[1]` is used to strip the `EnvironmentFile=-` prefix and recover the file path, but it splits on **every** `-` and unconditionally indexes `[1]`. This is wrong in two ways:

1. **Panic (`index out of range [1]`)** when the directive contains no `-` at all. systemd fully allows an `EnvironmentFile=` without the leading `-` (the `-` only means "ignore if missing"), and the path itself may contain no hyphen — e.g. `EnvironmentFile=/etc/default/kubelet`. Then `strings.Split(...)` returns a slice of length 1 and `[1]` panics, crashing the caller.

2. **Wrong path** when the file path contains a `-`. The kubeadm default is `EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env`. Splitting on `-` and taking index `[1]` yields `/var/lib/kubelet/kubeadm` (truncated at the hyphen inside `kubeadm-flags.env`). The subsequent read of that non-existent path fails, so this branch silently never finds the `--hostname-override`, and `GetNodeName` falls through to `/etc/hostname` — which can return a **different** name than the intended override.

This function is reachable in production: `node-servant convert` calls it via
`pkg/node-servant/convert/options.go:31` → `getNodeNameFunc = enutil.GetNodeName` → `options.Complete()` (`getNodeNameFunc(kubeadmConfPath)`).

`GetNodeName` should robustly extract the `EnvironmentFile` path (stripping only the optional leading `EnvironmentFile=` key and the optional systemd `-` prefix) and never panic on a valid systemd directive. For the standard kubeadm layout it should correctly resolve `/var/lib/kubelet/kubeadm-flags.env` and read `--hostname-override` from it.

Minimal reproduction of the parsing logic:

```go
package main

import (
	"fmt"
	"strings"
)

func extract(ef string) (res string, panicked bool) {
	defer func() {
		if recover() != nil {
			panicked = true
		}
	}()
	return strings.Split(ef, "-")[1], false
}

func main() {
	for _, c := range []string{
		"EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env", // kubeadm default
		"EnvironmentFile=/opt/kubelet/config.env",             // no '-' anywhere
	} {
		got, p := extract(c)
		if p {
			fmt.Printf("%-52s => PANIC (index out of range)\n", c)
		} else {
			fmt.Printf("%-52s => %q\n", c, got)
		}
	}
}
```

Output:

```
EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env  => "/var/lib/kubelet/kubeadm"      <- wrong path
EnvironmentFile=/opt/kubelet/config.env              => PANIC (index out of range)      <- crash
```

End to end: on a node whose `10-kubeadm.conf` has `EnvironmentFile=/opt/kubelet/config.env` (no hyphen) and no `--hostname-override` directly in `10-kubeadm.conf`, running `node-servant convert` panics while resolving the node name.

Suggested fix: replace the `Split(ef, "-")[1]` heuristic with an explicit, safe strip of the systemd directive — e.g. take everything after the first `=`, then trim a single optional leading `-`:

```go
path := ef
if i := strings.IndexByte(path, '='); i >= 0 {
    path = path[i+1:]
}
path = strings.TrimPrefix(strings.TrimSpace(path), "-")
```

This avoids the panic and preserves hyphens inside the path. Happy to send a PR with this fix and a table-driven test covering the cases above.

**Environment**:
- OpenYurt version: master (reproduced by reading current source)
- Kubernetes version (use `kubectl version`): N/A (client-side parsing bug in `node-servant` / `yurtadm`)
- OS: any Linux node using a systemd kubelet drop-in
- Kernel: N/A
- Install tools: kubeadm (default `EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env`)
- Others: affected code is `pkg/yurtadm/util/edgenode/edgenode.go` (`GetNodeName`, line 123)

/kind bug

akshita317:
#### What type of PR is this?

/kind bug

`GetNodeName` in `pkg/yurtadm/util/edgenode/edgenode.go` reads the kubelet systemd drop-in to discover the node name. To recover the path out of an `EnvironmentFile=` directive it did:

```go
for _, ef := range environmentFiles {
	ef = strings.Split(ef, "-")[1]
	...
}
```

`constants.KubeletEnvironmentFile` is `EnvironmentFile=.*`, so `ef` is the entire directive, e.g. `EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env`. Splitting on **every** `-` and unconditionally indexing `[1]` breaks in two ways:

1. **Wrong path.** For the kubeadm default above, the result is `/var/lib/kubelet/kubeadm` — truncated at the hyphen inside `kubeadm-flags.env`. Reading that non-existent file fails, so this branch never finds `--hostname-override`, and `GetNodeName` silently falls through to `/etc/hostname`, which can be a different name than the operator configured.
2. **Panic.** The leading `-` is optional in systemd — it only means "ignore if the file is missing" — and the path need not contain a hyphen either, e.g. `EnvironmentFile=/etc/default/kubelet`. Then `strings.Split` returns a one-element slice and `[1]` panics with `index out of range [1]`. This is reachable from `node-servant convert`, which calls `GetNodeName` via `options.Complete()` (`pkg/node-servant/convert/options.go`), so the command crashes instead of reporting an error.

This PR replaces the heuristic with a small explicit parser that drops the directive key and a single optional leading `-`, and keeps any hyphens inside the path:

```go
func parseEnvironmentFilePath(directive string) string {
	path := directive
	if i := strings.IndexByte(path, '='); i >= 0 {
		path = path[i+1:]
	}
	path = strings.TrimPrefix(strings.TrimSpace(path), "-")
	return strings.TrimSpace(path)
}
```

A directive that carries no path is now skipped rather than read as `""`.

While in the same function, the two `strings.Split(nodeName, NodeNameSplit)[1]` calls that pull the value out of `--hostname-override=<name>` became `strings.SplitN(..., 2)[1]`. `KubeletHostname` matches `[^"\s]*`, which permits `=`, so an override value containing `=` was previously truncated at the first one.

Fixes #2661

Verified both directions locally:

* `Test_parseEnvironmentFilePath` is a table test over the kubeadm default, a path with no `-` at all, a path with hyphens but no `-` prefix, trailing whitespace, and an empty directive.
* `Test_GetNodeNameFromEnvironmentFile` drives the real `GetNodeName` against a temporary drop-in + flags file for both layouts. Against the current code the first case fails (falls through past the truncated path) and the second panics; both pass with this change.

```release-note
Fix `node-servant convert` / yurtadm resolving the wrong node name — or panicking — when the kubelet drop-in's EnvironmentFile path contains a "-" or omits systemd's optional "-" prefix.
```

### DIFF
--- a/pkg/yurtadm/util/edgenode/edgenode.go
+++ b/pkg/yurtadm/util/edgenode/edgenode.go
@@ -96,6 +96,20 @@ func CopyFile(sourceFile string, destinationFile string, perm os.FileMode) error
 	return nil
 }
 
+// parseEnvironmentFilePath extracts the file path from a systemd EnvironmentFile
+// directive such as "EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env". It drops
+// the directive key and the optional leading "-" (which only tells systemd to ignore
+// a missing file), and keeps any "-" inside the path itself. An empty string is
+// returned when the directive carries no path.
+func parseEnvironmentFilePath(directive string) string {
+	path := directive
+	if i := strings.IndexByte(path, '='); i >= 0 {
+		path = path[i+1:]
+	}
+	path = strings.TrimPrefix(strings.TrimSpace(path), "-")
+	return strings.TrimSpace(path)
+}
+
 // GetNodeName gets the node name based on environment variable, parameters --hostname-override
 // in the configuration file or hostname
 func GetNodeName(kubeadmConfPath string) (string, error) {
@@ -108,7 +122,7 @@ func GetNodeName(kubeadmConfPath string) (string, error) {
 	//2. find --hostname-override in 10-kubeadm.conf
 	nodeName, err := GetSingleContentFromFile(kubeadmConfPath, constants.KubeletHostname)
 	if nodeName != "" {
-		nodeName = strings.Split(nodeName, NodeNameSplit)[1]
+		nodeName = strings.SplitN(nodeName, NodeNameSplit, 2)[1]
 		return nodeName, nil
 	} else {
 		klog.V(4).Info("get nodename err: ", err)
@@ -120,10 +134,13 @@ func GetNodeName(kubeadmConfPath string) (string, error) {
 		return "", err
 	}
 	for _, ef := range environmentFiles {
-		ef = strings.Split(ef, "-")[1]
+		ef = parseEnvironmentFilePath(ef)
+		if ef == "" {
+			continue
+		}
 		nodeName, err = GetSingleContentFromFile(ef, constants.KubeletHostname)
 		if nodeName != "" {
-			nodeName = strings.Split(nodeName, NodeNameSplit)[1]
+			nodeName = strings.SplitN(nodeName, NodeNameSplit, 2)[1]
 			return nodeName, nil
 		} else {
 			klog.V(4).Info("get nodename err: ", err)
--- a/pkg/yurtadm/util/edgenode/edgenode_test.go
+++ b/pkg/yurtadm/util/edgenode/edgenode_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package edgenode
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


zyjhtangtang:
@akshita317 Fix the conflicts.

PR #2717, opened 2026-08-02
fix: add node identity to yurt-hub-multiplexer-binding during node-se…

### LINKED ISSUE #2716: [BUG] Failure of node-servant-conversion with "context deadline exceeded"
**What happened**:
YurtHub is denied access to ConfigMap and the NodePool resource, because YurtHub's own identity was never added as a member of the openyurt:multiplexer group. Since nothing in the install/conversion flow adds node identities to the `openyurt:multiplexer` group, every YurtHub instance is denied these permissions by default, causing YurtHub to never reach a ready state. `node-servant convert` then times out waiting for YurtHub's readiness signal, and the conversion fails.

The node-servant conversion job should complete successfully, YurtHub should reach a ready state, and the node should be labeled `openyurt.io/is-edge-worker=true`.

1. Create a kind cluster with 1 control-plane + 2 worker nodes
2. Install yurt-manager: `helm upgrade --install yurt-manager -n kube-system openyurt/yurt-manager`
3. Install raven-agent: `helm upgrade --install raven-agent -n kube-system openyurt/raven-agent`
4. Create an Edge-type NodePool:
```yaml
   apiVersion: apps.openyurt.io/v1beta2
   kind: NodePool
   metadata:
     name: worker
   spec:
     type: Edge
```
5. Label a worker node: `kubectl label node <worker-node-name> apps.openyurt.io/nodepool=worker`
6. Observe the `node-servant-conversion-<node>` job fail after ~5 minutes

yurthub logs show repeated `403 Forbidden` on ConfigMap and NodePool reads for identity `system:node:<node-name>`.

**Environment**:
- OpenYurt version: yurt-manager chart 1.7.0 (app version v1.7.0), raven-agent chart 0.4.3
- Kubernetes version (use `kubectl version`): v1.36.1 (client and server)
- OS (e.g: `cat /etc/os-release`): Debian GNU/Linux 13 (trixie)
- Kernel (e.g. `uname -a`): Linux openyurt-dev-worker 6.18.33.2-microsoft-standard-WSL2 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
- Install tools: kind v0.32.0, Helm (openyurt/yurt-manager and openyurt/raven-agent charts)
- Others: Windows host, Docker Desktop with WSL2 backend

/kind bug

AyushSrivastava1818:
## Problem

During node-servant conversion, YurtHub authenticates using the kubelet's
existing client certificate (`CN=system:node:<node-name>`, `O=system:nodes`).
The static `yurt-hub-multiplexer-binding` ClusterRoleBinding grants access
only to the custom group `openyurt:multiplexer`, which is not present on
kubelet certificates and is never assigned by any dynamic mechanism.

This causes YurtHub to receive 403s on ConfigMaps/NodePools requests, and
the hub never becomes Ready after node-servant conversion.

Fixes #2716

## Fix

Added `ensureNodeInMultiplexerRBAC()` as the **first step** in
`nodeConverter.Do()` (before `installYurtHub()`). It idempotently patches
`yurt-hub-multiplexer-binding` to append a per-node User subject:
```yaml
- kind: User
  name: system:node:<node-name>
  apiGroup: rbac.authorization.k8s.io

### DIFF
--- a/charts/yurt-manager/templates/yurt-manager.yaml
+++ b/charts/yurt-manager/templates/yurt-manager.yaml
@@ -38,6 +38,35 @@ subjects:
   name: yurt-manager
   namespace: {{ .Release.Namespace }}
 ---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: node-servant
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: node-servant
+rules:
+  - apiGroups: ["rbac.authorization.k8s.io"]
+    resources: ["clusterrolebindings"]
+    resourceNames: ["yurt-hub-multiplexer-binding"]
+    verbs: ["get", "update"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-servant
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: node-servant
+subjects:
+  - kind: ServiceAccount
+    name: node-servant
+    namespace: kube-system
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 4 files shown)


zyjhtangtang:
# https://github.com/openyurtio/openyurt/blob/master/charts/yurt-manager/templates/yurthub-cfg.yaml

PR #2731, opened 2026-08-04
fix(yurthub): prevent in-memory cache drift on storage access conflic…

### LINKED ISSUE #2712: [BUG] Silent watch event drop during disk storage list-replacement causes permanent cache inconsistency between memory and disk
**What happened**:

When YurtHub processes a `LIST` response from the Cloud API Server, 
`saveListObject()` invokes `diskStorage.ReplaceComponentList()`, which 
acquires an exclusive pending lock on the target resource key hierarchy 
(e.g., `kubelet/pods/default`).

If a concurrent `WATCH` event (such as `watch.Modified` or `watch.Added`) 
for a resource under that hierarchy arrives while 
`ReplaceComponentList()` is executing disk I/O, `diskStorage.Update()` 
attempts to acquire a lock via `ds.lockKey()`. Because `lockKey()` is 
non-blocking and checks parent key prefixes, it immediately returns 
`false` (`storage.ErrStorageAccessConflict`).

In `cache_manager.go`, `storeObjectWithKey()` explicitly catches 
`storage.ErrStorageAccessConflict`, logs a debug message 
(`"skip to cache watch event because key(...) is under processing"`), 
and **swallows the error by returning `nil`**. As a result:

1. `saveWatchObject()` assumes `storeObjectWithKey()` succeeded and 
   updates `inMemoryCache` with the new watch object.
2. The disk storage file (e.g. `/etc/kubernetes/cache/kubelet/pods/default/pod-A`) 
   is **never updated on disk**.
3. If YurtHub restarts or queries disk cache during edge autonomy (cloud 
   disconnect), YurtHub reads the stale pre-watch object from disk, 
   leading to permanent cache divergence between memory and disk.

Watch events should never be silently dropped from persistent disk 
storage. When `diskStorage` is processing a `ReplaceComponentList()` 
operation, concurrent watch updates should either block/queue until the 
list replacement completes, or trigger a retry mechanism, so that 
`inMemoryCache` and local disk cache remain consistent.

1. Deploy YurtHub on an edge node with active workloads.
2. Trigger a full list sync for pods (e.g. restart Kubelet or issue a 
   `LIST` request through YurtHub).
3. Simultaneously push rapid watch events for pods in the same namespace 
   (e.g. update pod annotations/labels or patch status).
4. Inspect YurtHub logs for 
   `skip to cache watch event because key(...) is under processing`.
5. Disconnect the edge node from the cloud control plane and restart the 
   YurtHub process.
6. Issue a `GET` for the updated pod through YurtHub's local proxy; 
   observe that YurtHub serves the stale pre-watch object from disk.

**Root Cause Code Trace:**
1. `saveWatchObject()` in `pkg/yurthub/cachemanager/cache_manager.go` 
   (~line 433) receives `watch.Added` or `watch.Modified`.
2. Calls `cm.storeObjectWithKey()` in `cache_manager.go` (~line 672).
3. Delegates to `storageWrapper.Update()` in 
   `pkg/yurthub/cachemanager/storage_wrapper.go` (~line 205).
4. `diskStorage.Update()` in `pkg/yurthub/storage/disk/storage.go` 
   (~line 246) calls `ds.lockKey(key)`.
5. `lockKey()` in `storage.go` (~lines 567-577) checks 
   `keyPendingStatus`. Because `ReplaceComponentList()` holds 
   `kubelet/pods/default`, `lockKey()` returns `false` 
   (`ErrStorageAccessConflict`).
6. `storeObjectWithKey()` in `cache_manager.go` (~lines 686-689) 
   swallows `ErrStorageAccessConflict`:
```go
   case errors.Is(err, storage.ErrStorageAccessConflict):
       klog.V(2).Infof("skip to cache watch event because key(%s) is under processing", key.Key())
       return nil
```
7. `saveWatchObject()` sees `err == nil` and calls 
   `cm.updateInMemoryCache()`, updating memory but leaving disk stale.

**Proposed direction (not a full fix, just a starting point for 
discussion):** rather than returning `nil` on 
`ErrStorageAccessConflict`, `storeObjectWithKey()` or `diskStorage` 
could implement a retry/buffering queue for watch events while 
component list replacement is in progress. Alternatively, 
`ReplaceComponentList()` could merge in-flight watch updates into the 
pending list snapshot before finalizing the directory swap.

Severity: High. This affects the correctness of YurtHub's disk cache 
during edge autonomy — the exact scenario OpenYurt is designed to 
protect — meaning a disconnected edge node could serve incorrect, 
stale data to Kubelet after a restart.

**Environment**:
- OpenYurt version: master branch (v1.5.0+, commit at time of report)
- Kubernetes version (use `kubectl version`): v1.28.0+
- OS (e.g: `cat /etc/os-release`): Linux / Ubuntu 22.04
- Kernel (e.g. `uname -a`): 5.15.0-x86_64
- Install tools: yurtctl / helm
- Others:

**others**
/kind bug

WorrierKhushal:
#### What type of PR is this?
/kind bug
Fixes #2712

### 1. Architectural Bug Summary & Impact
In `pkg/yurthub/cachemanager/cache_manager.go`, when a watch event arrives while `ReplaceComponentList()` is holding a key-prefix lock on disk storage, `storeObjectWithKey()` catches `storage.ErrStorageAccessConflict`. 

Previously, `storeObjectWithKey()` swallowed `ErrStorageAccessConflict` and returned `nil`. Following that call, `saveWatchObject()` unconditionally invoked `updateInMemoryCache()` regardless of whether the disk write succeeded. 

**The Bug:** The in-memory cache was updated with the newer resource version from the watch event, but the disk write was silently dropped. This created a persistent state drift between memory and disk storage. If YurtHub restarted or an edge node disconnected, the node would fall back to stale disk data while in-memory operations served the newer version.

---

#### A. Error Propagation in `storeObjectWithKey()`
We updated `storeObjectWithKey()` to propagate `storage.ErrStorageAccessConflict` back to the caller instead of returning `nil`.

```go
// pkg/yurthub/cachemanager/cache_manager.go

func (cm *cacheManager) storeObjectWithKey(accessor meta.Object: ... ) error {
    ...
    err := cm.storageWrapper.Update(key, obj, rv)
    if err != nil {
        if errors.Is(err, storage.ErrStorageNotFound) {
            if err := cm.storageWrapper.Create(key, obj); err != nil {
                if errors.Is(err, storage.ErrStorageAccessConflict) {
                    klog.V(2).Infof("storage access conflict for %s, skip create: %v", key, err)
                    // BEFORE: return nil
                    // AFTER:
                    return err
                }
                return err
            }
            return nil
        } else if errors.Is(err, storage.ErrStorageAccessConflict) {
            klog.V(2).Infof("storage access conflict for %s, skip update: %v", key, err)
            // BEFORE: return nil
            // AFTER:
            return err
        }
        return err
    }
    return nil
}

### DIFF
--- a/pkg/yurthub/cachemanager/cache_manager.go
+++ b/pkg/yurthub/cachemanager/cache_manager.go
@@ -437,9 +437,11 @@ func (cm *cacheManager) saveWatchObject(ctx context.Context, info *apirequest.Re
 				} else {
 					updateObjCnt++
 				}
-				errMsg := cm.updateInMemoryCache(ctx, info, obj)
-				if errMsg != nil {
-					klog.Errorf("failed to update cache, %v", errMsg)
+				if err == nil {
+					errMsg := cm.updateInMemoryCache(ctx, info, obj)
+					if errMsg != nil {
+						klog.Errorf("failed to update cache, %v", errMsg)
+					}
 				}
 			case watch.Deleted:
 				err = cm.storage.Delete(key)
@@ -455,7 +457,11 @@ func (cm *cacheManager) saveWatchObject(ctx context.Context, info *apirequest.Re
 			}
 
 			if err != nil {
-				klog.Errorf("could not process watch object %s, %v", key.Key(), err)
+				if errors.Is(err, storage.ErrStorageAccessConflict) {
+					klog.V(2).Infof("could not process watch object %s, %v", key.Key(), err)
+				} else {
+					klog.Errorf("could not process watch object %s, %v", key.Key(), err)
+				}
 			}
 		case watch.Bookmark:
 			rv, _ := accessor.ResourceVersion(obj)
@@ -679,13 +685,13 @@ func (cm *cacheManager) storeObjectWithKey(key storage.Key, obj runtime.Object) 
 		if err := cm.storage.Create(key, obj); err != nil {
 			if errors.Is(err, storage.ErrStorageAccessConflict) {
 				klog.V(2).Infof("skip to cache obj because key(%s) is under processing", key.Key())
-				return nil
+				return err
 			}
 			return fmt.Errorf("could not create obj of key: %s, %v", key.Key(), err)
 		}
 	case errors.Is(err, storage.ErrStorageAccessConflict):
 		klog.V(2).Infof("skip to cache watch event because key(%s) is under processing", key.Key())
-		return nil
+		return err
 	default:
 		return fmt.Errorf("could not store obj with rv %s of key: %s, %v", newRv, key.Key(), err)
 	}
--- a/pkg/yurthub/cachemanager/cache_manager_test.go
+++ b/pkg/yurthub/cachemanager/cache_manager_test.go
@@ -26,6 +26,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -42,6 +43,7 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/openyurtio/openyurt/pkg/projectinfo"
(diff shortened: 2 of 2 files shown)


WorrierKhushal:
Hi maintainers sir @wawlian @carolove @Fei-Guo ,

I have opened a Pull Request to resolve issue #2712 regarding state divergence between YurtHub's in-memory cache and disk storage.

### Summary of Root Cause & Fix:
1. **Root Cause:** When `saveWatchObject()` hit a `storage.ErrStorageAccessConflict` during a concurrent list replace operation, `storeObjectWithKey()` swallowed the error and returned `nil`. This caused `updateInMemoryCache()` to execute anyway, caching a newer ResourceVersion in memory that was never written to disk.
2. **The Fix:**
   - Propagated `ErrStorageAccessConflict` from `storeObjectWithKey()`.
   - Wrapped `updateInMemoryCache()` in `saveWatchObject()` with an `if err == nil` check (matching `saveOneObject()`).
   - Adjusted the fallback log level in `saveWatchObject()` for conflict errors to `klog.V(2)` to avoid duplicate error log noise.
3. **Testing:** Added a fully deterministic unit test (`TestCacheWatchResponseKeepsInMemoryCacheConsistentWithDiskOnAccessConflict`) using channel synchronization to test concurrent List/Watch events without `time.Sleep`.

Please take a look when you have a chance. Thanks!

PR #2755, opened 2026-08-15
fix: stop leaked watch streams when pool-scope metadata is removed

### LINKED ISSUE #2753: [BUG] Multiplexer watch streams leak indefinitely when pool-scope metadata is removed from Leader Hub ConfigMap
**What happened**:

When the `leader-hub-{pool-name}` ConfigMap is updated to remove a resource 
from the `pool-scoped-metadata` list, the YurtHub multiplexer fails to 
terminate the underlying watch stream (reflector cache) connected to the 
cloud API Server or Leader Hub.

In `pkg/yurthub/multiplexer/multiplexer.go` (~line 195), 
`updateLeaderHubConfiguration` attempts cleanup by looking up and executing 
closure functions inside `m.lazyLoadedGVRCacheDestroyFunc`:

```go
for _, gvrStr := range deletedPoolScopeMetadata.UnsortedList() {
    if destroyFunc, ok := m.lazyLoadedGVRCacheDestroyFunc[gvrStr]; ok {
        destroyFunc()
    }
    delete(m.lazyLoadedGVRCacheDestroyFunc, gvrStr)
    delete(m.lazyLoadedGVRCache, gvrStr)
}
```

However, this is dead code left over from a past refactoring. The 
`lazyLoadedGVRCache` and `lazyLoadedGVRCacheDestroyFunc` maps are 
instantiated at initialization but are never populated. The actual cache 
instances are now managed by `filterStoreManager` 
(`m.filterStoreManager.filterStores`). Because 
`m.filterStoreManager.DeleteFilterStore(gvrStr)` is never called, the 
`Destroy()` method on the underlying `registry.Store` is never executed. 
The watch stream runs indefinitely in the background, continuously 
consuming memory, goroutines, and network bandwidth, even though the 
resource is no longer pool-scoped.

When a resource is removed from `pool-scoped-metadata` in the ConfigMap, 
the multiplexer should immediately terminate the watch connection for that 
resource and release the associated memory cache by correctly invoking 
`Destroy()` on its `filterStore`.

1. Start an edge node with YurtHub running the multiplexer and leader 
   election enabled.
2. Observe the active connections/goroutines watching the APIServer from 
   YurtHub.
3. Edit the `leader-hub-{pool-name}` ConfigMap and add a GVR (e.g. 
   `apps/v1/deployments`) to `pool-scoped-metadata`. Observe YurtHub 
   initiating a new watch stream.
4. Edit the ConfigMap again to remove `apps/v1/deployments` from 
   `pool-scoped-metadata`.
5. Observe that the watch stream to the APIServer for deployments remains 
   active and continues syncing events in the background, consuming memory 
   indefinitely.

This is a state-staleness bug between the ConfigMap-driven controller state 
and the actual long-running watch caches. If the ConfigMap is updated 
frequently, this results in a substantial goroutine and memory leak over 
time on edge nodes, which typically have limited resources.

**Environment**:
- Install tools: N/A
- Others:

**others**
/kind bug

WorrierKhushal:
#### What type of PR is this?
/kind bug

`MultiplexerManager.updateLeaderHubConfiguration` was supposed to tear 
down watch streams for resources removed from `pool-scoped-metadata`, 
but it was querying two dead fields (`lazyLoadedGVRCache`, 
`lazyLoadedGVRCacheDestroyFunc`) left over from a past refactor — these 
were initialized but never populated anywhere, so the cleanup logic 
never actually ran. Meanwhile, the real cache instances live in 
`filterStoreManager`, but `DeleteFilterStore` only removed the map entry 
without calling `Destroy()` on the underlying `filterStore`, so the 
watch stream and its resources (goroutines, memory, network connection) 
kept running indefinitely even after the resource was removed from 
`pool-scoped-metadata`.

This PR:
- Makes `filterStoreManager.DeleteFilterStore` call `fs.Destroy()` 
  before removing the entry.
- Updates `updateLeaderHubConfiguration` to call 
  `m.filterStoreManager.DeleteFilterStore(gvrStr)` directly instead of 
  the dead-map logic.
- Removes the now-unused `lazyLoadedGVRCache` / 
  `lazyLoadedGVRCacheDestroyFunc` fields.

Added a test verifying that removing a GVR from pool-scoped-metadata 
actually invokes `Destroy()` on its filter store.

Fixes #2753

- `go test -v ./pkg/yurthub/multiplexer/...` — all tests pass, including the new one
- `GOOS=linux go build ./...` and `go vet ./pkg/yurthub/multiplexer/...` — both clean
- `grep -r "lazyLoadedGVRCache" pkg/yurthub/` — zero remaining references, confirming safe removal
- No exported function signatures changed

```release-note
NONE
```

#### other Note

### DIFF
--- a/pkg/yurthub/multiplexer/filterstoremanager.go
+++ b/pkg/yurthub/multiplexer/filterstoremanager.go
@@ -164,5 +164,8 @@ func (fsm *filterStoreManager) DeleteFilterStore(gvrStr string) {
 	fsm.Lock()
 	defer fsm.Unlock()
 
-	delete(fsm.filterStores, gvrStr)
+	if fs, exists := fsm.filterStores[gvrStr]; exists {
+		fs.Destroy()
+		delete(fsm.filterStores, gvrStr)
+	}
 }
--- a/pkg/yurthub/multiplexer/multiplexer.go
+++ b/pkg/yurthub/multiplexer/multiplexer.go
@@ -57,8 +57,7 @@ type MultiplexerManager struct {
 	multiplexerUserAgent    string
 
 	sync.RWMutex
-	lazyLoadedGVRCache            map[string]Interface
-	lazyLoadedGVRCacheDestroyFunc map[string]func()
+
 	sourceForPoolScopeMetadata    string
 	poolScopeMetadata             sets.Set[string]
 	leaderAddresses               sets.Set[string]
@@ -81,8 +80,7 @@ func NewRequestMultiplexerManager(
 		healthCheckerForLeaders:       healthCheckerForLeaders,
 		loadBalancerForLeaders:        cfg.LoadBalancerForLeaderHub,
 		poolScopeMetadata:             poolScopeMetadata,
-		lazyLoadedGVRCache:            make(map[string]Interface),
-		lazyLoadedGVRCacheDestroyFunc: make(map[string]func()),
+
 		leaderAddresses:               sets.New[string](),
 		portForLeaderHub:              cfg.PortForMultiplexer,
 		nodeName:                      cfg.NodeName,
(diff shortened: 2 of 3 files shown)


WorrierKhushal:
Hi @rambohe-ch  @luc99hen  @wawlian  @carolove Sir i am khushal  ,

I have submitted a pull request fixing issue #2753 ([BUG] Multiplexer watch streams leak indefinitely when pool-scope metadata is removed from Leader Hub ConfigMap).

**Key Fix Details:**
- Implemented explicit `fs.Destroy()` inside `DeleteFilterStore` before map cleanup to release goroutines, memory, and active network connections.
- Refactored `updateLeaderHubConfiguration` to use `DeleteFilterStore` directly and purged legacy unpopulated fields (`lazyLoadedGVRCache`).
- Added unit test verification under `pkg/yurthub/multiplexer/` to ensure stream teardown when GVR is removed.

All tests (`go test -v ./pkg/yurthub/multiplexer/...`), `go vet`, and cross-compilation (`GOOS=linux go build ./...`) pass cleanly without errors.

Whenever you have some time, could you please take a look and share your feedback? Thanks!

Issue #1999, opened 2024-03-28
 e2e failed in prepare_autonomy_tests


vie-serendipity:
#### Which jobs are failing:
```plain-text
+ kubectl wait --for=condition=Ready pod/yurt-e2e-test-nginx-openyurt-e2e-test-worker --timeout=240s
Error from server (NotFound): pods "yurt-e2e-test-nginx-openyurt-e2e-test-worker" not found
make: *** [Makefile:114: e2e-tests] Error 1
Error: Process completed with exit code 2.
```
#### Reason for failure:
Unknown
#### labels

/kind failing-test

vie-serendipity:
@Congrool I notice e2e failed in https://github.com/openyurtio/openyurt/pull/1996 too. And it seems fine for now after rerun. Will it work if it keeps rerun or is there any way to work?

Congrool:
I guess it failed for bad performance of github action, which may timeout. If more failures occur, at that time, we have to figure out the reason. Currently, just rerun.

PR #2764, opened 2026-08-20
fix: prevent yurthub from overwriting valid kube-system token with empty bearer token in tenant mode

### LINKED ISSUE #2761: [BUG] Yurthub WithSaTokenSubstitute empties Authorization header when tenant token secret is missing
**What happened**:
`WithSaTokenSubstitute` (`pkg/yurthub/proxy/util/util.go:354-385`) replaces the request's `Authorization` header with the tenant's token whenever the request is authenticated as a kube-system service account, the configured tenant namespace differs, and the secret informer has synced (`:370-372`). It unconditionally calls **`GetTenantToken()`** and, if that returns the empty string, the header becomes `Authorization: Bearer ` (empty) — the original, still-valid token is destroyed.

`GetTenantToken()` returns `""` whenever `TenantSecret == nil` (`pkg/yurthub/tenant/tenant.go:88-96`), and `TenantSecret` is only set when the informer observes a default service-account token secret in the tenant namespace (`IsDefaultSASecret`, `tenant.go:158-161`; secret informer scoped to the tenant ns, `config.go:370-377`). On Kubernetes ≥1.24 legacy SA token secrets are no longer auto-created for the default SA, so in a tenant deployment the tenant default SA token secret may not exist at hub startup (or at all until explicitly provisioned). `WaitForCacheSync()` (`tenant.go:51-63`) only guarantees the informer's initial list was processed — it does not guarantee the token secret exists.

Secondary issue: `klog.V(2).Infof("replace token, old: %s, new: %s", oldToken, tenantMgr.GetTenantToken())` (`util.go:372`) logs full service-account tokens (credentials) in plaintext.

- Only substitute the token when a non-empty tenant token is actually available; otherwise keep the original header (and log a warning).
- Never log token contents.

**How to reproduce it (as minimally and precisely as possible)**
1. Start yurthub with a tenant organization/namespace configured but no default SA token secret present in the tenant namespace (the default on k8s ≥1.24).
2. Issue a request through the hub with a valid kube-system bearer token.
3. Observe the outbound `Authorization` header becomes `Bearer ` (empty) and the tenant kube-apiserver returns 401.

**Anything else we need to know?**
Root cause: the substitution path treats "tenant informer synced" as equivalent to "tenant token available", and never falls back to the original credential. No prior art: nothing in `git log` for `pkg/yurthub/proxy/util` or `pkg/yurthub/tenant` addresses this. Note: `TenantSecret` is also read/written without synchronization between the informer handlers and `GetTenantToken`/`WaitForCacheSync` (`tenant.go:40, :53, :88-96, :108-156`) — recommend guarding it while fixing this, but the functional bug above is the empty-token rewrite.

**Environment**:
- OpenYurt version: master (1.7+)
- Kubernetes version (use `kubectl version`): ≥1.24 (no auto-created default SA token secrets)
- OS (e.g: `cat /etc/os-release`): any
- Kernel (e.g. `uname -a`): any
- Install tools: yurthub tenant mode
- Others: tenant control-plane mode enabled
**others**
/kind bug
/priority P1

WorrierKhushal:
#### What type of PR is this?
/kind bug

`WithSaTokenSubstitute` (`pkg/yurthub/proxy/util/util.go`) replaces the 
request's `Authorization` header with the tenant's token whenever the 
request is authenticated as a kube-system service account and the tenant 
secret informer has synced. It unconditionally called `GetTenantToken()` 
and, if that returned an empty string, the header became 
`Authorization: Bearer ` (empty) — destroying the original, still-valid 
token.

`GetTenantToken()` returns `""` whenever `TenantSecret == nil`, and 
`TenantSecret` is only set once the informer observes a default 
service-account token secret in the tenant namespace. On Kubernetes 
>=1.24, legacy SA token secrets are no longer auto-created for the 
default SA, so in a tenant deployment this secret may not exist at hub 
startup (or at all until explicitly provisioned). `WaitForCacheSync()` 
only guarantees the informer's initial list was processed — it does not 
guarantee the token secret exists.

Consequence: as soon as the tenant informer synced but before the tenant 
token secret appeared, every kube-system request through the hub had its 
valid token overwritten with an empty one, and the tenant kube-apiserver 
returned 401 for all of them — with no fallback and no error surfaced, 
fully cutting off the edge node in tenant mode until the secret showed 
up.

Secondary issue: the existing log line printed the full old and new 
token values in plaintext.

This PR:
- Only rewrites the `Authorization` header when `GetTenantToken()` 
  returns a non-empty token; otherwise keeps the original header 
  untouched and logs a warning instead.
- Removes the plaintext token logging.
- Adds `TestWithSaTokenSubstituteWithStubTenant` with a stub 
  `tenant.Interface`, covering three cases: non-empty token → header 
  replaced, empty token → original header preserved, 
  `WaitForCacheSync()` false → original header preserved.

Verified locally (WSL2, matching CI's Go 1.25.0 / golangci-lint v2.11.4): 
all clean — 0 test failures, 0 data races, 0 vet issues, 0 lint issues.

Fixes #2761

When the tenant token is unavailable, requests now go out with the 
original kube-system token instead of an empty one. The tenant apiserver 
may still reject it (401/403) if RBAC doesn't map that identity, which is 
the same outcome as today during the missing-secret window — but we no 
longer actively destroy a token that could be valid (e.g., if the tenant 
apiserver is configured to accept the kube-system identity).

No other callers of `GetTenantToken`/`GetTenantNs` are affected by this 
change.

```release-note
fix(yurthub): stop overwriting valid kube-system service-account tokens with an empty bearer token in tenant mode when the tenant token secret is not yet available, which was cutting off all edge traffic until the secret appeared.
```

#### other Note

### DIFF
--- a/pkg/yurthub/proxy/util/util.go
+++ b/pkg/yurthub/proxy/util/util.go
@@ -368,8 +368,13 @@ func WithSaTokenSubstitute(handler http.Handler, tenantMgr tenant.Interface) htt
 					if tenantNs, _, err := serviceaccount.SplitUsername(oldClaim.Subject); err == nil {
 
 						if tenantMgr.GetTenantNs() != tenantNs && tenantNs == "kube-system" && tenantMgr.WaitForCacheSync() { // token is not from tenant's namespace
-							req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", tenantMgr.GetTenantToken()))
-							klog.V(2).Infof("replace token, old: %s, new: %s", oldToken, tenantMgr.GetTenantToken())
+							newToken := tenantMgr.GetTenantToken()
+							if newToken != "" {
+								req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", newToken))
+								klog.V(2).Infof("replace token for request %s", util.ReqString(req))
+							} else {
+								klog.Warningf("tenant token is not ready for request %s, keep original token", util.ReqString(req))
+							}
 						}
 
 					} else {
--- a/pkg/yurthub/proxy/util/util_test.go
+++ b/pkg/yurthub/proxy/util/util_test.go
@@ -711,6 +711,79 @@ func TestWithSaTokenSubstituteTenantTokenEmpty(t *testing.T) {
 	}
 }
 
+type stubTenantMgr struct {
+	tenantNs string
+	token    string
+	isSynced bool
+}
+
+func (s *stubTenantMgr) GetTenantNs() string {
+	return s.tenantNs
+}
+
+func (s *stubTenantMgr) GetTenantToken() string {
+	return s.token
+}
+
+func (s *stubTenantMgr) WaitForCacheSync() bool {
+	return s.isSynced
+}
+
+func (s *stubTenantMgr) SetSecret(sec *v1.Secret) {
+}
+
+func TestWithSaTokenSubstituteWithStubTenant(t *testing.T) {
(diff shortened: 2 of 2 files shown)


WorrierKhushal:
Hi @zyjhtangtang @luc99hen  sir , 
fixes #2761.

yurthub was replacing a valid kube-system token with an empty one
whenever the tenant token wasn't ready yet (common on k8s 1.24+), which
cut off all edge traffic until the tenant secret appeared. Fixed by only
replacing the token when it's actually available, otherwise keeping the
original. Also removed a line that was logging tokens in plaintext.

Added a test covering all three cases: token available, token missing,
cache not synced. Verified with go test -race, go vet, and golangci-lint
— all clean.

If CI shows a failure on prepare_autonomy_tests, that's unrelated too —
traced it and confirmed this code path isn't even active in the e2e
cluster (tenant mode is off there). Same known flake as #1999.

Thanks for reviewing!

PR #2770, opened 2026-08-26
Fix/2769 update node status

### LINKED ISSUE #2769: [BUG] UpdateNodeStatus silently swallows errors and returns stale node on failure
**What happened**:
In `pkg/yurthub/proxy/autonomy/autonomy.go`, the `updateNodeStatus` function implements a retry loop for updating node conditions. When intermittent errors occur during the retries, the code logs the error using `klog.ErrorS` but continues or exits the loop. If the loop completes with a non-nil `retNode` (from an earlier successful iteration or partial fetch), the function returns `retNode, nil` instead of propagating the failure. This silently swallows errors and returns a potentially stale node object to the proxy caller.

If the node status update fails across retries and encounters persistent non-retryable errors, the function should reject the operation and return an appropriate error (`fmt.Errorf`) to the caller rather than proceeding with a stale or un-updated node state, ensuring consistency in edge autonomy handling.

1. Trigger a node status update through the AutonomyProxy when the underlying storage or client manager encounters intermittent errors that exhaust the retry count but leave a non-nil `retNode` reference from an intermediate step.
2. Observe that `updateNodeStatus` logs the error via `klog` but returns `nil` error to `ServeHTTP`.
3. The HTTP handler proceeds as if the node autonomy state was successfully synchronized, causing state divergence.

- This is a core logic/correctness flaw affecting edge autonomy synchronization.
- Component: `pkg/yurthub/proxy/autonomy`

**Environment**:
- OpenYurt version: v1.4+ (Master/Current)
- Kubernetes version (use `kubectl version`): v1.26+
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04 / Linux
- Kernel (e.g. `uname -a`): Linux 5.15+
- Install tools: go 1.20+
- Others: N/A

/kind bug

WorrierKhushal:
What type of PR is this?
/kind bug

### Background
In `pkg/yurthub/proxy/autonomy/autonomy.go`, the `AutonomyProxy` handles node status updates via the `updateNodeStatus` method, which utilizes a retry loop (`nodeStatusUpdateRetry`) to ensure node conditions are synchronized reliably between edge components and storage/cloud managers.

### The Bug
When intermittent or persistent errors occur during the retry loop, the code previously handled them by logging via `klog.ErrorS` and continuing or breaking out of the loop based on specific error types (such as `ErrDirectClientMgr`). 

However, if the loop completed with a non-nil `retNode` reference (captured from a prior successful iteration or an intermediate partial fetch), the function fell through and returned `retNode, nil` — completely discarding the accumulated error state. 

// Before (simplified logic)
for i := 0; i < nodeStatusUpdateRetry; i++ {
    node, err = ap.tryUpdateNodeConditions(i, req)
    if node != nil {
        retNode = node
    }
    if errors.Is(err, ErrDirectClientMgr) {
        break
    } else if err != nil {
        klog.ErrorS(err, "Error getting or updating node status, will retry")
    } else {
        return retNode, nil
    }
}
if retNode == nil {
    return nil, fmt.Errorf("failed to get node")
}
klog.ErrorS(err, "failed to update node autonomy status")
return retNode, nil // <--- Returns stale node with nil error even when persistent errors occurred!

### Impact
Callers (such as `ServeHTTP` and downstream proxy consumers) received a `nil` error alongside a potentially stale or un-updated node object. This silently swallowed critical storage and synchronization failures, leading the system to proceed under the false assumption that node autonomy state was successfully updated when it actually failed. This created subtle state divergence in edge node autonomy management.

### The Fix
1. Introduced an explicit `hadError := false` tracking flag prior to entering the retry loop.
2. Set `hadError = true` whenever a non-retryable/persistent error occurs within the loop attempts.
3. Added a post-loop check: if `hadError` is true, the function now correctly rejects the operation by returning an appropriate error (`fmt.Errorf("failed to update node autonomy status after retries")`) instead of blindly returning `retNode, nil`.

// After (fixed logic)
hadError := false
for i := 0; i < nodeStatusUpdateRetry; i++ {
    node, err = ap.tryUpdateNodeConditions(i, req)
    if node != nil {
        retNode = node
    }
    if errors.Is(err, ErrDirectClientMgr) {
        break
    } else if err != nil {
        hadError = true
        klog.ErrorS(err, "Error getting or updating node status, will retry")
    } else {
        return retNode, nil
    }
}
if retNode == nil {
    return nil, fmt.Errorf("failed to get node")
}
if hadError {
    return nil, fmt.Errorf("failed to update node autonomy status after retries")
}
klog.ErrorS(err, "failed to update node autonomy status")
return retNode, nil

### Test Coverage
- Added a dedicated unit test in `autonomy_test.go` simulating persistent errors across retry iterations where an intermediate node is present, verifying that `updateNodeStatus` correctly surfaces the error rather than silently returning `retNode, nil`.
- Ensured all existing autonomy proxy tests continue to pass without regression.

Which issue(s) this PR fixes:
Fixes #2769

Special notes for your reviewer:
- `go test -v ./pkg/yurthub/proxy/autonomy/...` — all existing and new tests pass cleanly.
- `go build ./pkg/yurthub/proxy/...` — builds without errors.
- `go vet ./pkg/yurthub/proxy/autonomy/...` — zero warnings or issues.
- Confirmed `updateNodeStatus` is called exclusively from `AutonomyProxy.ServeHTTP`; downstream clients will now receive proper error propagation enabling appropriate fallback or retry behavior.

Does this PR introduce a user-facing change?
NONE


WorrierKhushal:
Hi @zyjhtangtang @luc99hen  Sir ,

I just opened a PR for issue #2769 to fix the silent error suppression issue in updateNodeStatus.

This update ensures that persistent errors during node condition updates are properly returned to callers instead of falling back to a stale node, and it includes a new unit test for coverage. All local tests, builds, and vet checks pass successfully.

I would appreciate your feedback when you have a moment. Thank you for your time!

PR #2772, opened 2026-08-27
fix: add missing return after WriteErr in GetPods handler to prevent superfluous response write

### LINKED ISSUE #2771: [BUG] Missing return statement after error response in otaupdate GetPods handler causes corrupted HTTP response
**What happened**:

In `pkg/yurthub/otaupdate/ota.go`, the HTTP handler function `GetPods` 
retrieves a list of pods from local cache storage and attempts to encode 
them via `util.EncodePods(podList)`. If `util.EncodePods` returns an 
error, the code logs it and calls 
`util.WriteErr(w, "Encode pod list failed", http.StatusInternalServerError)` 
to write a 500 response.

However, after calling `util.WriteErr`, the function does not `return`. 
Execution continues to `util.WriteJSONResponse(w, data)`. Since `data` is 
`nil` due to the encoding failure, `util.WriteJSONResponse` attempts to 
call `w.WriteHeader(http.StatusOK)` (200) and write `null` — after a 500 
has already been written. This produces a Go HTTP server warning 
(`http: superfluous response.WriteHeader call`) and corrupts the response 
body with both the 500 error payload and a trailing `null`.

```go
// pkg/yurthub/otaupdate/ota.go, lines 91-96
data, err := util.EncodePods(podList)
if err != nil {
    klog.Errorf("Encode pod list failed, %v", err)
    util.WriteErr(w, "Encode pod list failed", http.StatusInternalServerError)
}
util.WriteJSONResponse(w, data)
```

When `util.EncodePods` fails, `GetPods` should write the 500 response and 
immediately `return` from the handler.

1. Run `go test -v ./pkg/yurthub/otaupdate/...`
2. Simulate an encoding failure in `GetPods` (e.g. via a pod list that 
   fails JSON marshaling), or trigger the `/openyurt.io/v1/pods` endpoint 
   under such a condition.
3. Observe the response contains both the 500 error JSON and a trailing 
   `null`, along with the `http: superfluous response.WriteHeader call` 
   warning in logs.

Severity: Low. Scoped to one HTTP error-handling path. Does not crash 
YurtHub, but violates HTTP protocol semantics and corrupts error response 
payloads for API clients. Good first issue for new contributors.

**Environment**:
- Install tools: N/A
- Others:

**others**
/kind bug

WorrierKhushal:
#### What type of PR is this?
/kind bug

In `pkg/yurthub/otaupdate/ota.go`, the `GetPods` HTTP handler calls 
`util.EncodePods(podList)` and, on failure, correctly calls 
`util.WriteErr(w, "Encode pod list failed", http.StatusInternalServerError)` 
to write a 500 response — but was missing a `return` statement 
immediately after.

Without the `return`, execution falls through to 
`util.WriteJSONResponse(w, data)` with `data == nil`, which attempts to 
call `w.WriteHeader(http.StatusOK)` (200) on a response that has already 
been committed with a 500 status. This produces a Go HTTP warning 
(`http: superfluous response.WriteHeader call`) and appends `null` to the 
response body after the 500 error payload, corrupting the HTTP response 
for any client hitting this error path.

Every other error branch in this same file (lines 70, 76, 85, 110, 115, 
123, 131, 136, 145, 153) already follows the convention of 
`util.WriteErr(...)` immediately followed by `return` — this single 
branch was the only one missing it.

This PR adds the missing `return` statement and adds 
`TestGetPods_EncodePodsError` to verify that on an encoding failure only 
the 500 status is written and the handler returns cleanly.

Fixes #2771

- `go build ./pkg/yurthub/otaupdate/...` — clean
- `go vet ./pkg/yurthub/otaupdate/...` — clean
- `go test -v ./pkg/yurthub/otaupdate/...` — all tests pass, including the new `TestGetPods_EncodePodsError`
- One-line change, zero risk of behavioral regression on the success path
- No exported function signatures changed

```release-note
NONE
```

#### other Note


WorrierKhushal:
Hi maintainers @zyjhtangtang  @luc99hen sir 
i just wanted to tell about my this pr :
- **The Bug:** In `pkg/yurthub/otaupdate/ota.go`, the `GetPods` handler was missing a `return` statement right after `util.WriteErr(...)` when `EncodePods` failed. This caused a fall-through to `WriteJSONResponse`, resulting in a Go HTTP warning (`superfluous response.WriteHeader call`) and response corruption.

- **The Fix:** Added the missing `return` statement (consistent with every other error branch in this file) and added a new unit test `TestGetPods_EncodePodsError` to prevent regressions.

Everything builds cleanly and all tests pass. Would love a quick review when you get a chance! Thanks!

PR #2800, opened 2026-09-10
fix(util): resolve WriteObjectToYamlFile failure when target file does not exist

Norway-02:
#### What type of PR is this?
/kind bug

In `pkg/util/file/file.go`, `WriteObjectToYamlFile` failed when writing a Kubernetes object to a target path that does not yet exist because `backupFile` called `os.Open` without checking if the file existed, and `os.Remove` returned `os.ErrNotExist`.

This PR:
- Checks if the target file exists in `backupFile` before calling `os.Open`. If the file does not exist, backup is safely skipped.
- Handles `os.IsNotExist(err)` safely during target file removal prior to rename.
- Cleans up `tmpPath` if `os.Rename` fails.
- Adds a comprehensive unit test suite in `pkg/util/file/file_test.go`.

Fixes #

None

NONE


Norway-02:

PR #2801, opened 2026-09-12
fix(yurtmanager): propagate context in NodePool ValidateDelete webhook

Norway-02:
#### What type of PR is this?
/kind bug

In `pkg/yurtmanager/webhook/nodepool/v1beta2/nodepool_validation.go`, `ValidateDelete` currently discards the incoming admission request context (`_ context.Context`) and passes `context.TODO()` to `cli.List` inside `validateNodePoolDeletion`.

This PR:
- Wires the incoming `ctx` parameter into `ValidateDelete` and `validateNodePoolDeletion`.
- Adds an early `ctx.Err()` check in `validateNodePoolDeletion` to handle cancellation and deadlines.
- Passes `ctx` to `cli.List(ctx, &nodes, ...)` so request deadlines and cancellation signals propagate to the Kubernetes client.
- Updates unit tests in `nodepool_validation_test.go`.

NONE

None

NONE


Norway-02:

pkg/yurtmanager/webhook/nodepool/v1beta2/nodepool_validation.go
```diff
@@ -177,10 +177,16 @@ func validateNodePoolSpecUpdate(spec, oldSpec *appsv1beta2.NodePoolSpec) field.E
 
 // validateNodePoolDeletion validate the nodepool deletion event, which prevents
 // the default-nodepool from being deleted
-func validateNodePoolDeletion(cli client.Client, np *appsv1beta2.NodePool) field.ErrorList {
+func validateNodePoolDeletion(ctx context.Context, cli client.Client, np *appsv1beta2.NodePool) field.ErrorList {
+	if err := ctx.Err(); err != nil {
```
Aryanbhargava18, line 181: Wrapping `ctx.Err()` in `field.Forbidden` causes `ValidateDelete` to return `apierrors.NewForbidden(...)` (HTTP 403 Forbidden).

In Kubernetes admission webhooks, `403 Forbidden` signifies a **business validation denial** (e.g. attempting to delete a pool that still has nodes assigned). A context cancellation (`context.Canceled`) or client timeout (`context.DeadlineExceeded`) is an **infrastructure RPC failure**.

Returning `403 Forbidden` here has two adverse effects:
1. If the webhook is configured with `failurePolicy: Ignore`, the API server treats HTTP 403 as an intentional policy rejection rather than a webhook failure, unexpectedly blocking deletion requests during timeouts.
2. The user receives a misleading message that the NodePool deletion is forbidden due to context cancellation.

Instead, `validateNodePoolDeletion` should separate infrastructure errors from validation errors by returning `(field.ErrorList, error)`:

```go
func validateNodePoolDeletion(ctx context.Context, cli client.Client, np *appsv1beta2.NodePool) (field.ErrorList, error) {
	if err := ctx.Err(); err != nil {
		return nil, err
	}

	nodes := corev1.NodeList{}
	if err := cli.List(ctx, &nodes, client.MatchingLabels(map[string]string{projectinfo.GetNodePoolLabel(): np.Name})); err != nil {
		return nil, err
	}
	if len(nodes.Items) != 0 {
		return field.ErrorList([]*field.Error{
			field.Forbidden(field.NewPath("metadata").Child("name"),
				"cannot remove nonempty pool, please drain the pool before deleting")}), nil
	}
	return nil, nil
}
```

And in `ValidateDelete`:
```go
	allErrs, err := validateNodePoolDeletion(ctx, webhook.Client, np)
	if err != nil {
		return nil, err
	}
	if len(allErrs) > 0 {
		return nil, apierrors.NewForbidden(
			appsv1beta2.GroupVersion.WithResource("nodepools").GroupResource(),
			np.Name,
			allErrs[0],
		)
	}
```
This ensures controller-runtime returns an admission webhook error (HTTP 500 / timeout) so Kubernetes respects `failurePolicy`.

pkg/yurtmanager/webhook/nodepool/v1beta2/nodepool_validation_test.go
```diff
@@ -326,4 +326,15 @@ func TestValidateDelete(t *testing.T) {
...
+
+	t.Run("cancelled context propagates to ValidateDelete", func(t *testing.T) {
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		_, err := handler.ValidateDelete(ctx, &appsv1beta2.NodePool{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "beijing",
+			},
+		})
+		require.Error(t, err)
```
Aryanbhargava18, line 338: `require.Error(t, err)` is too permissive here. Because `ctx.Err()` was turned into a `field.Forbidden`, this assertion actually passed on an `*apierrors.StatusError` with HTTP code 403 Forbidden!

Once `ValidateDelete` returns the context error directly, we should assert the exact error and verify it is not reported as a forbidden status error:

```go
	t.Run("cancelled context propagates to ValidateDelete", func(t *testing.T) {
		ctx, cancel := context.WithCancel(context.Background())
		cancel()
		_, err := handler.ValidateDelete(ctx, &appsv1beta2.NodePool{
			ObjectMeta: metav1.ObjectMeta{
				Name: "beijing",
			},
		})
		require.Error(t, err)
		assert.ErrorIs(t, err, context.Canceled)
		assert.False(t, apierrors.IsForbidden(err), "context cancellation must not be reported as 403 Forbidden")
	})
```

Aryanbhargava18 (commented):
Thanks for working on wiring the context into `ValidateDelete`! Propagating `ctx` down to `cli.List` is a great improvement for proper cancellation and deadline propagation.

A couple of things to address before this can be merged:

### 1. PR Branch Contamination (Stacked Commits from PR #2800)
This PR currently includes commit `7e60192` (`pkg/util/file/file.go` and `file_test.go`), which belongs to PR #2800. It looks like `fix/nodepool-validatedelete-context` was branched off `fix/file-write-nonexistent-path` instead of `upstream/master`.
You can clean this up and rebase onto master with:
```bash
git fetch upstream
git rebase --onto upstream/master 7e601928e42955d604d67e039458574a731ddfe8 fix/nodepool-validatedelete-context
git push --force-with-lease origin fix/nodepool-validatedelete-context
```

### 2. Error Semantics: Context Cancellation is not HTTP 403 Forbidden
I've left inline comments on `nodepool_validation.go` and `nodepool_validation_test.go`. Context cancellation / timeout represents an infrastructure RPC failure rather than a business validation rejection (`field.Forbidden` / `apierrors.NewForbidden`). Returning 403 subverts `failurePolicy: Ignore` and confuses operators. Let's return `(field.ErrorList, error)` so internal errors bubble up to controller-runtime as admission errors.

PR #2681, opened 2026-07-27
fix(yurtmanager): initialize node.Labels before setting nodepool label in node mutating webhook

bhuvan-somisetty:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:

Fixes a potential runtime panic in the Node mutating admission webhook (`pkg/yurtmanager/webhook/node/v1/node_default.go`).

When a Node resource enters the admission webhook with `node.Labels == nil` and fallback to `apps.DesiredNodePoolLabel` occurs, line 44 attempts map assignment (`node.Labels[projectinfo.GetNodePoolLabel()] = npName`) before checking if `node.Labels` is initialized:

```go
npName := node.Labels[projectinfo.GetNodePoolLabel()]
if len(npName) == 0 {
    npName = node.Labels[apps.DesiredNodePoolLabel]
    if len(npName) != 0 {
        node.Labels[projectinfo.GetNodePoolLabel()] = npName // <-- PANIC if node.Labels is nil
    } else {
        return nil
    }
}
```

In Go, writing to a `nil` map panics with `panic: assignment to entry in nil map`. Although lines 56–58 check `if node.Labels == nil`, that check comes **after** line 44, making it ineffective.

This PR ensures `node.Labels` is initialized before writing the fallback node pool label.

#### Which issue(s) this PR fixes:

Fixes #2680

#### Special notes for your reviewer:

- Added a unit test case `"add labels for node with desired nodepool label"` in `node_default_test.go` to cover this path.
- All unit tests pass.

#### Does this PR introduce a user-facing change?

```release-note
Fix Node mutating admission webhook to initialize node.Labels map before assigning node pool labels, preventing potential runtime panic on nodes with uninitialized labels.
```

PR #2815, opened 2026-09-20
fix(loadbalancerset): initialize Labels map before updating pool service

Norway-02:
#### What type of PR is this?
/kind bug

In `compareAndUpdatePoolServiceLabel`, `currentPoolService.Labels` is read and updated with `desireLabels`. If `currentPoolService.Labels` is nil (e.g. when fetched from API server without initialized metadata labels), reading returns empty string while writing directly panics with `panic: assignment to entry in nil map`, crashing the yurt-manager reconcile loop.

This PR initializes `currentPoolService.Labels` with `make(map[string]string)` if nil before performing updates, preventing any potential runtime nil-map panics.

NONE

This follows the same nil-map defensive initialization pattern recently merged in #2710, #2708, #2692, and #2681.

NONE


Norway-02:

